PR #3, opened 2019-12-30
Fix permission denied when opening service.json

nareddyt:
Fixes permission denied in Config Manager when reading `/etc/endpoints/service.json`.

Tested manually and internally with our CI/CD.


nareddyt:
@TAOXUY please confirm that it is OK to ignore the CI errors here so I can force merge.

nareddyt:
Confirmed that public tutorial works now. 👍

PR #12, opened 2020-02-21
Remove `-local` flag for goimports

nareddyt:
Our imports are not formatted to respect this flag. Remove it, since it references the old name anyways.

Otherwise, all our imports will be moved around.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,6 @@
 
 SHELL 	:= /bin/bash
 BINDIR	:= bin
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +18,6 @@
 
 SHELL 	:= /bin/bash
 BINDIR	:= bin
-PKG 	:= cloudesf.googleresource.com/gcpproxy
```
nareddyt: Note that this is mis-spelled (google**re**source), which is probably why this never worked.

PR #20, opened 2020-02-24
Add scripts/docker_make.sh

qiwzhang:

### DIFF
--- a/scripts/docker_make.sh
+++ b/scripts/docker_make.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# Copyright 2020 Google LLC
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


scripts/docker_make.sh
```diff
@@ -0,0 +1,34 @@
...
+# could not run in image based on python:3.6-buster which has older libc.
+#
+# To build an envoy binary, run
+#
+#    scripts/docker_make.sh build-envoy
+#
+
+# The latest Prow docker image used for prow build/tests
+# It is built from docker/Dockerfile-prow-env
+IMAGE=gcr.io/cloudesf-testing/gcpproxy-prow:v20200207-v2.4.0-9-g17334b8-master
```
nareddyt: Can we just use the latest version of the prow image? Otherwise, we need to remember to always update this.

qiwzhang: Done

PR #21, opened 2020-02-25
Remove gcc

TAOXUY:
After servicecontrol proto build rule is updated, clang works without bazel cache conflict so remove gcc to build envoy during tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,7 @@ integration-test-run-parallel:
 
 integration-test: build  build-envoy build-grpc-interop build-grpc-echo integration-test-run-sequential
 
-integration-debug: build build-envoy build-grpc-interop build-grpc-echo
+integration-debug: build build-envoy-debug build-grpc-interop build-grpc-echo
 	@echo "--> running integration tests and showing debug logs"
 	@go test -timeout 20m ./tests/endpoints/...
 	@go test -v -timeout 20m ./tests/env/... --logtostderr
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -60,21 +60,15 @@ build-envoy-tsan:
...
-	@CC=gcc CXX=g++ bazel build --config=release //src/envoy:envoy
-	@cp -f bazel-bin/src/envoy/envoy bin/
 
-build-envoy-release:
+build-envoy:
 	@echo "--> building envoy (compilation_mode=release)"
 	@CC=clang-8 CXX=clang++-8 bazel build --config=clang-release //src/envoy:envoy
 	@cp -f bazel-bin/src/envoy/envoy bin/
 
-build-envoy:build-envoy-release
```
nareddyt: We don't support fastbuild for envoy? Is this worth adding in, or is `opt` ok?

Makefile
```diff
@@ -168,9 +162,9 @@ integration-test-run-parallel:
 	@go test -timeout 20m ./tests/utils/... --logtostderr
 	@go test -timeout 20m ./tests/integration_test/... --logtostderr
 
-integration-test: build  build-envoy-gcc build-grpc-interop build-grpc-echo integration-test-run-sequential
+integration-test: build  build-envoy build-grpc-interop build-grpc-echo integration-test-run-sequential
 
-integration-debug: build build-envoy-gcc build-grpc-interop build-grpc-echo
+integration-debug: build build-envoy build-grpc-interop build-grpc-echo
```
nareddyt: `build-envoy-debug` for debug integration tests?

PR #23, opened 2020-02-25
b/149491061: [Refactor 2/3] Decouple TokenSubscriber platform logic from TokenInfo for IAM and IMDS

nareddyt:
Notes:
- `TokenSubscriber`: Handles the init manager, refresh timer, making async http requests, etc.
- `TokenInfo`: A bridge for TokenSubscriber that provides request creation and response parsing logic.
- Follows the bridge design pattern.

Testing done:
- Added unit tests for TokenSubscriber and both TokenInfo.
- Kept fuzz test to work across both components.
- Verified all integration tests pass with no changes.
- Manually ran e2e tests with chain of delegates for backend auth identity token.

Future work:
- Switch over production code to use this new directory.
- The main request in b/149491061 will be implemented in a separate CL. The change is simple (removing one line), but new tests need to be added.

### DIFF
--- a/src/envoy/token/token_info.h
+++ b/src/envoy/token/token_info.h
@@ -1,0 +1,54 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "common/common/logger.h"
+#include "envoy/common/pure.h"
+#include "envoy/http/message.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace Token {
+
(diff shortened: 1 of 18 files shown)


src/envoy/token/token_info.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace Token {
+
+// For APIs that have no expiration time, use this.
+constexpr std::chrono::seconds kDefaultTokenExpiry(3599);
+
+typedef std::function<void(absl::string_view)> UpdateTokenCallback;
+typedef std::function<std::string()> GetTokenFunc;
```
qiwzhang: These two functions should not be here.  They are not used by this interface.

src/envoy/token/token_info.h
```diff
@@ -0,0 +1,54 @@
...
+#include "common/common/logger.h"
+#include "envoy/common/pure.h"
+#include "envoy/http/message.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace Token {
+
+// For APIs that have no expiration time, use this.
+constexpr std::chrono::seconds kDefaultTokenExpiry(3599);
```
qiwzhang: It feels odd to see a constant defined in a interface header file.  
Maybe create a separate header file to put all shared constants?

nareddyt: Done, also copied to tests.

src/envoy/token/token_info.h
nareddyt: Done

nareddyt:
PTAL

PR #32, opened 2020-02-27
add TLS support for downstream

JLXIA:
add TLS support for downstream, to accept HTTPS traffic
1,  when server_ssl_path is specified, means only HTTPS traffic is allowed at listener_port
2,  should support both HTTP/1.1 and HTTP2 clients with TLS, tests added.


src/go/configgenerator/listener_generator.go
```diff
@@ -50,8 +50,87 @@ const (
...
+	listener, err := makeListener(serviceInfo, port, isSsl)
+	if err != nil {
+		return nil, err
+	}
+	listeners = append(listeners, listener)
+
+	redirect_listener, err := makeRedirectListener(serviceInfo)
+	if err != nil {
+		return nil, err
+	}
```
TAOXUY: Would be better to move the part after checking `serviceInfo.Options.SslPort != 0`

docker/generic/start_proxy.py
```diff
@@ -189,6 +189,11 @@ def make_argparser():
        backward compatible for ESPv1 and will be deprecated.
        Please use the flag --listener_port.''')
 
+    parser.add_argument('-S', '--ssl_port', default=None, type=int, help='''
+    Expose a port for HTTPS requests.  Accepts both HTTP/1.x and HTTP/2
+    secure connections. Requires the certificate and key files
+    /etc/endpoints/ssl/envoy.crt and /etc/endpoints/ssl/envoy.key''')
```
TAOXUY: Are they already built in our image?

src/go/configgenerator/listener_generator.go
```diff
@@ -50,8 +50,87 @@ const (
...
+	var listeners []*v2pb.Listener
+
+	port := serviceInfo.Options.ListenerPort
+	isSsl := false
+	if serviceInfo.Options.SslPort != 0 {
+		port = serviceInfo.Options.SslPort
+		isSsl = true
+	}
+
+	listener, err := makeListener(serviceInfo, port, isSsl)
```
TAOXUY: Add a IsSsl  and getSslPort method to serviceInfo instead of assign a temporary variable, so makeListener only need one arg `serviceInfo`

src/go/configgenerator/listener_generator.go
```diff
@@ -50,8 +50,87 @@ const (
...
+		return nil, err
+	}
+	if serviceInfo.Options.SslPort != 0 {
+		listeners = append(listeners, redirect_listener)
+	}
+
+	return listeners, nil
+}
+
+func makeRedirectListener(serviceInfo *sc.ServiceInfo) (*v2pb.Listener, error) {
```
TAOXUY: Maybe move `makeRedirectListener` into MakeListener like other listeners because they are parallel,instead of moving all other listeners into `makeListener`?

TAOXUY:
Github just sent out PR email which made me think you are asking for review.
Ignore me if you aren't ready.

docker/generic/start_proxy.py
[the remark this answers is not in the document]
JLXIA: No.  User should do it themselves. For GKE, just use mountPath.

src/go/configgenerator/listener_generator.go
JLXIA: Done

docker/generic/start_proxy.py
JLXIA: Not necessary, these files are from API producers.

src/go/configgenerator/listener_generator.go
JLXIA: Done. Just move this into makeListener.

src/go/configgenerator/listener_generator.go
JLXIA: No, there is only one Listener right now.

This RedirectListener is in parallel to the existing listener.

docker/generic/start_proxy.py
qiwzhang: not need to mention "envoy",  how about server.crt, and server.key

We also need to add one or two flags to allow user to specify different path/name

src/go/configgenerator/listener_generator.go
```diff
@@ -50,8 +50,87 @@ const (
...
+	httpConMgr := &hcmpb.HttpConnectionManager{
+		CodecType:  hcmpb.HttpConnectionManager_AUTO,
+		StatPrefix: statPrefix,
+		RouteSpecifier: &hcmpb.HttpConnectionManager_RouteConfig{
+			RouteConfig: route,
+		},
+	}
+
+	jsonStr, _ := util.ProtoToJson(httpConMgr)
+	glog.Infof("adding Http Connection Manager config: %v", jsonStr)
```
qiwzhang: how about move this glog after line 100 to include filter chain data

docker/generic/start_proxy.py
```diff
@@ -518,6 +523,8 @@ def enforce_conflict_args(args):
     if len(port_flags) > 1:
         return "Multiple port flags {} are not allowed, please just use --listener_port flag".format(",".join(port_flags))
 
+    if args.ssl_port and args.ssl_port == args.listener_port:
```
qiwzhang: supporting both "ssl_port" and "listener_port" is confusing.

My intention was to make "listener_port" also supporting ssl.   Using a separate ssl flag to indicate it.

such as --ssl_server_cert_path to specify the path.  we don't hard code the path.

We may have to add other ssl flags too, such as ssl_version to indicate which versions to support

"force http redirect" is another feature. It is not very useful feature.  Normally, a container only open one port,  either http or https.   It is rare to open two ports; http and https

qiwzhang: I think it is better to have a design doc for this feature.

docker/generic/start_proxy.py
JLXIA: We can definitely add two flags for cert and key. But since ESPv1 doesn't do this, we can always require API producers to put the cert/key into the right path. 

We can follow up this later.

src/go/configgenerator/listener_generator.go
JLXIA: Done

docker/generic/start_proxy.py
JLXIA: Other features like "ssl_server_cert_path", "ssl_version" will be added later.

Eventually we need a flag, to indicate that this is a HTTPS server, and I think use ssl_port is totally fine, which is also compatible with ESPv1. 

I can delete the redirect part, if you think it is not useful. But I do see some examples of this.

WDYT?

docker/generic/start_proxy.py
```diff
@@ -517,6 +522,8 @@ def enforce_conflict_args(args):
         port_flags.append("--listener_port")
     if len(port_flags) > 1:
         return "Multiple port flags {} are not allowed, please just use --listener_port flag".format(",".join(port_flags))
+    if args.server_ssl_path and args.server_ssl_path[-1] == "/":
+        return "Don't include the tailing '/' in Flag --server_ssl_path."
```
qiwzhang: Why not?  we should allow both; with trailing "/" and without

src/go/configgenerator/route_generator.go
```diff
@@ -166,6 +168,37 @@ func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*v2pb.RouteConfigurat
 	}, nil
 }
 
+func MakeRedirectRouteConfig(serviceInfo *configinfo.ServiceInfo) (*v2pb.RouteConfiguration, error) {
```
qiwzhang: Is this used? If no, can we remove it?

src/go/configmanager/flags/flags.go
```diff
@@ -57,17 +57,14 @@ var (
 	ListenerPort = flag.Int("listener_port", 8080, "listener port")
 	Healthz      = flag.String("healthz", "", "path for health check of ESPv2 proxy itself")
 
+	ServerSslPath = flag.String("server_ssl_path", "", "Path to the certificate and key that ESPv2 uses to act as a HTTPS server")
```
qiwzhang: how ssl_server_path?   we can prefix all ssl related flags with ssl_

tests/endpoints/echo/client/client.go
```diff
@@ -197,3 +205,44 @@ func DoCorsPreflightRequest(url, origin, requestMethod, requestHeader, referer s
 	defer resp.Body.Close()
 	return resp.Header, nil
 }
+
+// DoHttpsGet performs a HTTPS Get request to a specified url
+func DoHttpsGet(url string, httpVersion int, certPath, keyPath string) ([]byte, error) {
```
qiwzhang: keyPath is not used, can we remove it?

src/go/configgenerator/listener_generator.go
```diff
@@ -50,8 +50,17 @@ const (
...
+// MakeListeners provides dynamic listeners for Envoy
+func MakeListeners(serviceInfo *sc.ServiceInfo) ([]*v2pb.Listener, error) {
+	listener, err := makeListener(serviceInfo)
+	if err != nil {
+		return nil, err
+	}
+	return []*v2pb.Listener{listener}, nil
+}
+
 // MakeListener provides a dynamic listener for Envoy
```
TAOXUY: \*m\*akeListener

src/go/configgenerator/route_generator.go
```diff
@@ -32,8 +32,10 @@ import (
 )
 
 const (
-	routeName       = "local_route"
-	virtualHostName = "backend"
+	routeName               = "local_route"
+	redirectRouteName       = "redirect_route"
+	virtualHostName         = "backend"
+	redirectVirtualHostName = "redirect_backend"
```
TAOXUY: If MakeRedirectRouteConfig is moved, ditto.

tests/endpoints/echo/client/client.go
TAOXUY: Maybe set enum for httpVersion

tests/integration_test/transport_security_test/transport_security_test.go
```diff
@@ -149,3 +150,86 @@ func TestServiceControlWithTLS(t *testing.T) {
...
+
+	if err := s.Setup(args); err != nil {
+		t.Fatalf("fail to setup test env, %v", err)
+	}
+
+	testData := []struct {
+		desc         string
+		httpsVersion int
+		certPath     string
+		keyPath      string
```
TAOXUY: No need for this if key path is not used by client side.

docker/generic/start_proxy.py
JLXIA: Done.

src/go/configgenerator/route_generator.go
JLXIA: Done

src/go/configmanager/flags/flags.go
JLXIA: I feel server_ssl_path/client_ssl_path is better than ssl_server_path/ssl_client_path.

tests/endpoints/echo/client/client.go
JLXIA: Good catch, thanks. Deleted.

src/go/configgenerator/route_generator.go
JLXIA: Done

tests/endpoints/echo/client/client.go
JLXIA: I think it is not necessary, enum doesn't give any benefit.

tests/integration_test/transport_security_test/transport_security_test.go
JLXIA: Deleted.

src/go/configgenerator/listener_generator.go
JLXIA: Done

JLXIA:
PTAL~

src/go/configgenerator/listener_generator.go
```diff
@@ -182,7 +191,28 @@ func MakeListener(serviceInfo *sc.ServiceInfo) (*v2pb.Listener, error) {
...
+	filterChain := &listenerpb.FilterChain{
+		Filters: []*listenerpb.Filter{
+			{
+				Name:       util.HTTPConnectionManager,
+				ConfigType: &listenerpb.Filter_TypedConfig{TypedConfig: httpFilterConfig},
+			},
+		},
+	}
+
+	listenerName := "http_listener"
```
TAOXUY: Maybe move to util

tests/integration_test/transport_security_test/transport_security_test.go
```diff
@@ -149,3 +148,80 @@ func TestServiceControlWithTLS(t *testing.T) {
...
+	for _, tc := range testData {
+		func() {
+			var resp []byte
+			var err error
+
+			url := fmt.Sprintf("https://localhost:%v/simpleget?key=api-key", s.Ports().ListenerPort)
+			resp, err = client.DoHttpsGet(url, tc.httpsVersion, tc.certPath)
+			if tc.wantError == nil {
+				if err != nil {
+					t.Fatal(err)
```
nareddyt: Be careful with your `Fatal`, it will prevent other tests in the loop from running.

src/go/configmanager/flags/flags.go
qiwzhang: I feel that if all ssl related flags can be prefixed with "ssl_",  it will be easier for users to find it.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -1103,6 +1103,140 @@ func TestHealthCheckFilter(t *testing.T) {
 	}
 }
 
+func TestMakeListeners(t *testing.T) {
```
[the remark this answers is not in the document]
JLXIA: Will do in seperate PR.

examples/auth/envoy_config.json
```diff
@@ -7,6 +7,7 @@
   "staticResources": {
     "listeners": [
       {
+        "name": "http_listener",
```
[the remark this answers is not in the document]
JLXIA: ssl is a wellknown config for envoy, not something specific to ESPv2. There are lots of envoy configs in Envoy repo already, I don't think we need to add extra here.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -1103,6 +1103,140 @@ func TestHealthCheckFilter(t *testing.T) {
...
+	testdata := []struct {
+		desc              string
+		protocol          string
```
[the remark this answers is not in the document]
JLXIA: Deleted.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -1103,6 +1103,140 @@ func TestHealthCheckFilter(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		listeners, err := MakeListeners(fakeServiceInfo)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if len(listeners) != len(tc.wantListeners) {
+			t.Errorf("Test Desc(%d): %s, MakeListeners failed,\ngot: %d, \nwant: %d", i, tc.desc, len(listeners), len(tc.wantListeners))
```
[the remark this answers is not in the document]
JLXIA: Done

src/go/util/transport_socket.go
```diff
@@ -15,14 +15,21 @@
...
 	"github.com/golang/protobuf/ptypes"
 
 	authpb "github.com/envoyproxy/go-control-plane/envoy/api/v2/auth"
 	corepb "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
 )
 
-// CreateTransportSocket creates a TransportSocket
-func CreateTransportSocket(hostname, rootCertsPath string, alpn_protocols []string) (*corepb.TransportSocket, error) {
+const (
+	defaultServerSslFilename = "server"
```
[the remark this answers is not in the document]
JLXIA: Done

src/go/util/transport_socket.go
```diff
@@ -53,3 +60,41 @@ func CreateTransportSocket(hostname, rootCertsPath string, alpn_protocols []stri
 		},
 	}, nil
 }
+
+// CreateDownstreamTransportSocket creates a TransportSocket for Downstream
+func CreateDownstreamTransportSocket(sslPath string) (*corepb.TransportSocket, error) {
```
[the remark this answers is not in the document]
JLXIA: Will do in seperate PR. Actually I have test cases for both with and without "/".

tests/integration_test/transport_security_test/transport_security_test.go
```diff
@@ -149,3 +148,80 @@ func TestServiceControlWithTLS(t *testing.T) {
...
+		{
+			desc:         "Fail for HTTP2 client, with incorrect key and cert",
+			httpsVersion: 2,
+			certPath:     platform.GetFilePath(platform.ProxyCert),
+			wantError:    fmt.Errorf("x509: certificate signed by unknown authority"),
+		},
+	}
+
+	for _, tc := range testData {
+		func() {
```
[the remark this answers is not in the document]
JLXIA: Run all tests in parallel.

tests/integration_test/transport_security_test/transport_security_test.go
JLXIA: If a test case fatal, why do we care about other cases?

nareddyt: Good point

tests/integration_test/transport_security_test/transport_security_test.go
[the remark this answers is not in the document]
nareddyt: I don't think it runs it in parallel unless you do `go func()`. Please remove, we don't need them to be in parallel anyways.

src/go/configgenerator/listener_generator.go
JLXIA: I think the principal is putting some common strings in util, so that it can be used by multi files. Here this listener name is specific used here only.

tests/integration_test/transport_security_test/transport_security_test.go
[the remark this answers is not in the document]
JLXIA: Done

src/go/configmanager/flags/flags.go
JLXIA: Done, changed to ssl_server_path

PR #35, opened 2020-02-28
Delete clusters async in serverless janitor

nareddyt:
Sometimes causes timeouts, deleting a cluster takes long.


nareddyt:
This code is not production code, ignoring test flake.

PR #36, opened 2020-02-28
b/149050012: Support HTTP protocols for remote backends

nareddyt:
Modify `ParseBackendProtocol` to differentiate between `HTTP1` and `HTTP2`.

Implementation notes:
- Add `UNKNOWN` to `BackendProtocol`, easier to test and safer in prod code.
- If http protocol is empty string, then default to `HTTP1`.

Testing:
- New unit tests.
- Added http protocol to dynamic routing example.
- Will add integration tests once #32 is merged, as that adds a h2 client.

### DIFF
--- a/examples/dynamic_routing/envoy_config.json
+++ b/examples/dynamic_routing/envoy_config.json
@@ -95,6 +95,7 @@
             },
             {
                 "connectTimeout": "20s",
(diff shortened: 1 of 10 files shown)


nareddyt:
@JLXIA PTAL

PR #37, opened 2020-02-29
b/149491061: Signal ready only when all tokens are successfully retried

nareddyt:
Change is simple: Just remove the `ready()` on failures in `TokenSubscriber`.

Testing Done:
- Unit tests for retry.
- Adjusted pre-existing integration tests to not fail when envoy fails to start.
- Added new integration tests to ensure envoy does not start up.

### DIFF
--- a/src/envoy/token/token_subscriber_test.cc
+++ b/src/envoy/token/token_subscriber_test.cc
@@ -70,7 +70,10 @@ class TokenSubscriberTest : public testing::Test {
 
     // Setup mock refresh timer.
     EXPECT_CALL(context_.dispatcher_, createTimer_(_))
-        .WillOnce(Invoke([this](Envoy::Event::TimerCb) { return mock_timer_; }))
+        .WillOnce(Invoke([this](Envoy::Event::TimerCb cb) {
+          timer_cb_ = cb;
+          return mock_timer_;
+        }))
         .RetiresOnSaturation();
 
     // Create token subscriber under test.
@@ -104,6 +107,7 @@ class TokenSubscriberTest : public testing::Test {
   // Mocks for init.
   Envoy::Init::TargetHandlePtr init_target_handle_;
   Envoy::Event::MockTimer* mock_timer_;
+  Envoy::Event::TimerCb timer_cb_;
   NiceMock<Envoy::Init::ExpectableWatcherImpl> init_watcher_;
   bool init_ready_ = false;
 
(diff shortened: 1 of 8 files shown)


nareddyt:
NOT ready for review

src/envoy/token/token_subscriber_test.cc
```diff
@@ -332,6 +332,58 @@ TEST_F(TokenSubscriberTest, Success) {
...
+
+  // Start class under test.
+  setUp(TokenType::AccessToken);
+
+  // Assert subscriber did not succeed.
+  ASSERT_EQ(call_count_, 0);
+  ASSERT_FALSE(init_ready_);
+
+  // Part 2: Retry, will result in a success.
+  init_target_handle_->initialize(init_watcher_);
```
qiwzhang: My understanding is:  this init call starts everything.   we don't do anything before this call. so expecting call_count as 0 is not testing anything.

we should use client_callback->onFailure or On Success to control the flow.

I am not sure if the whole test is doing this.  Could you double check it?

nareddyt: Line 363 `setUp` will actually start the class by calling `initialize()`. Then the async mock will kick in, the rest of the class with be executed.

Is there a way to make this more clear in our test case?

qiwzhang: In Envoy initalization flow, Envoy never call init_target_handle_->initialize() twice.  It is already called in setup(), so this call is not the same as production flow.

nareddyt: Discussed offline. Captured the callback for the timer, done.

tests/integration_test/iam_imds_data_path_test/iam_imds_data_path_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"--rollout_strategy=fixed",
+	"--backend_dns_lookup_family=v4only",
+	"--suppress_envoy_headers",
+}
+
+func TestDataPathImdsSuccessWhenIamDown(t *testing.T) {
+	s := env.NewTestEnv(comp.TestDataPathImdsSuccessWhenIamDown, platform.EchoRemote)
+
+	// Simulate Iam failures.
+	s.SetIamResps(map[string]string{}, 100)
```
qiwzhang: could you add comment on how IAM or Imds is configured in this test?  what configure determine IAM or Imds for service_control and backend_auth?

qiwzhang: Also don't undersand why IAM is down, but request still works.

nareddyt: I re-wrote the integration test with more fields and test cases. Maybe this will be clearer?

PR #38, opened 2020-03-02
add a positive test for TLS between proxy and remote backend

JLXIA:
Add a positive test for TLS between proxy and remote backend

### DIFF
--- a/tests/integration_test/dynamic_routing_test/dynamic_routing_integration_test.go
+++ b/tests/integration_test/dynamic_routing_test/dynamic_routing_integration_test.go
@@ -474,27 +474,53 @@ func TestServiceControlRequestForDynamicRouting(t *testing.T) {
 }
 
 func TestDynamicBackendRoutingTLS(t *testing.T) {
-
-	s := env.NewTestEnv(comp.TestDynamicBackendRoutingTLS, platform.EchoRemote)
-	s.UseWrongBackendCertForDR(true)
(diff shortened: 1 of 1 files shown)


tests/integration_test/dynamic_routing_test/dynamic_routing_integration_test.go
```diff
@@ -525,22 +551,38 @@ func TestDynamicBackendRoutingTLS(t *testing.T) {
...
+		func() {
+			s := env.NewTestEnv(comp.TestDynamicBackendRoutingTLS, platform.EchoRemote)
+			s.UseWrongBackendCertForDR(tc.useWrongCert)
+			defer s.TearDown()
 
-		if err == nil || !strings.Contains(err.Error(), tc.wantError) {
-			t.Errorf("Test (%s): failed, want error: %v, got error: %v", tc.desc, tc.wantError, err)
-		}
+			if err := s.Setup(testDynamicRoutingArgs); err != nil {
+				t.Fatalf("fail to setup test env, %v", err)
```
qiwzhang: change to "Failf"  or "Errorf"?

JLXIA: Based on go readiability guidance:

"t.Fatal is usually only appropriate when some piece of test setup fails, without which you cannot run the test at all. "

PR #41, opened 2020-03-03
cleanup duplicated normalizeJson

JLXIA:
Cleanup duplicated normalizeJson

Notes:  ideally this code should be in tests/utils, but there is a cyclic dependent for platform.


src/go/util/httptest_util.go
```diff
@@ -71,3 +73,18 @@ func InitMockServerFromPathResp(pathResp map[string]string) *httptest.Server {
...
+
+// JsonEqual compares two JSON strings after normalizing them.
+func JsonEqual(x, y *string) bool {
+	normalizeJson(x)
+	normalizeJson(y)
+	return strings.EqualFold(*x, *y)
+}
+
+// NormalizeJson returns normalized JSON string.
+func normalizeJson(input *string) {
```
qiwzhang: can we not modify the input? Maybe there are cases that input could not be modified.

src/go/util/httptest_util.go
```diff
@@ -71,3 +73,18 @@ func InitMockServerFromPathResp(pathResp map[string]string) *httptest.Server {
 		w.WriteHeader(http.StatusNotFound)
 	}))
 }
+
+// JsonEqual compares two JSON strings after normalizing them.
+func JsonEqual(x, y *string) bool {
```
qiwzhang: do we want to be consistent on the calling order; first one is "got", the second is "want"?  for consistency maybe?

nareddyt: If these are just being used for tests, why pass a pointer? Efficiency doesn't matter, just let this be a copy.

Also maybe return `error` instead of just a bool here. The error can be something like:
```
JSON differ: got: %v \n want: %v
```

Then each test can just output the error instead of duplicating this code.

src/go/util/httptest_util.go
TAOXUY: +1

src/go/util/httptest_util.go
TAOXUY: Maybe we can use this 
https://gist.github.com/turtlemonvh/e4f7404e28387fadb8ad275a99596f67

src/go/util/httptest_util.go
JLXIA: The reason I modify the input is to make it easier for debug. 

The expected and wanted JSON string may have different order. After normalize, it is much easier to see clearly where the difference is.

JLXIA: add a comments, to clearify when to use this.

src/go/util/httptest_util.go
JLXIA: Done.

nareddyt: Also if you end up returning an `error`, it will be printed in normalized format without having to modify the source.

src/go/util/httptest_util.go
```diff
@@ -71,3 +74,22 @@ func InitMockServerFromPathResp(pathResp map[string]string) *httptest.Server {
 		w.WriteHeader(http.StatusNotFound)
 	}))
 }
+
+// JsonEqual compares two JSON strings after normalizing them.
+// Should be used for test only.
+func JsonEqual(want, got string) error {
```
TAOXUY: Would be better to put in src/go/util/marshal.go than in this file?

src/go/util/httptest_util.go
```diff
@@ -71,3 +74,22 @@ func InitMockServerFromPathResp(pathResp map[string]string) *httptest.Server {
...
+		return fmt.Errorf("got: %s \n want: %s", got, want)
+	}
+	return nil
+}
+
+// normalizeJson returns normalized JSON string.
+func normalizeJson(input string) string {
+	var jsonObject map[string]interface{}
+	json.Unmarshal([]byte(input), &jsonObject)
+	outputString, _ := json.Marshal(jsonObject)
```
TAOXUY: throw error when marshal gets failed

src/go/configmanager/config_manager_test.go
```diff
@@ -1438,12 +1437,10 @@ func TestFetchListeners(t *testing.T) {
...
 			}
 
-			// Normalize both wantedListeners and gotListeners.
-			gotListeners = normalizeJson(gotListeners, t)
-			if want := normalizeJson(tc.wantedListeners, t); gotListeners != want {
+			if err := util.JsonEqual(tc.wantedListeners, gotListeners); err != nil {
 				t.Errorf("Test Desc(%d): %s, snapshot cache fetch got unexpected Listeners", i, tc.desc)
 				t.Errorf("Actual: %s", gotListeners)
-				t.Errorf("Expected: %s", want)
+				t.Errorf("Expected: %s", tc.wantedListeners)
```
nareddyt: Nit: Change this to print out `err`, which has the normalized JSONs

TAOXUY:
Apply `jsonEqual` here to compare the json after marshalling 

https://oss-prow.knative.dev/view/gcs/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/41/ESPv2-presubmit/1234922064201125888

PR #43, opened 2020-03-04
add backward compability for ssl_port

JLXIA:
add ESPv1 backward compability for ssl_port

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -155,16 +155,6 @@ environment variable or by passing "-k" flag to this script.
        It supports HTTP/1.x, HTTP/2, and gRPC connections.
        Default is {port}'''.format(port=DEFAULT_LISTENER_PORT))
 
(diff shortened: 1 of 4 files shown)


docker/generic/start_proxy.py
```diff
@@ -506,6 +513,9 @@ def enforce_conflict_args(args):
     if args.backend_dns_lookup_family and args.backend_dns_lookup_family not in {"auto", "v4only", "v6only"}:
         return "Flag --backend_dns_lookup_family must be 'auto', 'v4only' or 'v6only'."
 
+    if args.ssl_port and args.ssl_server_path:
+        return "Flag --ssl_port is going to be deprecated, please use --ssl_server_path only."
```
nareddyt: Nit: Make it clear that `--listener_port` / `--http_port` will service SSL traffic in this case.

JLXIA: it is stated on --ssl_server_path.

PR #44, opened 2020-03-04
b/149484856,b/150725812: Support custom jwt locations

TAOXUY:
- Support custom jwt locations
- Fix hard-coded ignoreQueryParam in transcoder filter
- add integration test for api-key location under grpc transcoding
- add integration test for jwt locations under grpc transcoding


src/go/configinfo/service_info.go
```diff
@@ -504,6 +510,22 @@ func (s *ServiceInfo) processUsageRule() error {
 	return nil
 }
 
+func (s *ServiceInfo) processJwtInQueryParams() {
+	authn := s.serviceConfig.GetAuthentication()
+	for _, provider := range authn.GetProviders() {
```
qiwzhang: need to do that per Provider.   Default is per provider:  if jwt_locations is defined, use it, otherwise use default.

src/go/configinfo/service_info.go
```diff
@@ -516,12 +538,16 @@ func (s *ServiceInfo) processSystemParameters() error {
 		if err != nil {
 			return err
 		}
-		extractAPIKeyLocations(method, apiKeyLocationParameters)
+		s.extractAPIKeyLocations(method, apiKeyLocationParameters)
 	}
+
+	s.ApiKeyQueryParams[util.DefaultApiKeyQueryParamKey] = true
```
qiwzhang: default is per rule or selector.

src/go/configgenerator/listener_generator.go
```diff
@@ -601,11 +633,21 @@ func makeTranscoderFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
...
+			}
+			for apiKeyQueryParam := range serviceInfo.ApiKeyQueryParams {
+				transcodeConfig.IgnoredQueryParameters = append(transcodeConfig.IgnoredQueryParameters, apiKeyQueryParam)
+			}
+
+			for jwtQueryParam := range serviceInfo.JwtQueryParams {
+				transcodeConfig.IgnoredQueryParameters = append(transcodeConfig.IgnoredQueryParameters, jwtQueryParam)
 			}
+
+			sort.Sort(sort.StringSlice(transcodeConfig.IgnoredQueryParameters))
```
nareddyt: Any specific reason for the sort? Is it just for readability?

src/go/configgenerator/listener_generator.go
```diff
@@ -279,6 +280,44 @@ func hasPathParameter(httpPattern string) bool {
...
+
+	for _, jwtLocation := range provider.JwtLocations {
+		switch x := jwtLocation.In.(type) {
+		case *confpb.JwtLocation_Header:
+			jwtHeaders = append(jwtHeaders, &jwtpb.JwtHeader{
+				Name:        jwtLocation.GetHeader(),
+				ValuePrefix: jwtLocation.GetValuePrefix(),
+			})
+		case *confpb.JwtLocation_Query:
+			jwtParams = append(jwtParams, jwtLocation.GetQuery())
```
nareddyt: Should we validate that `value_prefix` is empty in this case?

src/go/configgenerator/listener_generator_test.go
```diff
@@ -71,9 +71,86 @@ func TestTranscoderFilter(t *testing.T) {
...
+   "name":"envoy.grpc_json_transcoder",
+   "typedConfig":{
+      "@type":"type.googleapis.com/envoy.config.filter.http.transcoder.v2.GrpcJsonTranscoder",
+      "autoMapping":true,
+      "convertGrpcStatus":true,
+      "ignoredQueryParameters":[
+         "access_token",
+         "api_key",
+         "jwt_query_param",
          "key",
```
nareddyt: Why is `access_token`, `api_key`, and `key` still here? It's not in the system params or jwt locations? I thought we would not use the default values if user overrides jwt locations.

src/go/configinfo/service_info.go
```diff
@@ -49,6 +49,10 @@ type ServiceInfo struct {
 	Methods map[string]*methodInfo
 	// Stores url segment names, mapping snake name to Json name.
 	SegmentNames []*pmpb.SegmentName
+	// Stores the api key locations in query parameters.
+	ApiKeyQueryParams map[string]bool
```
nareddyt: Comments are not clear.

> Stores the query params that are checked for API Key validation.

> Stores the query params that are checked for JWT authentication.

src/go/configgenerator/listener_generator.go
```diff
@@ -601,11 +633,21 @@ func makeTranscoderFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
 				DescriptorSet: &transcoderpb.GrpcJsonTranscoder_ProtoDescriptorBin{
 					ProtoDescriptorBin: configContent,
 				},
-				IgnoredQueryParameters: []string{"api_key", "key", "access_token"},
-				AutoMapping:            true,
-				ConvertGrpcStatus:      true,
+				AutoMapping:       true,
+				ConvertGrpcStatus: true,
+			}
+			for apiKeyQueryParam := range serviceInfo.ApiKeyQueryParams {
```
nareddyt: I would also ensure the value is set to `true`. If if it will never be `false`, it will catch future bugs.

src/go/configinfo/service_info.go
```diff
@@ -504,6 +510,22 @@ func (s *ServiceInfo) processUsageRule() error {
...
+			switch jwtLocation.In.(type) {
+			case *confpb.JwtLocation_Query:
+				s.JwtQueryParams[jwtLocation.GetQuery()] = true
+			default:
+				continue
+			}
+		}
+	}
+
+	s.JwtQueryParams[util.DefaultJwtQueryParamAccessToken] = true
```
nareddyt: I thought that if a user defines a provider, then we will not use the default anymore?

tests/endpoints/echo/client/client.go
```diff
@@ -68,11 +68,12 @@ func DoWithHeaders(url, method, message string, headers map[string]string) ([]by
 		return nil, fmt.Errorf("create request error: %v", err)
 	}
 
-	if method == "POST" || method == "PATCH" || method == "PUT" {
+	if method != "GET" {
```
nareddyt: `OPTIONS`? Make sure our CORS tests don't use this

tests/env/testdata/fake_jwt.go
```diff
@@ -14,13 +14,18 @@
...
+)
+
 type ProviderConfig struct {
 	Id               string
 	Issuer           string
 	Keys             string
 	IsInvalid        bool   // If invalid, then a server that always returns 503 is started up
 	IsNonexistent    bool   // If non-existent, then a server is not started up
 	HardcodedJwksUri string // This needs to be set for non-existent, since the URL cannot be derived
+	JwtLocations     []*confpb.JwtLocation
```
nareddyt: Shouldn't this be set in the service config? I'm confused how this is applied to the service config

tests/endpoints/echo/client/client.go
TAOXUY: This method is not for OPTIONS. Anyway, change the condition for only checking if the body is empty.

src/go/configinfo/service_info.go
TAOXUY: Theoretically, they can define the combination of custom ones and default ones.

src/go/configgenerator/listener_generator.go
TAOXUY: Good point.

tests/env/testdata/fake_jwt.go
TAOXUY: This is per provider http://google3/google/api/auth.proto?l=320&rcl=297378306

tests/integration_test/jwt_locations_test/jwt_locations_test.go
```diff
@@ -0,0 +1,179 @@
+// Copyright 2019 Google LLC
```
JLXIA: just merge this tests into jwt_auth_integration_test.go.

We have too many tests folders now, which is totally unnecessary.

TAOXUY: Gotcha. 

BTW, our tests folders are totally unnecessary since we can still make our integration tests without put into different library. I suggest we can move them out of the test folders into one simple integration_test folder like before. Are you OK with that?

nareddyt: > We have too many tests folders now, which is totally unnecessary.

This is so the tests run in parallel. Each folder runs all the tests as one process. Currently all tests run in 2 minutes on my desktop, which is very useful when developing.

> I suggest we can move them out of the test folders into one simple integration_test folder like before

If this is just to make our folder structure look nice, I don't agree. I would rather not wait 25 minutes for tests to run.

api/envoy/http/service_control/requirement.proto
```diff
@@ -18,11 +18,11 @@ package google.api.envoy.http.service_control;
 
 import "validate/validate.proto";
 
-// APIKeyLocation defines the location to extract api key.
+// ApiKeyLocation defines the location to extract api key.
```
nareddyt: This is a breaking change for API Gateway team. From my understanding, we don't have a way of versioning our protos yet. 

@toddbeckman-g, could you clarify if this is safe to import?

src/go/configgenerator/listener_generator.go
nareddyt: Just curious, still have the same question

tests/integration_test/jwt_locations_test/jwt_locations_test.go
TAOXUY: > > We have too many tests folders now, which is totally unnecessary.
> 
> This is so the tests run in parallel. Each folder runs all the tests as one process. Currently all tests run in 2 minutes on my desktop, which is very useful when developing.
> 
> > I suggest we can move them out of the test folders into one simple integration_test folder like before
> 
> If this is just to make our folder structure look nice, I don't agree. I would rather not wait 25 minutes for tests to run.

still make our integration tests *parallel*

As what I see, there is no difference between putting tests into different package and manually setting t.parallel with `go test -p `

https://medium.com/@xcoulon/how-to-avoid-parallel-execution-of-tests-in-golang-763d32d88eec
https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently

nareddyt: I see, could you try it out on our repo and make sure it works? If they still run in parallel, then that's definitely cleaner.

api/envoy/http/service_control/requirement.proto
TAOXUY: Good catch!

nareddyt: Oh well maybe just changing the message name is not breaking?

toddbeckman-g: We don't really use any of these fields directly through code, so the main question is if the generated json fields remain the same before and after this change.

tests/integration_test/jwt_locations_test/jwt_locations_test.go
TAOXUY: > just merge this tests into jwt_auth_integration_test.go.
> 
> We have too many tests folders now, which is totally unnecessary.

The jwt_authn already has two tests and it maybe better to jwt_location in a separate file. 

Will remove folders in separate PR.

api/envoy/http/service_control/requirement.proto
TAOXUY: As discussed offline, simple renaming without change field is backward compatible.

src/go/configgenerator/listener_generator.go
TAOXUY: Stable order for test

PR #45, opened 2020-03-04
b/139312546: Add override for platform.

toddbeckman-g:
This will be used to make a distinction in metrics between manage-your-own gateways and managed ones.

### DIFF
--- a/src/go/gcsrunner/main/runner.go
+++ b/src/go/gcsrunner/main/runner.go
@@ -54,7 +54,6 @@ var (
 	envoyBinaryPath = flag.String("envoy_bin_path", "bin/envoy", "Location of the Envoy binary.")
 	envoyLogLevel   = flag.String("envoy_log_level", "info",
 		"Envoy logging level. Default is `info`. Options are: [trace][debug][info][warning][error][critical][off]")
-	overridePlatform = flag.String("override_platform", "", "If specified, configures `platform` field in Service Control reports.")
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


src/go/gcsrunner/main/runner.go
```diff
@@ -54,6 +54,7 @@ var (
 	envoyBinaryPath = flag.String("envoy_bin_path", "bin/envoy", "Location of the Envoy binary.")
 	envoyLogLevel   = flag.String("envoy_log_level", "info",
 		"Envoy logging level. Default is `info`. Options are: [trace][debug][info][warning][error][critical][off]")
+	overridePlatform = flag.String("override_platform", "", "If specified, configures `platform` field in Service Control reports.")
```
JLXIA: There is "ComputePlatformOverride" already in the options. 

I would think you only need to set this in the src/go/bootstrap/static/bootstrap.go, set it to "Cloud API Gateway".

src/go/gcsrunner/transform_config.go
```diff
@@ -39,13 +39,20 @@ var (
...
+		overridePlatform = oldAttrs.GetPlatform()
+	}
 	co := options.DefaultCommonOptions()
 	co.MetadataURL = opts.MetadataURL
 	mf := metadata.NewMetadataFetcher(co)
 	attrs, err := mf.FetchGCPAttributes()
 	if err != nil {
 		return err
 	}
+	if overridePlatform != "" {
```
qiwzhang: This seems buggy.   How about we define a separate field "platform_override" in service_control filter config. GCPAttributes are always fetched from Metadata.   Let c++ code to do the platform override.

toddbeckman-g: As was pointed out above:
https://github.com/GoogleCloudPlatform/esp-v2/pull/45#discussion_r387839265

this platform override field is already a feature of the bootstrap. This change will just stop GCS Runner from ignoring that override with this one. It might make sense in the future to update Envoy by changing this implementation, but for now I was hoping to unblock our use of the feature to get metrics asap.

qiwzhang: Ok, then.

PR #50, opened 2020-03-04
update script after migrating to github

TAOXUY:


TAOXUY:
It is release script which is not related to functionality. Submit directly

PR #51, opened 2020-03-04
Update envoy to 2020-03-03

qiwzhang:
Key changes:
1)  HeaderMap splited into 
    RequestHeaderMap, RequestTrailerMap, ResponseHeaderMap, ResponseTrailerMap
2) stream_info.filterState() is a share_ptr,  not a reference
3) handle_impLtest fails, change its macro for easy debug test falure.

### DIFF
--- a/src/envoy/http/backend_auth/filter.cc
+++ b/src/envoy/http/backend_auth/filter.cc
@@ -30,7 +30,7 @@ namespace BackendAuth {
 using Http::FilterDataStatus;
 using Http::FilterHeadersStatus;
 using Http::FilterTrailersStatus;
-using Http::HeaderMap;
+using Http::RequestHeaderMap;
 
 namespace {
 constexpr char kBearer[] = "Bearer ";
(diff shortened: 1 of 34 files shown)


src/envoy/http/backend_auth/filter.cc
```diff
@@ -43,9 +43,9 @@ typedef ConstSingleton<RcDetailsValues> RcDetails;
 
 }  // namespace
 
-FilterHeadersStatus Filter::decodeHeaders(HeaderMap& headers, bool) {
+FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
   absl::string_view operation = Utils::getStringFilterState(
-      decoder_callbacks_->streamInfo().filterState(), Utils::kOperation);
+      *decoder_callbacks_->streamInfo().filterState(), Utils::kOperation);
```
nareddyt: Should we change `Utils::getStringFilterState` to take in a shared pointer instead of dereferencing everytime?

qiwzhang: In general, if a function is not going to take ownership of object, it is better not to pass in share_ptr.

PR #52, opened 2020-03-04
Initial version

JLXIA:
Add mTLS support for upstream connection


src/go/configmanager/flags/flags.go
```diff
@@ -53,8 +53,9 @@ var (
 	ListenerPort = flag.Int("listener_port", 8080, "listener port")
 	Healthz      = flag.String("healthz", "", "path for health check of ESPv2 proxy itself")
 
-	SslServerPath = flag.String("ssl_server_path", "", "Path to the certificate and key that ESPv2 uses to act as a HTTPS server")
-	RootCertsPath = flag.String("root_certs_path", util.DefaultRootCAPaths, "Path to the root certificates to make TSL connection.")
+	SslServerPath  = flag.String("ssl_server_path", "", "Path to the certificate and key that ESPv2 uses to act as a HTTPS server")
+	SslBackendPath = flag.String("ssl_backend_path", "", "Path to the certificate and key that ESPv2 uses to enable TLS mutual authentication for HTTPS backend")
+	RootCertsPath  = flag.String("root_certs_path", util.DefaultRootCAPaths, "Path to the root certificates to make TSL connection.")
```
nareddyt: TLS not TSL

docker/generic/start_proxy.py
```diff
@@ -160,6 +160,11 @@ def make_argparser():
     HTTP/2 secure connections on listener_port. Requires the certificate and
     key files "server.crt" and "server.key" within this path.''')
 
+    parser.add_argument('--ssl_backend_path', default=None, help='''
```
qiwzhang: should we call "ssl_client_path" since the other one is called "ssl_server_path"?

qiwzhang: well, then we need another flag to ssl_backend_mtls to indicate backend need mutual tls, not service config. 
OK,  it really mean ssl_client_cert path for backend.

src/go/util/transport_socket.go
```diff
@@ -25,17 +25,17 @@ import (
 )
 
 const (
-	defaultServerSslFilename = "server"
+	defaultServerSslFilename  = "server"
+	defaultBackendSslFilename = "backend"
```
qiwzhang: call it "client"?

src/go/configmanager/flags/flags.go
JLXIA: Done

docker/generic/start_proxy.py
JLXIA: Changed to ssl_client_cert_path.

src/go/util/transport_socket.go
JLXIA: Done

PR #53, opened 2020-03-05
Add fuzz test for gen auth token

nareddyt:
Line coverage for this file is at 16%. Fuzz test will help cover nested parsing logic.


nareddyt:
Flakey test, this is only adding new unit tests.

PR #55, opened 2020-03-06
Add rudimentary fuzz test for service control filter

nareddyt:
- Fuzz the entire service control filter config and upstream/downstream/sidestream requests.
- Add some initial valid values to the corpus to kick-start OSS-Fuzz. After running locally for 5 minutes, 69% of lines are covered.
- This test is very rudimentary and has some items that can be improved (see inline TODOs and FIXMEs). Currently, this is very slow when running locally (4 exec/s).


src/envoy/http/service_control/http_call.cc
```diff
@@ -72,32 +72,40 @@ class HttpCallImpl : public HttpCall,
   // HTTP async receive methods
   void onSuccess(Http::ResponseMessagePtr&& response) override {
     ENVOY_LOG(trace, "{}", __func__);
-    const uint64_t status_code =
```
nareddyt: Diff here is messy. All I did was move this code inside a try block to handle missing `:status` header.

src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -0,0 +1,212 @@
...
+      }
+    }
+    // There should be at least 1 sidestream response, otherwise no point.
+    if (input.sidestream_response().size() < 1) {
+      throw ProtoValidationException("num sidestream", input);
+    }
+
+    // Setup mocks.
+    NiceMock<MockFactoryContext> context;
+    NiceMock<Http::MockAsyncClientRequest> response(
```
qiwzhang: should be "request"

src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -0,0 +1,212 @@
...
+          // Callback.
+          callback.onSuccess(std::move(msg));
+          return &response;
+        }));
+
+    // Fuzz the stream info.
+    TestStreamInfo stream_info =
+        Envoy::Fuzz::fromStreamInfo(input.stream_info());
+    EXPECT_CALL(mock_decoder_callbacks, streamInfo())
+        .WillRepeatedly(ReturnRef(stream_info));
```
asraa: Instead of setting expectation (slow-down), can you set `mock_decoder_callbacks.stream_info_ = stream_info`, the expectation to return `stream_info_` is already made in initialization. Does this speed-up?

src/envoy/http/service_control/filter_fuzz_test.cc
nareddyt: Done

src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -0,0 +1,212 @@
...
+  auto downstream_body =
+      Buffer::OwnedImpl(input.downstream_request().data().Get(0));
+  auto downstream_trailers =
+      Envoy::Fuzz::fromHeaders<Envoy::Http::TestRequestTrailerMapImpl>(
+          input.downstream_request().trailers());
+  // TODO(b/146671523): Uncomment this when implemented upstream.
+  // stream_info.addBytesReceived(downstream_body.length());
+
+  // Downstream functions under test.
+  filter.decodeHeaders(downstream_headers, false);
```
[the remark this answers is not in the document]
nareddyt: Good point, I copied your code from the uber filter fuzzer. Done

src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -0,0 +1,212 @@
...
+  filter.decodeData(downstream_body, false);
+  filter.decodeData(downstream_body, true);
+  filter.decodeTrailers(downstream_trailers);
```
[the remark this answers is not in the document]
nareddyt: Done

src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -0,0 +1,212 @@
...
+  // Report function under test.
+  filter.log(&downstream_headers, &upstream_headers, nullptr, stream_info);
+}
+
+DEFINE_PROTO_FUZZER(
+    const tests::fuzz::protos::ServiceControlFilterInput& input) {
+  ENVOY_LOG_MISC(trace, "{}", input.DebugString());
+
+  try {
+    TestUtility::validate(input);
```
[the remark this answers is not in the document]
nareddyt: Done

tests/fuzz/structured_inputs/service_control_filter.proto
```diff
@@ -0,0 +1,27 @@
...
+  // The response back from the backend to the user.
+  test.fuzz.HttpData upstream_response = 3 [(validate.rules).message.required = true];
+
+  // The stream_info for the connection.
+  test.fuzz.StreamInfo stream_info = 4 [(validate.rules).message.required = true];
+
+  // FIXME(nareddyt): The async client response. For now, just use a auto-incrementing index.
+  // There should be 4 for the following order: IMDS access token, SC Check, SC Quota, SC Report
+  // Due to possibility of retries and intermediate reports, multiply the space by 3.
+  repeated test.fuzz.HttpData sidestream_response = 5 [(validate.rules).repeated = {min_items: 0, max_items: 12}];
```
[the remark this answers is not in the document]
nareddyt: Done

src/envoy/http/service_control/filter_fuzz_test.cc
nareddyt: Unfortunately, this didn't work due to mocks vs fakes.

```
src/envoy/http/service_control/filter_fuzz_test.cc:182:41: error: no match for 'operator=' (operand types are 'testing::NiceMock<Envoy::StreamInfo::MockStreamInfo>' and 'Envoy::TestStreamInfo')
  182 |   mock_decoder_callbacks.stream_info_ = stream_info;
```

asraa: Ahhhh. Gotcha. Didn't realize `stream_info_` was a mock. Boo.

I think this is fine as is as a start, and you can iterat for speed ups. I think it'll be possible to make some of the mocks and expectations static and updating/cleaning up per-run, but it's a good task to iterate on.

src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -0,0 +1,222 @@
...
+  // Decode headers.
+  bool end_stream = false;
+  auto downstream_headers =
+      Envoy::Fuzz::fromHeaders<Envoy::Http::TestRequestHeaderMapImpl>(
+          input.downstream_request().headers());
+  if (input.downstream_request().data().size() == 0 &&
+      !input.downstream_request().has_trailers()) {
+    end_stream = true;
+  }
+  filter.decodeHeaders(downstream_headers, end_stream);
```
asraa: And then similarly, is there any reason to consider the filter status in case there's a reason to stop decoding?

nareddyt: Not for this filter, we only ever `StopIteration` or `Continue`, and that will probably never change. Our filter should support all 3 being called.

src/envoy/http/service_control/filter_fuzz_test.cc
nareddyt: Good point, could you share an example of this in Envoy? I'm not familiar with how this will speed up execution.

Yup, I can iterate on this next week. I'd like to get this merged first so OSS-Fuzz will pick it up and display some initial fuzz stats.

src/envoy/http/service_control/filter_fuzz_test.cc
qiwzhang: why do we need to make sure end_stream is correct?   Sometime, due to bug,  end_stream may not called correctly.  we still need to fuzz these data flow.

asraa: Are you saying it's possible to execute `decodeHeaders(_, true)` and then `decodeData(_, _)` somewhere?

src/envoy/http/service_control/filter_fuzz_test.cc
asraa: Sounds like a plan!

nareddyt: As discussed offline: https://github.com/envoyproxy/envoy/pull/8397

src/envoy/http/service_control/filter_fuzz_test.cc
nareddyt: @qiwzhang I think it's reasonable to expect `end_stream` is set correctly, this is part of the Envoy filter API. If it's set incorrectly, some upstream filters would not function. I don't think this will ever happen, but we can fuzz it if you believe it can.

PR #56, opened 2020-03-07
Add script to upload publicly-viewable C++ unit test coverage data

nareddyt:
Example test coverage report: https://storage.googleapis.com/esp-v2-coverage/local-run/third_party/tools/coverage/coverage_tests/index.html

Waiting for GoogleCloudPlatform/oss-test-infra#314 to auto-generate reports per PR.

### DIFF
--- a/prow/gcpproxy-coverage.sh
+++ b/prow/gcpproxy-coverage.sh
@@ -17,13 +17,11 @@
 # Presubmit script triggered by Prow.
 
 # Fail on any error.
-set -eox pipefail
+set -eo pipefail
 
 gcloud config set core/project cloudesf-testing
-gcloud auth list
-echo $CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE
-echo $GOOGLE_APPLICATION_CREDENTIALS
-gcloud auth activate-service-account --key-file=/etc/cloudesf-testing-github-prow-service-account/service-account.json
+gcloud auth activate-service-account \
+  --key-file=/etc/cloudesf-testing-github-prow-service-account/service-account.json
 
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "${ROOT}"
--- a/prow/gcpproxy-coverage.sh
+++ b/prow/gcpproxy-coverage.sh
@@ -46,12 +46,11 @@ echo '======================================================='
 PUBLIC_DIRECTORY=""
 case "${JOB_TYPE}" in
   "presubmit")
-    # Store in directory with job id.
-    # IMPORTANT: Default to a some directory to prevent messy copy on unset var.
-    PUBLIC_DIRECTORY="${PROW_JOB_ID:-tmp}"
+    # Store in directory with the SHA for each presubmit run.
+    PUBLIC_DIRECTORY=$(get_tag_name)
     ;;
   "periodic")
-    # Overwrite global directory with latest coverage.
+    # Overwrite global directory with latest coverage for all continuous runs.
     PUBLIC_DIRECTORY="latest"
     ;;
   *)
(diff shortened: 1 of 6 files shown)


prow/gcpproxy-coverage.sh
```diff
@@ -0,0 +1,71 @@
...
+
+echo '======================================================='
+echo '==================== C++ Coverage ====================='
+echo '======================================================='
+. ${ROOT}/third_party/tools/coverage/cpp_unit.sh
+
+echo '======================================================='
+echo '=================== Upload Coverage ==================='
+echo '======================================================='
+PUBLIC_DIRECTORY=""
```
qiwzhang: should we key with SHA so we will store coverage for each SHA,  
it may be useful if we want to investigate a code coverage problem?

nareddyt: Done

nareddyt: Note that I set the GCS bucket to delete entries after 60 days

prow/gcpproxy-coverage.sh
```diff
@@ -46,12 +46,11 @@ echo '======================================================='
...
   "presubmit")
-    # Store in directory with job id.
-    # IMPORTANT: Default to a some directory to prevent messy copy on unset var.
-    PUBLIC_DIRECTORY="${PROW_JOB_ID:-tmp}"
+    # Store in directory with the SHA for each presubmit run.
+    PUBLIC_DIRECTORY=$(get_tag_name)
     ;;
   "periodic")
-    # Overwrite global directory with latest coverage.
+    # Overwrite global directory with latest coverage for all continuous runs.
```
qiwzhang: it seems that we don't need code-coverage run for periodic tests since presubmit tests should have run that for each sha.

Should we disable it in period run if it is expensive( taking too long, using too much memory?  If it is cheap, just keep it.

nareddyt: I want it to run on periodic tests so that we can display our C++ unit test coverage at HEAD. You should be able to just visit a single URL and see the latest coverage at any time, similar [to this](https://storage.googleapis.com/esp-v2-coverage/local-run/third_party/tools/coverage/coverage_tests/index.html).

If we just keep presubmits, then it'll be hard to sync that to the `latest` URL since jobs can run out-of-order, 2 jobs can run at once, etc.

Shouldn't be expensive, especially with the remote cache. This might be one of the fastest jobs we have actually.

nareddyt:
Blocked by https://github.com/GoogleCloudPlatform/oss-test-infra/issues/318

prow/gcpproxy-coverage.sh
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Presubmit script triggered by Prow.
+
+# Fail on any error.
+set -eo pipefail
+
+gcloud config set core/project cloudesf-testing
+gcloud auth activate-service-account \
```
TAOXUY: Maybe login by referencing env instead of absolute path?
BTW, I looked at our other jobs, which all include the project setting step but no activate-service-account step t so maybe you can try removing no activate-service-account step

nareddyt: Sure, let's use env var.

I double checked some other jobs, it seems like they aren't using the correct service account. They are still using the default prow one as far as I can tell. Take a look at the first few lines here: https://oss-prow.knative.dev/view/gcs/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/56/ESPv2-presubmit/1237771179012395010

Let's keep the activation in this script. I'm make another PR to either make this consistent across all non-kubekins scripts or try to debug why it's not activated by default.

PR #57, opened 2020-03-07
b/140446546: add PrintOptions for transcoder

TAOXUY:
- Add flags to flip those print options
- Add an integration test

Here is the doc for those [options](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/transcoder/v2/transcoder.proto)


src/go/options/common.go
```diff
@@ -77,26 +81,28 @@ type IAMCredentialsOptions struct {
...
+		TracingOutgoingContext:                "",
+		TracingMaxNumAttributes:               32,
+		TracingMaxNumAnnotations:              32,
+		TracingMaxNumMessageEvents:            128,
+		TracingMaxNumLinks:                    128,
+		MetadataURL:                           "http://169.254.169.254/computeMetadata",
+		IamURL:                                "https://iamcredentials.googleapis.com",
+		ServiceControlCredentials:             nil,
+		BackendAuthCredentials:                nil,
+		TranscodingAlwaysPrintPrimitiveFields: false,
```
JLXIA: These are not necessary.  In go, the default value for bool is false.

tests/integration_test/transcoding_print_options/transcoding_print_options_test.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2019 Google LLC
```
JLXIA: what is the plan for re-organize these tests?

Maybe just put into grpc_test/grpc_integration_test.go

docker/generic/start_proxy.py
```diff
@@ -489,6 +489,24 @@ def make_argparser():
     Default value: Not enabled. Please provide the certificate and key files
     /etc/nginx/ssl/backend.crt and /etc/nginx/ssl/backend.key.''')
 
+    # always_print_primitive_fields.
+    parser.add_argument('--transcoding_always_print_primitive_fields',
+                        action='store_true',
+                        help=argparse.SUPPRESS)
```
qiwzhang: My understanding is SUPRESS is to hide the flag.  I don't think we need to hide it

src/go/options/common.go
qiwzhang: then, we should remove other fields with default values too., such as TracingProjectId,  and ServiceControlCredentials.

src/go/options/common.go
```diff
@@ -49,6 +49,10 @@ type CommonOptions struct {
 	ServiceControlCredentials *IAMCredentialsOptions
 	// Configures the identity used when making requests to backends.
 	BackendAuthCredentials *IAMCredentialsOptions
+
```
nareddyt: These should go in config generator options, not common options. None of these flags are needed by the ADS Bootstrapper, they are only needed by the config generator.

nareddyt: Similar for the credentials above, but it's a little late to change that.

src/go/options/common.go
TAOXUY: Just be consistent with other null flags

tests/integration_test/transcoding_print_options/transcoding_print_options_test.go
TAOXUY: working on it.

moved to grpc_transcoding_test

src/go/options/common.go
JLXIA: Try to clean up in seperate PR.

TAOXUY: Gocha.

PR #58, opened 2020-03-09
add ssl_minimum_protocol and ssl_maximum_protocol for downstream

JLXIA:
Support ssl_minimum_protocol and ssl_maximum_protocol for downstream

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -151,25 +151,36 @@ environment variable or by passing "-k" flag to this script.
         '''.format(backend=DEFAULT_BACKEND))
 
     parser.add_argument('--listener_port', default=None, type=int, help='''
-       The port to accept downstream connections.
-       It supports HTTP/1.x, HTTP/2, and gRPC connections.
-       Default is {port}'''.format(port=DEFAULT_LISTENER_PORT))
+        The port to accept downstream connections.
+        It supports HTTP/1.x, HTTP/2, and gRPC connections.
+        Default is {port}'''.format(port=DEFAULT_LISTENER_PORT))
 
     parser.add_argument('--ssl_server_cert_path', default=None, help='''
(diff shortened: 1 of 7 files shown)


src/go/util/transport_socket_test.go
```diff
@@ -152,14 +155,18 @@ func TestCreateDownstreamTransportSocket(t *testing.T) {
...
 					}
 				}
 			} `,
 		},
 		{
-			desc:    "Downstream Transport Socket for TLS, for legacy ESPv1",
-			sslPath: "/etc/nginx/ssl",
+			desc:               "Downstream Transport Socket for TLS, for legacy ESPv1",
+			sslPath:            "/etc/nginx/ssl",
+			sslMaximumProtocol: "TLSv1.3",
```
qiwzhang: can we add a test to have both min and max?

JLXIA: Done

docker/generic/start_proxy.py
```diff
@@ -540,6 +562,13 @@ def enforce_conflict_args(args):
     if len(port_flags) > 1:
         return "Multiple port flags {} are not allowed, please just use --listener_port flag".format(",".join(port_flags))
 
+    if args.ssl_minimum_protocol and args.ssl_minimum_protocol not in {"TLSv1.0", "TLSv1.1", "TLSv1.2", "TLSv1.3"}:
```
nareddyt: Suggest you use this instead: https://docs.python.org/3/library/argparse.html#choices

nareddyt: You can then also remove the values from the help text

JLXIA: Thanks, will do in following PR, since there are some other flags also need to change.

nareddyt:
Please ignore the coverage job failure until #56 is merged

JLXIA:
/ test all

PR #59, opened 2020-03-09
b/150869228: move integration tests out of individual packages

TAOXUY:
Move all the integration tests from individual package to one integration_test package, which looks more organized. 

Add t.Parallel in all the integration tests, which will creates one gorounte for each test to run.
- For integration-test-run-parallel, it takes around 4mins.
- For integration-test-run-sequential, it takes around 25mins.


tests/integration_test/grpc_streaming_test.go
```diff
@@ -29,6 +29,8 @@ import (
 )
 
 func TestGRPCStreaming(t *testing.T) {
+	//TODO(taoxuy): add t.parallel()
```
nareddyt: As discussed offline, these probably fail because they make conflicting files. Let's try to write the files with different names to `/tmp`.

TAOXUY:
> It seems that you did not rebase. Your changes have changes from others.

Done

tests/env/env.go
```diff
@@ -488,3 +488,8 @@ func (e *TestEnv) TearDown() {
 
 	glog.Infof("finish tearing down...")
 }
+
+func NewBackendAuthTestEnv(port uint16) *TestEnv {
```
nareddyt: Nit: Should this be here? I don't think `env.go` should reference a specific test name

Do we even need this function? All it does is call another function anyways with 1 hardcoded argument

nareddyt:
> * For integration-test-run-parallel, it takes around 4mins.
> * For integration-test-run-sequential, it takes around 25mins.

Thanks for verifying that parallel run works and providing metrics. This should keep local development easy.

PR #61, opened 2020-03-09
b/151105334: remove useless default option settings

TAOXUY:


TAOXUY:
ignore unready ESPv2-presubmit-coverage

PR #63, opened 2020-03-10
Validate urls in fuzz tests

nareddyt:
As discussed offline, URLs generated by config manager should be valid. We don't need to fuzz test these. This will unblock fuzzing.


nareddyt:
Changes to fuzz tests not related to flakey anthos test or incomplete coverage job. Merging so oss-fuzz will pick up this change this evening.

PR #66, opened 2020-03-11
Add badges for oss-fuzz and coverage reports

nareddyt:
Follows standard OSS practices. Clicking on the badge gives detailed logs/reports.


nareddyt:
Flakey test has nothing to do with README change

PR #67, opened 2020-03-11
Optimize Admin control

JLXIA:
1, Delete EnableAdmin Flag
2, Use AdminPort to control whether Admin is enabled or not
3, Add --status_port, -N for backward comparability with ESPv1
4, Inside ConfigManager Go, the default admin port is still 8081, to minimize the changes. Only the ApiGateway may be affected.


src/go/bootstrap/ads/bootstrap_test.go
```diff
@@ -115,7 +116,7 @@ func TestCreateBootstrapConfig(t *testing.T) {
 			desc: "bootstrap with options",
 			args: map[string]string{
 				"disable_tracing": "true",
-				"enable_admin":    "true",
+				"admin_port":      "8001",
```
nareddyt: Nit: Clean up this test to set Options instead of flags. Then you don't need to reset the flags in the following tests.

nareddyt:
@JLXIA the tests failures signal a real breakage. Please double check the Cloud Run deployment logs on Stackdriver Logging. I took a quick look, it looks like Envoy never started up. Did you change something in `start_proxy.py` that prevents starting envoy?

src/go/bootstrap/ads/bootstrap_test.go
JLXIA: I think ads is using flag to generate the config, different from static one. Should be fine.

nareddyt:
Cloud run e2e failures still look real. GKE e2e tests should pass.

nareddyt:
Please take a look at the logs on Stackdriver Logging @JLXIA. As mentioned earlier, still think something is wrong in `start_proxy.py`. Seems like Envoy is not starting up.

```
Service 34.82.185.5:80 is not ready
Command failed. Will retry in 56 seconds.
8. Executing check_grpc_service 34.82.185.5:80 false
Parsing stdin
Running tests
Writing test outputs
results {
  status {
    code: 14
    details: "failed to connect to all addresses"
  }
}
```

JLXIA:
> Please take a look at the logs on Stackdriver Logging @JLXIA. As mentioned earlier, still think something is wrong in `start_proxy.py`. Seems like Envoy is not starting up.
> 
> ```
> Service 34.82.185.5:80 is not ready
> Command failed. Will retry in 56 seconds.
> 8. Executing check_grpc_service 34.82.185.5:80 false
> Parsing stdin
> Running tests
> Writing test outputs
> results {
>   status {
>     code: 14
>     details: "failed to connect to all addresses"
>   }
> }
> ```

Thanks. The root cause is that start_proxy must pass flag with string value to configmanager, even the golang flag is int type.

PR #68, opened 2020-03-11
b/151189064: add missing logEntries

TAOXUY:
- add logEntries:
  - service_config_id
  - service_agent
  - error_cause
  - api_version(not passed by configmanager before)
  - api_name(not passed by configmanager before)
- add label:
  - version

Here is the [log](https://screenshot.googleplex.com/cUaPfz5n8SO) with this change.
Here is the [log](https://screenshot.googleplex.com/EzdKio9f6zW) before the change.

Here is the ESPv1's [log](https://screenshot.googleplex.com/8D9zvwuWadL)


src/api_proxy/service_control/request_builder.cc
```diff
@@ -923,7 +923,9 @@ constexpr char kLogFieldNameApiKey[] = "api_key";
 constexpr char kLogFieldNameApiMethod[] = "api_method";
 constexpr char kLogFieldNameApiName[] = "api_name";
 constexpr char kLogFieldNameApiVersion[] = "api_version";
+constexpr char kLogFieldNameClientIp[] = "client_ip";
 constexpr char kLogFieldNameHttpMethod[] = "http_method";
+const char kLogFieldNameErrorCause[] = "error_cause";
```
qiwzhang: can we all use constexpr

src/api_proxy/service_control/request_builder.cc
```diff
@@ -992,6 +997,9 @@ void FillLogEntry(const ReportRequestInfo& info, const std::string& name,
   (*fields)[kLogFieldNameTimestamp].set_number_value(
       (double)current_time.seconds() +
       (double)current_time.nanos() / (double)1000000000.0);
+    (*fields)[kLogFieldNameConfigId].set_string_value(config_id);
+      (*fields)[kLogFieldNameServiceAgent].set_string_value(
```
qiwzhang: wrong indent

tests/utils/service_control_utils_test.go
```diff
@@ -2034,6 +2105,7 @@ func TestCreateAggregateReport(t *testing.T) {
 		URL:               "/shelves",
 		ApiMethod:         "ListShelves",
 		ApiKey:            "api-key",
+		ApiName: "API-NAME",
```
qiwzhang: wrong format?

TAOXUY:
PTAL

tests/integration_test/cors_integration_test.go
```diff
@@ -537,6 +537,7 @@ func TestServiceControlRequestWithAllowCors(t *testing.T) {
 					URL:               "/bookstore/shelves?key=api-key",
 					ApiKey:            "api-key",
 					ApiMethod:         "1.echo_api_endpoints_cloudesf_testing_cloud_goog.CorsShelves",
+					ApiName:           "1.echo_api_endpoints_cloudesf_testing_cloud_goog",
```
nareddyt: Check API version here too?

nareddyt: There's a few more tests where this isn't checked in other integration test files, can you also check there? Or is it not needed?

TAOXUY: Those methods don't have apiVersion because they aren't under Apis.Methods(only Apis define the version) but generated by  HttpRules.

PR #70, opened 2020-03-12
Update envoy to 2020-03-12

nareddyt:
This is needed to improve tech debt related to fuzz failures.

Breaking change in the header map interface: https://github.com/envoyproxy/envoy/pull/10252


nareddyt:
@JLXIA please confirm if it's OK to update Envoy to the latest commit (from a security / embargo standpoint).

PR #71, opened 2020-03-12
Handle null pointer for running envoy in a named pipe

nareddyt:
`Address` can either be a IP address or a named pipe. We try to read from a `nullptr` if the address is a named pipe. Even though we will never run into this problem with the current deployment model of ESPv2, we should still handle this case for other open-source users and simpler fuzzing.

Added reproducer test-case.

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21174

### DIFF
--- a/api/envoy/http/service_control/config.proto
+++ b/api/envoy/http/service_control/config.proto
@@ -69,10 +69,14 @@ message Service {
   string backend_protocol = 5;
 
   // The array of request headers demanded to be logged
-  repeated string log_request_headers = 6 [(validate.rules).repeated.items.string.well_known_regex = HTTP_HEADER_NAME];
+  repeated string log_request_headers = 6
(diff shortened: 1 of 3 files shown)


api/envoy/http/service_control/config.proto
```diff
@@ -69,10 +69,14 @@ message Service {
   string backend_protocol = 5;
 
   // The array of request headers demanded to be logged
-  repeated string log_request_headers = 6 [(validate.rules).repeated.items.string.well_known_regex = HTTP_HEADER_NAME];
+  repeated string log_request_headers = 6
+      [(validate.rules).repeated .items.string.well_known_regex =
```
nareddyt: Ignore diff here, this is from `clang-format`.

PR #73, opened 2020-03-13
Validate URIs in envoy config

nareddyt:
Bug type: Tech Debt

Use new `protoc-gen-validate` well-known type to validate all URIs don't cause assertion failures in Envoy. This makes fuzz testing a lot simpler and maintains consistency with upstream Envoy config validation.

By setting `strict = false`, this validation just prevents `\r`, `\n`, and `\0` characters in URIs. Note this validation is not compliant with RFC 7230 (all other characters are allowed). This is not a breaking change, as `\r\n\0` characters cause envoy assertion failures anyways.

Added a reproducer test-case for service control URI.

Ref: https://github.com/envoyproxy/protoc-gen-validate/pull/323
Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21172

### DIFF
--- a/api/envoy/http/backend_routing/config.proto
+++ b/api/envoy/http/backend_routing/config.proto
@@ -32,7 +32,6 @@ message BackendRoutingRule {
   // Example for APPEND_PATH_TO_ADDRESS:
   //     https://my-project-id.appspot.com
   string path_prefix = 3 [(validate.rules).string = {
-    min_bytes: 1,
     well_known_regex: HTTP_HEADER_VALUE,
     strict: false
   }];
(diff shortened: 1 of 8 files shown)


nareddyt:
FYI coverage job is failing due to OOM. Applied the following default resource requests and limits to the `test-pods` namespace. Will retry now.

```
apiVersion: v1
kind: LimitRange
metadata:
  name: mem-limit-range
spec:
  limits:
  - default:
      cpu: 4000m
      memory: 20Gi
    defaultRequest:
      cpu: 2000m
      memory: 4Gi
    type: Container
```

PR #74, opened 2020-03-13
Catch EnvoyException when fuzzing stream info

nareddyt:
Bug type: Tech debt

Fuzzing the `StreamInfo` can result in an `EnvoyException` being thrown. This is not interesting, we should catch this exception. This will reduce false positives in OSS Fuzz.

Added reproducer test-case where named pipe `address` is not valid.

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21181


src/envoy/http/service_control/filter_fuzz_test.cc
```diff
@@ -177,14 +177,15 @@ DEFINE_PROTO_FUZZER(
...
 
-  // Fuzz the stream info.
-  TestStreamInfo stream_info = Envoy::Fuzz::fromStreamInfo(input.stream_info());
-  EXPECT_CALL(mock_decoder_callbacks, streamInfo())
-      .WillRepeatedly(ReturnRef(stream_info));
-  EXPECT_CALL(mock_encoder_callbacks, streamInfo())
-      .WillRepeatedly(ReturnRef(stream_info));
-
   try {
+    // Fuzz the stream info.
```
qiwzhang: Then doTest() should be outside of exception try catch block.  Envoy only throw exception on config flow,  not in data process flow.

nareddyt: Done. Still kept it inside the `try/catch` scope to simplify filter setup.

PR #76, opened 2020-03-14
b/151274198: refactor service_config_fetcher

TAOXUY:
- Decouple service_config_fecter into a new struct with following changes:
    - do the first time fetch in NewServiceConfigFetcher and hide fetchConfig method
    - move the periodic fetch logic into SetFetchConfigTimer
    - let both configManager and serviceCOnfigFetcher keep a pointer of curServiceConfig, based on which they can get the curConfigId. 

- Getting rollout related code will be replaced by calling service control report later.


src/go/configmanager/config_manager.go
```diff
@@ -45,24 +48,21 @@ var (
...
 type ConfigManager struct {
-	serviceName        string
-	serviceInfo        *configinfo.ServiceInfo
-	envoyConfigOptions options.ConfigGeneratorOptions
-	curRolloutID       string
-	curConfigID        string
+	serviceName               string
+	serviceInfo               *configinfo.ServiceInfo
+	envoyConfigOptions        options.ConfigGeneratorOptions
+	configIdServiceConfigFile string
```
qiwzhang: How about keep configId here?  if from fetcher,  update the config_id here too.

src/go/configmanager/config_manager.go
```diff
@@ -154,64 +146,52 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
...
-	m.curConfigID = serviceConfig.GetId()
+	m.configIdServiceConfigFile = serviceConfig.GetId()
 
 	return m.applyServiceConfig(serviceConfig)
 }
 
 func (m *ConfigManager) applyServiceConfig(serviceConfig *confpb.Service) error {
 	var err error
-	m.serviceInfo, err = configinfo.NewServiceInfoFromServiceConfig(serviceConfig, m.curConfigID, m.envoyConfigOptions)
+	m.serviceInfo, err = configinfo.NewServiceInfoFromServiceConfig(serviceConfig, m.curConfigId(), m.envoyConfigOptions)
```
qiwzhang: serviceConfig should have id() field which is the config_id.  not need to call m.curConfigId() to get it.

src/go/configmanager/config_manager.go
```diff
@@ -15,20 +15,23 @@
...
+	"bytes"
 	"flag"
 	"fmt"
-	"net/http"
+	"io/ioutil"
 	"time"
 
 	"github.com/GoogleCloudPlatform/esp-v2/src/go/commonflags"
 	"github.com/GoogleCloudPlatform/esp-v2/src/go/configinfo"
+
```
nareddyt: Nit: Remove new line

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/metadata"
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/proto"
+
+	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
+	smpb "google.golang.org/genproto/googleapis/api/servicemanagement/v1"
+)
+
+type ServiceConfigFetcher struct {
```
nareddyt: Please use `git mv` to preserve the commit history for this file. It'll also make the diff easier to read.

src/go/configmanager/config_manager.go
nareddyt: Is there a specific reason for this Wayne? I think it makes sense to keep all service config related items decoupled from ConfigManager. I like the current implementation of placing `configId` in the `ServiceConfigFetcher`

nareddyt: `configIdServiceConfigFile` is a little confusing, can you find a better name for it?

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
+		return nil, fmt.Errorf("http call to %s returns not 200 OK: %v", path, resp.Status)
+	}
+	return resp, nil
+}
+
+// Fetch the service config by given configId. If configId is empty, try to
+// fetch the latest service config,.
+func (scf *ServiceConfigFetcher) FetchConfig(configId string) (*confpb.Service, error) {
```
nareddyt: Could you move the public function above all the private ones?

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+	if configId != "" {
+		scf.curConfigId = configId
+		token, _, err := scf.accessToken()
+		if err != nil {
+			return nil, fmt.Errorf("fail to get access token: %v", err)
+		}
+		return scf.callServiceManagement(util.FetchConfigURL(*flags.ServiceManagementURL, scf.serviceName, configId), token)
```
nareddyt: I recommend removing all the flags here, change them to options now that ServiceConfigFetcher is a struct. Otherwise it's hard to read; I don't even know where this flag is defined.

src/go/serviceconfig/service_config_fetcher.go
nareddyt: I will finish reviewing once the diff is easier to read

src/go/configmanager/config_manager.go
```diff
@@ -154,64 +146,52 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
 				return nil, fmt.Errorf("service config id is not specified, required on a non-gcp deployment")
 			}
 		}
-		m.curConfigID = configID
-		if err := m.updateSnapshot(); err != nil {
-			return nil, err
-		}
+		serviceConfig, err = m.serviceConfigFetcher.FetchConfig(configID)
```
nareddyt: Why don't you move `configID` into constructor `NewServiceConfigFetcher`. It's only ever needed once.

src/go/configmanager/config_manager.go
qiwzhang: We need config_id from service_config_path flow too.  I just want two flows to store it in the same place.

src/go/configmanager/config_manager.go
```diff
@@ -51,10 +51,10 @@ var (
...
 // TODO(jilinxia): handles multi service name.
 type ConfigManager struct {
-	serviceName               string
-	serviceInfo               *configinfo.ServiceInfo
-	envoyConfigOptions        options.ConfigGeneratorOptions
-	configIdServiceConfigFile string
+	serviceName                   string
+	serviceInfo                   *configinfo.ServiceInfo
+	envoyConfigOptions            options.ConfigGeneratorOptions
+	configIdFromServiceConfigFile string
```
qiwzhang: Hmm,  I am wondering if we can define an interface to share between ServiceConfigFile and ServiceConfigFetcher.  ConfigMgr only use the interface.

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -0,0 +1,252 @@
...
+					RootCAs: caCertPool,
+				},
+			},
+			Timeout: opts.HttpRequestTimeout,
+		},
+		serviceName: serviceName,
+		mf:          mf,
+		opts:        opts,
+	}
+	_, _ = scf.fetchConfig(configId)
```
nareddyt: Be careful about this, this makes it harder to unit test. Ideally, the constructor shouldn't call another function to make external requests like this. Instead, `configId` should be stored in the `struct` and the client should call `FetchConfig`.

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -0,0 +1,252 @@
...
+	if scf.curServiceConfig == nil {
+		return ""
+	}
+	return scf.curServiceConfig.Id
+}
+
+// Fetch the service config by given configId. If configId is empty, try to
+// fetch the latest service config,.
+func (scf *ServiceConfigFetcher) fetchConfig(configId string) (*confpb.Service, error) {
+	_fetchConfig := func(configId string) (*confpb.Service, error) {
```
nareddyt: What is the point of this internal function? It's just called like a normal function right after, so I don't see why it's needed.

src/go/configmanager/config_manager.go
```diff
@@ -45,24 +46,22 @@ var (
...
 type ConfigManager struct {
-	serviceName        string
-	serviceInfo        *configinfo.ServiceInfo
-	envoyConfigOptions options.ConfigGeneratorOptions
-	curRolloutID       string
-	curConfigID        string
+	serviceName                   string
+	serviceInfo                   *configinfo.ServiceInfo
+	envoyConfigOptions            options.ConfigGeneratorOptions
+	configIdFromServiceConfigFile string
```
nareddyt: Where is this variable used? Something seems wrong, this should definitely be set somewhere right?

src/go/configmanager/config_manager.go
```diff
@@ -121,31 +120,18 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
...
 	} else {
 		// rollout strategy is fixed mode
-		configID := *ServiceConfigID
-		if configID == "" {
+		configId := *ServiceConfigID
+		if configId == "" {
 			if checkMetadata && mf != nil {
-				configID, err = mf.FetchConfigId()
-				if configID == "" || err != nil {
+				configId, err = mf.FetchConfigId()
```
nareddyt: I don't fully understand, where is Metadata Fetcher getting the config id from? Is it a custom attribute attached to the VM? cc @qiwzhang, is this for ESPv1 compatibility?

src/go/configmanager/config_manager.go
```diff
@@ -121,31 +119,18 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
...
+		if m.serviceConfigFetcher, err = sc.NewServiceConfigFetcher(mf, opts, m.serviceName, ""); err != nil {
+			return nil, fmt.Errorf(`failed to create https client to call ServiceManagement service, got error: %v`, err)
+
 		}
 	} else {
 		// rollout strategy is fixed mode
-		configID := *ServiceConfigID
-		if configID == "" {
+		configId := *ServiceConfigID
+		if configId == "" {
```
JLXIA: nit:  from Golang style, configID is preferred.

src/go/serviceconfig/service_config_fetcher.go
TAOXUY: It is implementation of fetching service config and then decorated with assigning the result to the object.

src/go/configmanager/config_manager.go
```diff
@@ -154,64 +139,52 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
 				return nil, fmt.Errorf("service config id is not specified, required on a non-gcp deployment")
 			}
 		}
-		m.curConfigID = configID
-		if err := m.updateSnapshot(); err != nil {
-			return nil, err
+
+		if m.serviceConfigFetcher, err = sc.NewServiceConfigFetcher(mf, opts, m.serviceName, configId); err != nil {
```
JLXIA: initialize serviceConfigFetcher in single place?

src/go/serviceconfig/service_config_fetcher.go
JLXIA: +1 
what is the purpose for this?

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+		}
+	}()
+}
+
+func (scf *ServiceConfigFetcher) CurRolloutId() string {
+	return scf.curRolloutId
+}
+
+func (scf *ServiceConfigFetcher) curConfigId() string {
```
JLXIA: No need have two methods, one for public, one for private.

src/go/util/url_util.go
```diff
@@ -142,3 +142,22 @@ func ExtraAddressFromURI(jwksUri string) (string, error) {
 	}
 	return fmt.Sprintf("%s:%v", hostname, port), nil
 }
+
+const (
```
JLXIA: shouldn't these in util.go?

src/go/serviceconfig/service_config_fetcher.go
TAOXUY: Do you mean CurRolloutId()? They are different.

src/go/util/url_util.go
TAOXUY: They are private const and only used by FetchConfigURL/FetchRolloutsURL

src/go/configmanager/config_manager.go
TAOXUY: We already use Id, Uri, Url and Http in lots of place so I suggest we still keep following this convention.

PR #77, opened 2020-03-16
Support strict transport security

JLXIA:
1.  when --enbale_strict_transport_security is enabled, add strict_transport_security in response header.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -177,6 +177,11 @@ environment variable or by passing "-k" flag to this script.
         help=''' Maximum TLS protocol version for client side connection.
         Please refer to https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/cert.proto#common-tls-configuration.
         ''')
+    
+    parser.add_argument('--enable_strict_transport_security', action='store_true',
+        help='''Enable HSTS (HTTP Strict Transport Security). "Strict-Transport-Security" response header
+        with value "max-age=31536000; includeSubdomains;" is added for all responses from local backend.
+        Not valid for remote backends.''')
 
     parser.add_argument('-z', '--healthz', default=None, help='''Define a
         health checking endpoint on the same ports as the application backend.
@@ -649,6 +654,9 @@ def gen_proxy_config(args):
         args.ssl_protocols.sort()
         proxy_conf.extend(["--ssl_minimum_protocol", args.ssl_protocols[0]])
         proxy_conf.extend(["--ssl_maximum_protocol", args.ssl_protocols[-1]])
+
+    if args.enable_strict_transport_security:
+            proxy_conf.append("--enable_strict_transport_security")
 
     if args.service:
         proxy_conf.extend(["--service", args.service])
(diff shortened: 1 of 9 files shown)


src/go/configgenerator/route_generator_test.go
```diff
@@ -21,13 +21,84 @@ import (
...
+                             "virtualHosts": [
+                                 {
+                                     "domains": [
+                                         "*"
+                                     ],
+                                     "name": "backend",
+                                     "routes": [
+                                         {
+                                             "match": {
+                                                 "prefix": "/"
```
qiwzhang: why not making this for all backend?  why just remote backend?  As I see it,  HSTS is per server flag,  proxy is one listener is one server, this flag should apply to all backend the proxy is supporting.

I understand that it is only useful for side-car (local host) backend, which usually is connected using http.  so Backend will not respond with HSTS header. Remote backend usually is https, it is up to each https backend to response such header. 

But I see such flag is a per-server flag, it should apply to ESPv2 proxy as a whole.
It also help with Cloud Run for anthos which is using http remote backend.

tests/integration_test/transport_security_test.go
```diff
@@ -214,3 +214,62 @@ func TestHttpsClients(t *testing.T) {
...
+		{
+			desc:           "Succcess for HTTP2 client with HSTS",
+			httpsVersion:   2,
+			certPath:       platform.GetFilePath(platform.ServerCert),
+			wantHSTSHeader: "max-age=31536000; includeSubdomains",
+		},
+	}
+
+	for _, tc := range testData {
+		url := fmt.Sprintf("https://localhost:%v/simpleget?key=api-key", s.Ports().ListenerPort)
```
qiwzhang: Could we have a test, that client trying to downgrade https to http(I don't know how that work, but as my understanding this is what HSTS is trying to prevent).  then test it, without HSTS, it is successful, it will fail such downgrade it it is enabled.

nareddyt: HSTS (like same-origin policy) is a protocol that **browsers** must adhere to. I don't think we can test this in our integration tests. All Envoy is doing is adding a header in the responses, it's the browser that prevents using HTTP.

docker/generic/start_proxy.py
```diff
@@ -177,6 +177,11 @@ def make_argparser():
         help=''' Maximum TLS protocol version for client side connection.
         Please refer to https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/cert.proto#common-tls-configuration.
         ''')
+    
+    parser.add_argument('--enable_strict_transport_security', action='store_true',
```
nareddyt: Can we validate that SSL is enabled (`--ssl_server_cert_path` is used) for this setting to apply? Otherwise, this flag would cause conflicting behavior. Not sure if that maintains backwards compatibility.

tests/integration_test/transport_security_test.go
JLXIA: Agree, we just need to make sure the header is returned.

src/go/configgenerator/route_generator_test.go
JLXIA: Added.

docker/generic/start_proxy.py
JLXIA: Will follow up to see whether it is needed or not.

PR #79, opened 2020-03-17
b/150729554: enable ignoredQueryParameters and ignoredUnknownQueryPar…

TAOXUY:
…ameters


src/go/configgenerator/listener_generator.go
```diff
@@ -639,27 +639,16 @@ func makeTranscoderFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
 				DescriptorSet: &transcoderpb.GrpcJsonTranscoder_ProtoDescriptorBin{
 					ProtoDescriptorBin: configContent,
 				},
-				AutoMapping:       true,
-				ConvertGrpcStatus: true,
+				AutoMapping:                  true,
+				ConvertGrpcStatus:            true,
+				IgnoredQueryParameters:       transcodingIgnoreQueryParameters(serviceInfo),
+				IgnoreUnknownQueryParameters: serviceInfo.Options.TranscodingIgnoreUnknownQueryParameters,
```
qiwzhang: If IgnoreUnknownQueryParameters is true,  we don't need to set another one.  It doesn't make sense to set both.

TAOXUY: Good point, will add exception in start_esp.py

src/go/configgenerator/listener_generator.go
```diff
@@ -799,3 +788,38 @@ func makeRouterFilter(opts options.ConfigGeneratorOptions) *hcmpb.HttpFilter {
 	}
 	return routerFilter
 }
+
+func transcodingIgnoreQueryParameters(serviceInfo *sc.ServiceInfo) []string {
```
JLXIA: move this logic into serviceinfo.go, and considering to consolidate TranscoderIgnoredJwtQueryParams, TranscoderIgnoredApiKeyQueryParams, transcodingIgnoreQueryParameters.

1
```diff
@@ -0,0 +1,2122 @@
+..usage: python3 -m unittest [-h] [-s SERVICE] [-v VERSION]
```
nareddyt: Remove file

TAOXUY:
PTAL

PR #81, opened 2020-03-17
Align `--disable_cloud_trace_auto_sampling` with ESPv1

nareddyt:
Note that ESPv1 does not have `--tracing_sample_rate`, but ESPv2 does.
So when `--disable_cloud_trace_auto_sampling`, just override the
tracing sample rate to 0.

### DIFF
--- a/tests/start_proxy/start_proxy_test.py
+++ b/tests/start_proxy/start_proxy_test.py
@@ -19,7 +19,7 @@ import os, inspect
 
 currentdir = os.path.dirname(
     os.path.abspath(inspect.getfile(inspect.currentframe())))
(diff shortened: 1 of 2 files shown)


tests/start_proxy/start_proxy_test.py
```diff
@@ -19,7 +19,7 @@
 
 currentdir = os.path.dirname(
     os.path.abspath(inspect.getfile(inspect.currentframe())))
-sys.path.append(currentdir + "/../../docker/generic")
+sys.path.insert(0, currentdir + "/../../docker/generic")
```
nareddyt: I couldn't get the tests to run on my machine, SO suggests to replace the first path instead of appending. This works now, should work in CI.

PR #82, opened 2020-03-18
b/151344434:  Suppress envoy debug headers when `--enable_debug=false`

nareddyt:
Option in config manager already exists, just set it appropriately via `start_proxy.py`. By default, suppress debug headers.

Testing done:
- Adjust tests to align with default option.
- Add unit tests for enable debug.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -464,7 +464,7 @@ environment variable or by passing "-k" flag to this script.
         Enables a variety of debug features in both Config Manager and Envoy, such as:
         - Debug level per-request application logs in Envoy
         - Debug level service configuration logs in Config Manager
-        - Admin interface in Envoy
+        - Debug HTTP response headers
         ''')
 
     # Start Deprecated Flags Section
@@ -806,6 +806,10 @@ def gen_proxy_config(args):
         proxy_conf.extend(["--service_account_key", args.service_account_key])
     if args.non_gcp:
         proxy_conf.append("--non_gcp")
+
+    if args.enable_debug:
+        proxy_conf.append("--suppress_envoy_headers=false")
+
     return proxy_conf
 
 def gen_envoy_args(args):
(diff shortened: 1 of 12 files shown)


docker/generic/start_proxy.py
```diff
@@ -464,7 +464,7 @@ def make_argparser():
         Enables a variety of debug features in both Config Manager and Envoy, such as:
         - Debug level per-request application logs in Envoy
         - Debug level service configuration logs in Config Manager
-        - Admin interface in Envoy
+        - Debug HTTP response headers
```
qiwzhang: do we not adding Admin port with enable_debug=true?   I think we should

docker/generic/start_proxy.py
```diff
@@ -806,6 +806,10 @@ def gen_proxy_config(args):
         proxy_conf.extend(["--service_account_key", args.service_account_key])
     if args.non_gcp:
         proxy_conf.append("--non_gcp")
+
+    if args.enable_debug:
+        proxy_conf.append("--suppress_envoy_headers=false")
```
qiwzhang: I am wondering if we should expose this flag here.   Maybe some users may find it useful.

docker/generic/start_proxy.py
nareddyt: We were previously, but @JLXIA removed this in #67. I just updated the docs to match this.

One difficulty to add this behavior back is: we need to define a default port to enable admin interface (instead of user specifying port).

docker/generic/start_proxy.py
nareddyt: @JLXIA PTAL, should we expose this flag to the user?

docker/generic/start_proxy.py
qiwzhang: hmm,  maybe in a separate PR to allow user to enable admin page. For GKE, it will be useful for some users

docker/generic/start_proxy.py
qiwzhang: Well, maybe not now. we can do that if there are users required it.

PR #83, opened 2020-03-19
b/151274198: fetch config id from servicecontrol

TAOXUY:
Get serivce_rollout_id from service control report
- add service_rollout_fetcher for sending empty Report to service control
- add config_id_detector to  find the latest rollout id.
- move callGoogleapis to util and shared by config_id_detector and service_config_fetcher


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	cloud.google.com/go v0.41.1-0.20190709211438-47e9997a900f
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/census-instrumentation/opencensus-proto v0.2.1
+	github.com/envoyproxy/envoy v1.13.1 // indirect
```
nareddyt: I don't think this is needed. Envoy repo does not have any golang protos, `go-control-plane` does?

src/go/configmanager/config_manager_test.go
```diff
@@ -46,6 +46,7 @@ import (
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
 	annotationspb "google.golang.org/genproto/googleapis/api/annotations"
 	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
+	googleapisc "google.golang.org/genproto/googleapis/api/servicecontrol/v1"
```
nareddyt: should end with `pb`

scpb?

src/go/serviceconfig/config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+
+type ConfigIdFetcher struct {
+	serviceName       string
+	serviceControlUrl string
+	client            http.Client
+	accessToken       func() (string, time.Duration, error)
+}
+
+func NewConfigIdFetcher(serviceName, serviceControlUrl string, client http.Client,
+	accessToken func() (string, time.Duration, error)) *ConfigIdFetcher {
```
nareddyt: Consider moving `(string, time.Duration, error)` into its own struct. Not sure what the Effective Go says about this

nareddyt: Or type alias the function at least

```
type GetAccessTokenFunc func() (string, time.Duration, error)
```

Similar to `token_subscriber.h` in our c++ code

src/go/serviceconfig/config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+	return &ConfigIdFetcher{
+		serviceName:       serviceName,
+		serviceControlUrl: serviceControlUrl,
+		client:            client,
+		accessToken:       accessToken,
+	}
+
+}
+
+func (cif *ConfigIdFetcher) fetchNewConfigId() (string, error) {
```
nareddyt: Nit: `c` instead of `cif`

https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

src/go/serviceconfig/config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+	token, _, err := cif.accessToken()
+	if err != nil {
+		return "", fmt.Errorf("fail to get access token: %v", err)
+	}
+
+	reportResponse, err := cif.callServiceControl(util.FetchConfigIdURL(cif.serviceControlUrl, cif.serviceName), token)
```
nareddyt: Suggest you move the inner function to a new var

src/go/serviceconfig/config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (cif *ConfigIdFetcher) callServiceControl(path, token string) (*scpb.ReportResponse, error) {
+	respBytes, err := util.CallWithAccessToken(cif.client, util.POST, path, token)
+	if err != nil {
+		return nil, err
+	}
+
+	reportResponse := new(scpb.ReportResponse)
+	if err := proto.Unmarshal(respBytes, reportResponse); err != nil {
+		return nil, fmt.Errorf("fail to unmarshal ListServiceRolloutsResponse: %s", err)
```
nareddyt: Fix error message, mismatching types

src/go/serviceconfig/config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return reportResponse.ServiceConfigId, nil
+}
+
+func (cif *ConfigIdFetcher) callServiceControl(path, token string) (*scpb.ReportResponse, error) {
+	respBytes, err := util.CallWithAccessToken(cif.client, util.POST, path, token)
+	if err != nil {
```
nareddyt: Combine these 2 lines
```
if err := ...; err != nil { ... }
```

src/go/serviceconfig/config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	"github.com/golang/protobuf/proto"
+	scpb "google.golang.org/genproto/googleapis/api/servicecontrol/v1"
+)
+
+type ConfigIdFetcher struct {
```
nareddyt: I prefer `ServiceConfigIdFetcher`. Config sounds too general, we have many configs in our code :)

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -19,32 +19,31 @@ import (
...
 type ServiceConfigFetcher struct {
 	serviceName         string
 	checkRolloutsTicker *time.Ticker
 	client              http.Client
-	mf                  *metadata.MetadataFetcher
-	opts                options.ConfigGeneratorOptions
+	opts                *options.ConfigGeneratorOptions
+
+	accessToken func() (string, time.Duration, error)
+	newConfigId func() (string, error)
```
nareddyt: Same comments about type aliasing the function(s)

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -19,32 +19,31 @@ import (
...
 
 type ServiceConfigFetcher struct {
 	serviceName         string
 	checkRolloutsTicker *time.Ticker
 	client              http.Client
-	mf                  *metadata.MetadataFetcher
-	opts                options.ConfigGeneratorOptions
+	opts                *options.ConfigGeneratorOptions
+
+	accessToken func() (string, time.Duration, error)
```
nareddyt: Would prefer if name clarifies the need for an async call:

```
accessTokenGetter? accessTokenFunc?
configIdGetter? configIdFunc?
```

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -61,37 +60,61 @@ func NewServiceConfigFetcher(mf *metadata.MetadataFetcher, opts options.ConfigGe
...
+
+	configIdFetcher := NewConfigIdFetcher(serviceName, opts.ServiceControlURL,
+		scf.client, func() (string, time.Duration, error) { return scf.accessToken() })
+	scf.newConfigId = func() (string, error) {
+		return configIdFetcher.fetchNewConfigId()
+	}
+
+	scf.accessToken = func() (string, time.Duration, error) {
+		if accessTokenFromImds == nil && scf.opts.ServiceAccountKey == "" {
+			return "", 0, fmt.Errorf("If --non_gcp is specified, --service_account_key has to be specified.")
```
nareddyt: Unclear how this case is `--non_gcp`, please clarify with comment or try to make the if check clearer.

src/go/util/request.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+func CallWithAccessToken(client http.Client, method, path, token string) ([]byte, error) {
+	req, _ := http.NewRequest(method, path, nil)
+	req.Header.Add("Authorization", "Bearer "+token)
```
nareddyt: `make format`?

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -61,37 +60,61 @@ func NewServiceConfigFetcher(mf *metadata.MetadataFetcher, opts options.ConfigGe
...
+		}
+		if scf.opts.ServiceAccountKey != "" {
+			return util.GenerateAccessTokenFromFile(scf.opts.ServiceAccountKey)
+		}
+		return accessTokenFromImds()
```
nareddyt: I don't understand: Why does this call `accessTokenFromImds()`, but in this line below you call `mf.FetchAccessToken()`.

https://github.com/GoogleCloudPlatform/esp-v2/pull/83/files#diff-f2a296fe272e72d0390a8830217a156aR141

Shouldn't both of these be using the same access token?

Additionally, this looks like you fetch the access token every time you make a call? Could you clarify how the access token is fetched / cached?

src/go/serviceconfig/config_id_fetcher.go
TAOXUY: In that case, respBytes need b declared in advance, which seems no much benefit.

src/go/serviceconfig/service_config_fetcher.go
TAOXUY: Per https://golang.org/doc/effective_go.html#Getters
Getter is just the name of variable

src/go/serviceconfig/service_config_fetcher.go
TAOXUY: - It is just a wrapper of mf.FetchAccessToken()
- It is cached in mf.FetchAccessToken has cache mechanism but I admit the name looks like just simply fetching.

src/go/configmanager/config_manager_test.go
```diff
@@ -46,6 +46,7 @@ import (
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
 	annotationspb "google.golang.org/genproto/googleapis/api/annotations"
 	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
+	googleapiscpb "google.golang.org/genproto/googleapis/api/servicecontrol/v1"
```
nareddyt: `scpb`? to match the other imports that are also googleapis?

src/go/configmanager/config_manager.go
```diff
@@ -138,7 +137,13 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
 
 	}
 
-	if m.serviceConfigFetcher, err = sc.NewServiceConfigFetcher(mf, opts, m.serviceName); err != nil {
+	var accessTokenFromImds sc.GetAccessTokenFunc
+	if mf != nil {
+		accessTokenFromImds = func() (string, time.Duration, error) { return mf.FetchAccessToken() }
+	}
```
nareddyt: What happens if it's nil? Wouldn't this cause a panic in line 146 when you pass a function without a behavior?

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -19,39 +19,41 @@ import (
...
+	accessToken GetAccessTokenFunc
+	newConfigId GetNewConfigIdFunc
 
 	curServiceConfig *confpb.Service
-	curRolloutId     string
 }
 
-func NewServiceConfigFetcher(mf *metadata.MetadataFetcher, opts options.ConfigGeneratorOptions, serviceName string) (*ServiceConfigFetcher, error) {
+func NewServiceConfigFetcher(opts *options.ConfigGeneratorOptions,
+	serviceName string, accessTokenFromImds func() (string, time.Duration, error)) (*ServiceConfigFetcher, error) {
```
nareddyt: The reason I asked to type alias the functions was so you can use the type aliases here as well to make it clearer.

```
func NewServiceConfigFetcher(opts *options.ConfigGeneratorOptions,
	serviceName string, accessTokenFromImds GetAccessTokenFunc) (*ServiceConfigFetcher, error) {
```

See if they can be used elsewhere as well. It makes it a lot easier to understand.

src/go/serviceconfig/service_config_fetcher_test.go
```diff
@@ -15,34 +15,138 @@
...
+	}
+
+	serviceManagementServer := initServiceManagementForTestServbiceConfigFetcherFetchConfig(t, serviceConfig, serviceName)
 	opts := options.DefaultConfigGeneratorOptions()
-	timeout := 1 * time.Second
-	opts.HttpRequestTimeout = timeout
+	opts.ServiceManagementURL = serviceManagementServer.URL
 
-	serviceConfigFetcher, err := NewServiceConfigFetcher(nil, opts, "service-name")
+	scf, err := NewServiceConfigFetcher(&opts, serviceName, func() (string, time.Duration, error) { return "access-token", time.Duration(60), nil })
```
nareddyt: Strongly recommend here (and all other places) to define the func as a named var in the line above and pass the var. It's really difficult to read on github otherwise.

src/go/serviceconfig/service_config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	"github.com/golang/protobuf/proto"
+	scpb "google.golang.org/genproto/googleapis/api/servicecontrol/v1"
+)
+
+type ServiceConfigIdFetcher struct {
+	serviceName       string
+	serviceControlUrl string
+	client            http.Client
+	accessToken       func() (string, time.Duration, error)
```
nareddyt: Please use the type aliases throughout the code. Consider moving them to utils/ if it creates cyclic dependencies.

src/go/configmanager/config_manager_test.go
nareddyt: Please fix this across all files, not only this one

TAOXUY: I use this because scpb has been occupied by `scpb "github.com/GoogleCloudPlatform/esp-v2/src/go/proto/api/envoy/http/service_control"` in L37
Change to servicecontrolpb; hope it can make you feel better.

TAOXUY:
@nareddyt, DONE. Thanks for your comments.

src/go/serviceconfig/service_config_id_fetcher.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return "", fmt.Errorf("fail to get access token: %v", err)
+	}
+
+	reportResponse, err := c.callServiceControl(util.FetchConfigIdURL(c.serviceControlUrl, c.serviceName), token)
+	if err != nil {
+		return "", err
+	}
+
+	return reportResponse.ServiceConfigId, nil
```
qiwzhang: We should use rollout ID.  The key is to check rolloutID, not serviceConfigId.

This class should be called service_rollout_id_fetcher

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -61,57 +60,72 @@ func NewServiceConfigFetcher(mf *metadata.MetadataFetcher, opts options.ConfigGe
...
 			}
-			return scf.callServiceManagement(util.FetchConfigURL(scf.opts.ServiceManagementURL, scf.serviceName, configId), token)
+
+			return s.callServiceManagement(util.FetchConfigURL(s.opts.ServiceManagementURL, s.serviceName, configId), token)
 		}
 
-		glog.Infof("check new rollouts for service %v", scf.serviceName)
-		newRolloutId, newConfigId, err := scf.loadConfigFromRollouts(scf.serviceName, scf.curRolloutId, scf.curConfigId())
+		glog.Infof("check new config id for service %v", s.serviceName)
+		newConfigId, err := s.newConfigId()
```
qiwzhang: Did not quite follow the logic here. I could not find where newConfigId() is defined.

qiwzhang: Basically, what we need to do is:
1) on very minute timer, we call empty report to get latest Rollout ID. 
2) if rollout ID is changed,  call ServiceManaement to get Latest rollout info.   A rollout may have multiple service config_id.  ESPv1, we support traffic percentage rollout.  But ESPv2, we don't.  We just get the first largest pecentage config_id.  That is the config_id we need to use.
E.g.  a rollout may have 80% for id1 and 20% for id2.  we just use id1 since it is 80%.

3) use that config_id to get the service config.

TAOXUY:
PTAL @qiwzhang @nareddyt  @JLXIA

src/go/configmanager/config_manager.go
```diff
@@ -138,7 +137,21 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
 
 	}
 
-	if m.serviceConfigFetcher, err = sc.NewServiceConfigFetcher(mf, opts, m.serviceName); err != nil {
+	accessToken := func() (string, time.Duration, error) {
+		// when --non_gcp  is set, instance metadata server(imds) is not defined so
+		// accessToken is unavailable from imds and --service_account_key must be
+		// set to generate accessToken.
+		if mf == nil && opts.ServiceAccountKey == "" {
```
qiwzhang: We should move this check out of the function.  This check should be done earlier before it is called.

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -61,57 +61,91 @@ func NewServiceConfigFetcher(mf *metadata.MetadataFetcher, opts options.ConfigGe
...
+		glog.Infof("start checking new rollouts every %v seconds", interval)
+		s.checkRolloutsTicker = time.NewTicker(interval)
 
-		for range scf.checkRolloutsTicker.C {
-			glog.Infof("check new rollouts for service %v", scf.serviceName)
+		for range s.checkRolloutsTicker.C {
+			glog.Infof("check new rollouts for service %v", s.serviceName)
 
-			serviceConfig, err := scf.FetchConfig("")
+			serviceConfig, err := s.FetchConfig("")
```
qiwzhang: I have not read FetchConfig() with empty yet.  But readability wise, I prefer the timer to call fetchLatestRolloutID() code
Or even better, the timer code is in the RolloutID fetcher. It will trigger ServiceConfigFetcher.

So the change to ServiceConfigFetcher should be smaller.

src/go/serviceconfig/service_rollout_id_fetcher.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	scpb "google.golang.org/genproto/googleapis/api/servicecontrol/v1"
+)
+
+type ServiceRolloutIdFetcher struct {
```
qiwzhang: let us call this as RolloutChangeDetector. periodical timer should be here.

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -61,57 +61,91 @@ func NewServiceConfigFetcher(mf *metadata.MetadataFetcher, opts options.ConfigGe
...
-				return nil, fmt.Errorf("fail to get access token: %v", err)
-			}
-			return scf.callServiceManagement(util.FetchConfigURL(scf.opts.ServiceManagementURL, scf.serviceName, newConfigId), token)
+
+		if newRolloutId == s.curRolloutId {
+			return "", nil, nil
+		}
+
+		rollout := new(smpb.Rollout)
+		fetchRolloutUrl := util.FetchRolloutURL(s.opts.ServiceManagementURL, s.serviceName, newRolloutId)
```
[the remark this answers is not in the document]
TAOXUY: Actually only one way. 
fetchNewRolloutId() is based on service control report.

src/go/serviceconfig/service_rollout_id_fetcher.go
```diff
@@ -0,0 +1,52 @@
...
+	serviceName       string
+	serviceControlUrl string
+	client            http.Client
+	accessToken       func() (string, time.Duration, error)
```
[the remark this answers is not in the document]
TAOXUY: Thanks for the reminder.

TAOXUY:
Done @qiwzhang @nareddyt 
The unit tests of service_config_fetcher and that of config_id_detector will be uploaded after we finalize the functional codes.

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -15,239 +15,77 @@
...
-				return nil, fmt.Errorf("fail to get access token: %v", err)
-			}
-			return scf.callServiceManagement(util.FetchConfigURL(scf.opts.ServiceManagementURL, scf.serviceName, configId), token)
-		}
-
-		glog.Infof("check new rollouts for service %v", scf.serviceName)
-		newRolloutId, newConfigId, err := scf.loadConfigFromRollouts(scf.serviceName, scf.curRolloutId, scf.curConfigId())
-		if err != nil {
-			glog.Errorf("error occurred when checking new rollouts, %v", err)
+		serviceConfig := new(confpb.Service)
```
qiwzhang: why do we need to make it as a function.  not easy to read

src/go/configmanager/config_manager.go
```diff
@@ -119,49 +122,94 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
...
-		return nil, fmt.Errorf(`failed to create https client to call ServiceManagement service, got error: %v`, err)
-	}
+			configId, err = mf.FetchConfigId()
+			if configId == "" || err != nil {
+				return nil, fmt.Errorf("failed to read metadata with key endpoints-service-version from metadata server")
+			}
+		}
+	} else if rolloutStrategy == util.ManagedRolloutStrategy {
+		m.RolloutIdDetector = sc.NewRolloutIdDetector(client, opts.ServiceControlURL, m.serviceName, accessToken)
+		latestRolloutId, err := m.RolloutIdDetector.FetchLatestRolloutId()
```
qiwzhang: At start,  we don't need to call RolloutIDDetector to get the latest rollout ID
we just call ServiceConfigFetcher to get the latest rollout info, that call doens't need any rollout_id.   Its returns will get the latest rollout id.  and the config_id to use, and remember this config_id as curr_config_id. 
Next get latest rollout may return the same config_id,  if so, skip it.

qiwzhang: I see.  you are fetching the rollout info by a specific rollout ID. That is why you need rollout ID first.

Hmm.  I like to keep the old way:  we always call for the latest rollout.  with this URL

"/v1/services/$serviceName/rollouts?filter=status=SUCCESS"

Empty report only to detect rollout change,  actually we don't need that rollout id.  

Once there is rollout change,
1) we always ask for the latest rollout,   from that to find out the config_id,  
2) if config_id is changed,  then fetch and apply it.

TAOXUY:
Done @qiwzhang

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -15,239 +15,74 @@
...
-	req.Header.Add("Authorization", "Bearer "+token)
-	req.Header.Set("Content-Type", "application/x-protobuf")
-	resp, err := scf.client.Do(req)
-	if err != nil {
-		return nil, err
-	}
-	if resp.StatusCode != http.StatusOK {
-		resp.Body.Close()
-		return nil, fmt.Errorf("http call to %s returns not 200 OK: %v", path, resp.Status)
+	if !(math.Abs(100.0-highestPercent) < 1e-9) {
```
nareddyt: Is this needed? We don't care about the absolute percentage, just the highest one.

TAOXUY: This is just for warning no other service configs except the highest one are not applied

qiwzhang:
Hi @TAOXUY ,  we may need to cache the token.  Since we are calling report every minute, I am not sure if we can call GCP metadata server for access token every minute.   It may NOT be designed to handle such load since it needs to support many instances.

TAOXUY:
> Hi @TAOXUY , we may need to cache the token. Since we are calling report every minute, I am not sure if we can call GCP metadata server for access token every minute. It may NOT be designed to handle such load since it needs to support many instances.

Hi Wayne, actually both  token_generator and metadata_fetcher already have cache mechanism:
https://github.com/GoogleCloudPlatform/esp-v2/blob/96553cd57a066a7ca1e59280c39c0603e7fe5db9/src/go/metadata/metadata_fetcher.go#L96
https://github.com/GoogleCloudPlatform/esp-v2/blob/96553cd57a066a7ca1e59280c39c0603e7fe5db9/src/go/util/token_generator.go#L45

qiwzhang:
Cool.  Thanks @TAOXUY .

We may need to mutex protect the token cache for token_generator.go 
Also the checking the time and return cached token can be moved to before reading the file.

TAOXUY:
> Cool. Thanks @TAOXUY .
> 
> We may need to mutex protect the token cache for token_generator.go
> Also the checking the time and return cached token can be moved to before reading the file.

Done

src/go/util/token_generator.go
```diff
@@ -38,13 +44,26 @@ func GenerateAccessTokenFromFile(serviceAccountKey string) (string, time.Duratio
...
+	if tokenCache.AccessToken == "" || now.After(tokenCache.Expiry.Add(-time.Second*60)) {
+		return "", 0
+
 	}
+
+	return tokenCache.AccessToken, tokenCache.Expiry.Sub(now)
+}
+
+func generateAccessToken(keyData []byte) (string, time.Duration, error) {
+	if token, duration := activeAccessToken(); token != "" {
```
qiwzhang: why do we check it again? The cache is checked in line 35.

src/go/configmanager/config_manager_test.go
```diff
@@ -1717,9 +1733,7 @@ func TestServiceConfigAutoUpdate(t *testing.T) {
 		if resp.Version != newConfigID {
```
qiwzhang: How do we verify the new config has been rollouted to configMgr?

src/go/configmanager/config_manager_test.go
```diff
@@ -1565,13 +1566,23 @@ func TestServiceConfigAutoUpdate(t *testing.T) {
...
 		desc: "Success for service config auto update",
+		fakeOldScReport: fmt.Sprintf(`{
+                "serviceConfigId": "%s",
+                "serviceRolloutId": "%s"
+            }`, oldRolloutID, oldConfigID),
+		fakeNewScReport: fmt.Sprintf(`{
+                "serviceConfigId": "%s",
+                "serviceRolloutId": "%s"
+            }`, newRolloutID, newConfigID),
 		fakeOldServiceRollout: fmt.Sprintf(`{
```
qiwzhang: fakeOldSrcReport and fakeNewSrReport are generated, do they have to be in the testCase struct?  Can they be just a variable?

src/go/serviceconfig/service_config_fetcher.go
qiwzhang: This is not easy to read,  maybe just 

highestPercent < 100.0

src/go/util/marshal_test.go
```diff
@@ -0,0 +1,69 @@
...
+		if !proto.Equal(getResp, wantResp) {
+			t.Errorf("fail in UnmarshalBytesToPbMessage, want: %v, ge: %v", wantResp, getResp)
+		}
+	}
+
+	wantResp = &confpb.Service{
+		Id: "test-id",
+	}
+	getResp = &confpb.Service{}
+	_test()
```
qiwzhang: can we separate them into different tests?

TAOXUY: What do you mean by separating them into different tests? 
I feel those test cases belong to same category.

src/go/configmanager/config_manager_test.go
TAOXUY: Putting in the struct is consistent with other serviceRollout/serviceConfig layouts, which make testcases clear.

src/go/util/marshal_test.go
qiwzhang: Into different Test*** so if one of them fails, the error message is clear.  For now, if one of them fail, the whole test fails.  have to read the log to find out which test fails.

src/go/configmanager/config_manager_test.go
qiwzhang: OK, I am fine either way.

src/go/util/marshal_test.go
TAOXUY: - Replace Fatal with Error so one failure won't effect others. 
- Print the type of marshal target, so it looks like following and it is straightforward.
`marshal_test.go:36: fail in UnmarshalBytesToPbMessage on *servicecontrol.ReportRequest...`

TAOXUY: > The logic for config_manager and RolloutDetector and ConfigFetcher are good.
> 
> Is the test code updated for the new logic?

Done.

qiwzhang:
Well, LGTM so far.  Once all tests are passed, I will approve it

src/go/util/token_generator_test.go
```diff
@@ -15,21 +15,32 @@
...
+	"io/ioutil"
+	"os"
 	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/esp-v2/src/go/util/testdata"
 )
 
 func TestGenerateAccessToken(t *testing.T) {
+	tmpFile, err := ioutil.TempFile(os.TempDir(), "TestGenerateAccessToken-")
```
nareddyt: I suggest unit tests do not use files. This will be difficult to import into google3 because they need to use `runfiles` format.

src/go/serviceconfig/service_config_fetcher_test.go
```diff
@@ -0,0 +1,171 @@
...
+			return
+		}
+
+		if !proto.Equal(getConfig, serviceConfig) {
+			t.Fatalf("want service config: %v, get service config: %v", serviceConfig, getConfig)
+		}
+	}
+
+	// Test success of fetching the service config.
+	_test(serviceConfigId, serviceConfig, "")
```
nareddyt: These (and other tests) don't seem to follow table-driven golang test style? Could you refactor? Code should be pretty much the same.

src/go/serviceconfig/rollout_id_change_fetcher_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"time"
+
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	"github.com/golang/protobuf/proto"
+
+	scpb "google.golang.org/genproto/googleapis/api/servicecontrol/v1"
+)
+
+func initServiceControlForTestNewRolloutId(t *testing.T, serviceRolloutId *string) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
nareddyt: Consider consolidating all mock servers here? https://github.com/GoogleCloudPlatform/esp-v2/blob/96553cd57a066a7ca1e59280c39c0603e7fe5db9/src/go/util/httptest_util.go#L26

src/go/util/token_generator_test.go
TAOXUY: The functional code does read from file, so unit test based on file is necessary.
Is it very difficult to do this in copybara? If so, since this piece of code is not used by gateway, so maybe just skip  in copybara?

nareddyt: Previously, this test was working because it did not test the `GenerateTokenFromFile` function. That was written so the file can be decoupled from the unit test. Can you re-write it for that? Or just test `generateAccessToken` and `activeAccessToken`?

Otherwise, you will need to hook in with the integration test framework 
 `tests/env/platform/files.go` for this to work in google3. There's no easy way to copybara it otherwise.

TAOXUY: put the file path in `tests/env/platform/files.go` and create/write/read/delete in the runtime. Do you think it works in google3?

nareddyt: Can you remove the creating/deleting of the file? It probably won't work in google3. Why not just write the file and check it into our repo and use it for the test? Just reading the file will work (see other tests that use `platform` to read files).

Also name the file `FakeServiceAccount` or something like that. The file doesn't have an access token in it.

TAOXUY: Done

PR #85, opened 2020-03-19
Add missing tests and validation for Path Matcher filter

nareddyt:
Based on comments for path matcher envoy security review and current C++ unit test coverage.

### DIFF
--- a/api/envoy/http/common/base.proto
+++ b/api/envoy/http/common/base.proto
@@ -37,12 +37,16 @@ message Pattern {
   //     /dictionary/{term:1}/{term}
   //     /search{?q*,lang}
   //
-  string uri_template = 1;
+  string uri_template = 1 [(validate.rules).string = {
+    well_known_regex: HTTP_HEADER_VALUE,
+    strict: false
+  }];
 
   // HTTP request method to match against as defined by
   // [rfc7231](https://tools.ietf.org/html/rfc7231#page-21). For
   // example: GET, HEAD, POST, PUT, DELETE.
-  string http_method = 2;
+  string http_method = 2
+      [(validate.rules).string.well_known_regex = HTTP_HEADER_NAME];
 }
 
 message HttpUri {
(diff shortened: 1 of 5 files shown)


api/envoy/http/common/base.proto
```diff
@@ -37,12 +37,16 @@ message Pattern {
   //     /dictionary/{term:1}/{term}
   //     /search{?q*,lang}
   //
-  string uri_template = 1;
+  string uri_template = 1 [(validate.rules).string = {
+    well_known_regex: HTTP_HEADER_VALUE,
```
qiwzhang: http_template may allow {}*, not sure if it is allowed in HTTP_HEADER_VALUE.

src/envoy/http/path_matcher/filter_config_test.cc
```diff
@@ -159,6 +158,49 @@ rules {
...
+      FilterConfig cfg(config_pb, EMPTY_STRING, mock_factory),
+      ProtoValidationException, "invalid pattern");
+}
+
+TEST(FilterConfigTest, NonStandardHttpMethod) {
+  const char kFilterConfigBasic[] = R"(
+rules {
+  operation: "1.cloudesf_testing_cloud_goog.Bar"
+  pattern {
+    http_method: "NonStandardMethod"
```
qiwzhang: we do allow custom []([method](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L372).)

api/envoy/http/common/base.proto
nareddyt: `HTTP_HEADER_VALUE` with `strict = false` essentially allows all characters except for `\r\n\0`. This was specifically made to be compliant with Envoy, and Envoy uses this validation upstream as well. So `{}*` characters should be allowed.

Confirmed by:
- All integration tests pass
- https://github.com/envoyproxy/protoc-gen-validate/pull/323

PR #87, opened 2020-03-20
Ignore invalid http header characters in backend routing filter fuzz test

nareddyt:
Path matcher ensures the `binding_query_params` has valid header values. Prevent the fuzzer from generating invalid characters.

Added reproducer test case.

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21229


src/envoy/http/backend_routing/filter_fuzz_test.cc
```diff
@@ -22,46 +22,43 @@ DEFINE_PROTO_FUZZER(
...
+  Utils::setStringFilterState(
+      *mock_decoder_callbacks.stream_info_.filter_state_, Utils::kQueryParams,
+      input.binding_query_params());
 
-    // Generate the user request.
-    auto headers =
-        Envoy::Fuzz::fromHeaders<Envoy::Http::TestRequestHeaderMapImpl>(
-            input.user_request().headers());
+  // Create the filter.
+  FilterConfigSharedPtr config = std::make_shared<FilterConfig>(
```
qiwzhang: Normally, FilterConfig can throw exception,  so this can be run inside try-catch block.

nareddyt: Done

nareddyt:
Flakey test and unrelated change, merging.

PR #90, opened 2020-03-24
Add X-Forwarded-Authorization header

qiwzhang:

### DIFF
--- a/tests/integration_test/backend_auth_using_imds_test.go
+++ b/tests/integration_test/backend_auth_using_imds_test.go
@@ -48,7 +48,6 @@ func TestBackendAuthWithImdsIdToken(t *testing.T) {
 		desc     string
 		method   string
 		path     string
-		headers  map[string]string
 		message  string
 		wantResp string
 	}{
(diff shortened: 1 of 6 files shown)


tests/integration_test/backend_auth_using_imds_test.go
```diff
@@ -61,7 +62,8 @@ func TestBackendAuthWithImdsIdToken(t *testing.T) {
 			desc:     "Add Bearer token for APPEND_PATH_TO_ADDRESS backend that requires JWT token",
 			method:   "GET",
 			path:     "/bearertoken/append?key=api-key",
-			wantResp: `{"Authorization": "Bearer ya29.append", "RequestURI": "/bearertoken/append?key=api-key"}`,
+			headers:  map[string]string{"Authorization": "Bearer origin-token"},
+			wantResp: `{"Authorization": "Bearer ya29.append", "RequestURI": "/bearertoken/append?key=api-key", "X-Forwarded-Authorization":"Bearer origin-token"}`,
```
nareddyt: I suggest you keep this test case as-is and add a new test case. Change the description to something like: "Downstream authorization header is copied when upstream requires JWT token".

Also consider adding a test to `backend_auth_using_iam_test.go` for completeness

https://github.com/GoogleCloudPlatform/esp-v2/blob/fc0d4fbb331e6c8e1bc452bc26a0d318c409d7fd/tests/integration_test/backend_auth_using_iam_test.go#L61

tests/integration_test/backend_auth_using_imds_test.go
```diff
@@ -48,6 +48,7 @@ func TestBackendAuthWithImdsIdToken(t *testing.T) {
 		desc     string
 		method   string
 		path     string
+		headers  map[string]string
 		message  string
 		wantResp string
```
TAOXUY: add integration like you did in ESP for the combinations of cases:
- if the request originally has "Authorization"
- if the "Authorization" token is needed to be overwritten.

qiwzhang: Done, add two cases in backend_auth_disable_auth_test.go

tests/integration_test/backend_auth_using_imds_test.go
qiwzhang: Done, add two cases in backend_auth_disable_auth_test.go

PR #94, opened 2020-03-25
b/152394779: put gcsrunner in official release images

TAOXUY:


scripts/all-utilities.sh
```diff
@@ -310,11 +310,15 @@ function get_serverless_image_name() {
...
+  echo -n "gcr.io/${ESP_RELEASE_PROJECT}/endpoints-runtime"
 }
 
 function get_serverless_image_release_name() {
-  echo -n 'gcr.io/endpoints-release/endpoints-runtime-serverless'
+  echo -n "gcr.io/${ESP_RELEASE_PROJECT}/endpoints-runtime-serverless"
+}
+
+function get_gcsrunner_image_release_name() {
+  echo -n "gcr.io/${ESP_RELEASE_PROJECT}/gcsrunner"
```
qiwzhang: can we keep it in our cloudese-testing project for gcsrunner?

endpoints-release is for users.  gcsrunner is not for users.

KyleWiese: Xuyang made a good comment offline that `cloudesf-testing` would also have all the CI builds in it which might get messy. If we don't want to put it in `endpoints-release`, would it be preferable for us to have a separate GCR repo for the gcsrunner releases?

TAOXUY: Yes, we can do that. The destination project totally depends on you guys.

KyleWiese: Okay, let me talk with the team to see what we want it to be named.

TAOXUY: Updated to be ` gcr.io/espv2-gcsrunner-release/gcsrunner`

PTAL

scripts/release/release-stable.sh
```diff
@@ -29,6 +29,8 @@ This script will release stable ESPv2 docker image with format of:
   $(get_proxy_image_release_name):\${MAJOR_BASE_VERSION}
   $(get_serverless_image_release_name):\${MINOR_BASE_VERSION}
   $(get_serverless_image_release_name):\${MAJOR_BASE_VERSION}
+  $(get_gcsrunner_image_release_name):\${MINOR_BASE_VERSION}
+  $(get_gcsrruner_image_release_name):\${MAJOR_BASE_VERSION}
```
KyleWiese: Is this where the patch version would go? Or does that part of the tagging happen separately?

TAOXUY: We haven't put the patch part. It is only the release part.

KyleWiese: Gotcha. LGTM

TAOXUY:
Change only involves release scripts so skip the CI.

PR #96, opened 2020-03-27
Add fuzz test for Path Matcher filter

nareddyt:
Other minor changes:
- Check for missing headers in filter to improve fuzzing
- Expand on README as mentioned in security review

### DIFF
--- a/src/envoy/http/path_matcher/README.md
+++ b/src/envoy/http/path_matcher/README.md
@@ -4,7 +4,7 @@
 
 This filter operates on downstream headers to determine the operation name and
 map variable bindings for an API request.
-It them updates the shared filter state, reducing redundant work in later filters.
+It then updates the shared filter state, reducing redundant work in later filters.
 
 State modifications:
 - Modifies shared filter state
--- a/src/envoy/http/path_matcher/README.md
+++ b/src/envoy/http/path_matcher/README.md
@@ -4,7 +4,7 @@
 
 This filter operates on downstream headers to determine the operation name and
 map variable bindings for an API request.
-It them updates the shared filter state, reducing redundant work later filters.
+It them updates the shared filter state, reducing redundant work in later filters.
 
 State modifications:
 - Modifies shared filter state
@@ -20,6 +20,14 @@ in the shared filter state. The results of this match are used the following fil
 - [Backend Auth](../backend_auth/README.md)
 - [Backend Routing](../backend_routing/README.md)
 - [Service Control](../service_control/README.md)
+- [JWT Authn](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/jwt_authn_filter)
+
+This selector is used by these filters (instead of the path) for a variety of operations:
+
+- Selecting a JWT provider for client authentication
+- Checking if an API key is required
+- Rewriting paths for remote backends
+- Attached a JWT for backend authentication
 
 ### Variable Bindings
 
@@ -31,6 +39,10 @@ and stores them in the shared filter state. The result of the transformation is
 used by the following filters:
 
 - [Backend Routing](../backend_routing/README.md)
+
+The variable bindings are used for:
+
+- Attaching query parameters to the request
 
 ## Configuration
 
--- a/src/envoy/http/path_matcher/filter.cc
+++ b/src/envoy/http/path_matcher/filter.cc
@@ -42,10 +42,10 @@ typedef Envoy::ConstSingleton<RcDetailsValues> RcDetails;
 Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
     Envoy::Http::RequestHeaderMap& headers, bool) {
   if (!headers.Method()) {
-    rejectRequest(Envoy::Http::Code(404), "No method in request headers.");
+    rejectRequest(Envoy::Http::Code(400), "No method in request headers.");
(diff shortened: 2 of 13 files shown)


src/envoy/http/path_matcher/README.md
```diff
@@ -4,7 +4,7 @@
 
 This filter operates on downstream headers to determine the operation name and
 map variable bindings for an API request.
-It them updates the shared filter state, reducing redundant work later filters.
+It them updates the shared filter state, reducing redundant work in later filters.
```
qiwzhang: typo?

src/envoy/http/path_matcher/filter.cc
```diff
@@ -41,6 +41,14 @@ typedef Envoy::ConstSingleton<RcDetailsValues> RcDetails;
 
 Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
     Envoy::Http::RequestHeaderMap& headers, bool) {
+  if (!headers.Method()) {
+    rejectRequest(Envoy::Http::Code(404), "No method in request headers.");
```
qiwzhang: using 400 for invalid parameter?

src/envoy/http/path_matcher/filter_fuzz_test.cc
```diff
@@ -0,0 +1,73 @@
...
+      mock_decoder_callbacks;
+  NiceMock<Envoy::Server::Configuration::MockFactoryContext>
+      mock_factory_context;
+
+  try {
+    // Create the filter.
+    FilterConfigSharedPtr config = std::make_shared<FilterConfig>(
+        input.config(), "fuzz-test-stats", mock_factory_context);
+    Filter filter(config);
+    filter.setDecoderFilterCallbacks(mock_decoder_callbacks);
```
qiwzhang: can we combine two try-catch blocks into one and not need to use ASSERT_NO_THROW?
the code should be much shorter

src/envoy/http/path_matcher/README.md
nareddyt: Done

src/envoy/http/path_matcher/filter.cc
nareddyt: Done

src/envoy/http/path_matcher/filter_fuzz_test.cc
nareddyt: Could you clarify how they can be combined? They check for different exceptions.

Also, we should ensure that `doTest` will not throw any exception. If we move it outside the try/catch block, then we need to make `Filter filter` a pointer in the test.

qiwzhang: If we don't care about exception type, we could do this:

filter_share_ptr filter
try {
     validdate proto
     create config and filter 
} catch (...)

doTest(filter, input)

qiwzhang: doTest() may not need to be a separate function

src/envoy/http/path_matcher/filter_test.cc
```diff
@@ -170,6 +170,23 @@ TEST_F(PathMatcherFilterTest, DecodeHeadersNoMatch) {
...
 
+TEST_F(PathMatcherFilterTest, DecodeHeadersMissingHeaders) {
+  Envoy::Http::TestRequestHeaderMapImpl missingPath{{":method", "POST"}};
+  Envoy::Http::TestRequestHeaderMapImpl missingMethod{{":path", "/bar"}};
+
+  // Filter should reject this request
+  EXPECT_CALL(
+      mock_cb_.stream_info_,
+      setResponseFlag(
+          Envoy::StreamInfo::ResponseFlag::UnauthorizedExternalService))
```
TAOXUY: Why it is UnauthorizedExternalService? I saw it is rejected by 400 so where it is set?

src/envoy/http/path_matcher/filter_fuzz_test.cc
nareddyt: But then even the mocks (`MockFactoryContext`) need to be moved up. Unless I make all of these pointers as well, this will result in a performance hit. The mocks will be constructed even when the proto is invalid.

What do you think is cleaner? I prefer the current solution, but I can change everything to pointers if needed.

src/envoy/http/path_matcher/filter_test.cc
nareddyt: In `rejectReject` sets this, it's independent of the HTTP status code. None of the other enum values seem like a better match.

src/envoy/http/path_matcher/filter_fuzz_test.cc
TAOXUY: Though you don't need pointer to combine these two tryCatch if adding multiple catch block, I also prefer current struct, which looks clear.

src/envoy/http/path_matcher/filter_test.cc
TAOXUY: Oh, I see.

src/envoy/http/path_matcher/filter_fuzz_test.cc
qiwzhang: OK, it could be either way.   Current way is fine

PR #97, opened 2020-03-27
b/149050012: Run e2e tests with HTTP/2

nareddyt:
For HTTP dynamic routing, set the protocol to be `h2` in `x-google-backend`.

Will verify this works by checking trace annotations.


nareddyt:
FYI I confirmed this works as expected

image

PR #98, opened 2020-03-28
b/149484856: e2e test x-google-jwt-locations

TAOXUY:


tests/e2e/client/apiproxy_bookstore_test.py
```diff
@@ -152,6 +152,18 @@ def verify_list_books(self, shelf, books):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.json().get('books', []), books)
 
+    def verify_swagger_importer(self):
```
nareddyt: I don't understand the naming here. What is "swagger importer"?

tests/endpoints/bookstore/bookstore_swagger_template.json
```diff
@@ -162,7 +162,11 @@
             }
           }
         },
-        "security": []
+        "security": [
+          {
+            "auth0_jwk": []
```
nareddyt: Can you add security to a different endpoint? I believe `/shelves` is used by other tests as a basic test endpoint that requires no downstream security.

tests/e2e/client/apiproxy_bookstore_test.py
TAOXUY: swagger_importer is the actual name of api compiler.

tests/endpoints/bookstore/bookstore_swagger_template.json
TAOXUY: I already add auth token to others requests so it should be fine.

I try to find one not depend on backend status(like bookId/shelfId ). Seems no other apis except this one and `get books`.

TAOXUY: Move to `GET /echo`

TAOXUY: Seems GET /echo is also used by load test. Change back to GET /shelves

tests/e2e/client/apiproxy_bookstore_test.py
```diff
@@ -152,6 +152,20 @@ def verify_list_books(self, shelf, books):
...
         self.assertEqual(response.json().get('books', []), books)
 
+    def verify_jwt_locations(self):
+        response = self._call_http(
+            '/shelves')
+        self.assertEqual(response.status_code, 401)
+
+        response = self._call_http(
+            '/shelves', userHeaders={"Jwt-Header-Name": "Jwt-Value-Prefix {}".format(FLAGS.auth_token)})
+        self.assertEqual(response.status_code, 200)
```
qiwzhang: Could you add a  good token to x-google-iap-assert (one of our deffault locations) make sure it get 401, and 
add one with x-google-iap-assert to a wrong token and a good token to Authroization, to get 200

TAOXUY: Sorry, that PR is submitted and it is updated here https://github.com/GoogleCloudPlatform/esp-v2/pull/101

PR #99, opened 2020-03-30
b/152686575: update to latest filter names

TAOXUY:
[Here](https://www.envoyproxy.io/docs/envoy/latest/intro/deprecated) is the latest filter name table.

### DIFF
--- a/examples/auth/envoy_config.json
+++ b/examples/auth/envoy_config.json
@@ -18,7 +18,7 @@
           {
             "filters": [
               {
-                "name": "envoy.http_connection_manager",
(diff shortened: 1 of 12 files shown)


qiwzhang:
Envoy must support alias, otherwise how come our old name works

If envoy supports alias, we should be able to make such changes without breaking api-gateway.

qiwzhang:
Just looked at Envoy,  it plans to deprecated these old names in 1.4.  We need to be careful when we update to Envoy 1.4.

So Envoy 1.4 will be a breaking change for them.  Unless their gcsrunner can do a string replacement

PR #100, opened 2020-03-30
add integration test for gRPCs mTLS

JLXIA:
Notes:
    I feel it is better that ESPv2 don't support grpc_backend_ssl_private_key_file and grpc_backend_ssl_cert_chain_file for backward compatiblity.  Two reasons:
   1,   user should just use --ssl_client_cert_path
   2,  this two are file names, but --ssl_client_cert_path is path, requires extra code changes

### DIFF
--- a/tests/integration_test/dynamic_routing_integration_test.go
+++ b/tests/integration_test/dynamic_routing_integration_test.go
@@ -773,7 +773,7 @@ func TestDynamicGrpcBackendTLS(t *testing.T) {
 			wantResp:       `{"id":"100","theme":"Kids"}`,
 		},
 		{
-			desc:           "gRPC client calling gRPCs remote backend through mTLS failed with incorrect client root cert",
+			desc:           "HTTP2 client calling gRPCs remote backend through mTLS failed with incorrect client root cert",
 			clientProtocol: "http2",
 			methodOrUrl:    "/v1/shelves/200?key=api-key",
 			mtlsCertFile:   platform.GetFilePath(platform.ProxyCert),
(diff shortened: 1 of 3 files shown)


tests/integration_test/dynamic_routing_integration_test.go
```diff
@@ -766,13 +764,35 @@ func TestDynamicGrpcBackendTLS(t *testing.T) {
...
+		{
+			desc:           "gRPC client calling gRPCs remote backend with mTLS succeed",
+			clientProtocol: "grpc",
+			methodOrUrl:    "GetShelf",
+			mtlsCertFile:   platform.GetFilePath(platform.ServerCert),
+			header:         http.Header{"x-api-key": []string{"api-key"}},
+			wantResp:       `{"id":"100","theme":"Kids"}`,
+		},
+		{
+			desc:           "gRPC client calling gRPCs remote backend through mTLS failed with incorrect client root cert",
```
nareddyt: Nit: Http client

JLXIA: Done

JLXIA:
> I am fine with not supporting these two flags, I bet not many users are using them. For these using them, they can always migrate to use v2 flag
> 
> But we are using fixed file names for the key file and cert file. what file names do we use?
> 
> We may need a doc to describe all these ssl flags for v2. Otherwise each time we have to read the code to understand how it works.

Currently, it uses client.key and client.crt under ssl_client_cert_path.  See description of --ssl_client_cert_path.

Yes, we need a detailed documentation for migration.

PR #105, opened 2020-04-02
support generate_self_signed_certificate

JLXIA:
1,   Put the cert/key under /tmp/, since /etc/ usually requires permission
2,  Don't install openssl by default, which may increase image size

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -190,6 +190,14 @@ environment variable or by passing "-k" flag to this script.
         with value "max-age=31536000; includeSubdomains;" is added for all responses from local backend.
         Not valid for remote backends.''')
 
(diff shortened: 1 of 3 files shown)


docker/generic/start_proxy.py
```diff
@@ -190,6 +190,14 @@ def make_argparser():
         with value "max-age=31536000; includeSubdomains;" is added for all responses from local backend.
         Not valid for remote backends.''')
 
+    parser.add_argument('--generate_self_signed_cert', action='store_true',
+        help='''Generate a self-signed certificate and key at start, then
+        store them in /etc/ssl/endpoints/server.crt and /etc/ssl/endponts/server.key.
```
qiwzhang, line 195: /tmp

JLXIA: Done. Thanks.

PR #106, opened 2020-04-02
b/153006364: Fix support for `x-http-method-override`

nareddyt:
If the override header is present, treat the remainder of the request as the overridden method.

Added integration tests for the backend request and service control report. This ensures overridden requests to the backend are spec compliant.


src/envoy/utils/http_header_utils.cc
```diff
@@ -38,13 +38,17 @@ absl::string_view extractHeader(const Envoy::Http::HeaderMap& headers,
...
 }
 
-absl::string_view getRequestHTTPMethodWithOverride(
-    absl::string_view originalMethod, const Envoy::Http::HeaderMap& headers) {
+bool handleHttpMethodOverride(Envoy::Http::HeaderMap& headers) {
   const auto* entry = headers.get(kHttpMethodOverrideHeader);
   if (entry) {
-    return entry->value().getStringView();
+    // Move the header.
+    headers.remove(Envoy::Http::Headers::get().Method);
```
qiwzhang, line 45: You can use build-in method:  headers.removeMethod()

src/envoy/utils/http_header_utils.cc
```diff
@@ -38,13 +38,17 @@ absl::string_view extractHeader(const Envoy::Http::HeaderMap& headers,
...
+    headers.addCopy(Envoy::Http::Headers::get().Method,
+                    entry->value().getStringView());
```
qiwzhang, line 47: you may be using a freed memory after removing Method.

src/envoy/utils/http_header_utils.cc
```diff
@@ -38,13 +38,17 @@ absl::string_view extractHeader(const Envoy::Http::HeaderMap& headers,
   return readHeaderEntry(entry);
 }
 
-absl::string_view getRequestHTTPMethodWithOverride(
-    absl::string_view originalMethod, const Envoy::Http::HeaderMap& headers) {
+bool handleHttpMethodOverride(Envoy::Http::HeaderMap& headers) {
   const auto* entry = headers.get(kHttpMethodOverrideHeader);
   if (entry) {
-    return entry->value().getStringView();
+    // Move the header.
```
qiwzhang, line 44: add a debug log entry here

tests/integration_test/http_method_override_test.go
```diff
@@ -0,0 +1,260 @@
...
+		},
+		{
+			desc:   "Overridden GET has no body for the backend to handle",
+			path:   "/echo?key=api-key",
+			method: "GET",
+			headers: map[string]string{
+				"X-HTTP-Method-Override": "POST",
+			},
+			body:          "This body will not be sent in the request",
+			httpCallError: fmt.Errorf(`{"code":500,"message":"Could not get body: EOF"}`),
```
qiwzhang, line 150: what will happen for POST overrided by GET?  Will the body be send to the backend,  will the backend echo it body?

tests/integration_test/http_method_override_test.go
```diff
@@ -0,0 +1,260 @@
...
+	}{
+		{
+			desc:    "Overridden POST is displayed as GET in SC Report",
+			url:     fmt.Sprintf("http://localhost:%v%v", s.Ports().ListenerPort, "/echo/nokey/OverrideAsGet"),
+			message: "hello hello",
+			method:  "POST",
+			requestHeader: map[string]string{
+				"X-HTTP-Method-Override": "GET",
+			},
+			httpCallError: fmt.Errorf("405 Method Not Allowed"),
```
qiwzhang, line 217: where is this 405 replied,  by echo backend server? 
can we have a case with a success one?

tests/integration_test/http_method_override_test.go
```diff
@@ -0,0 +1,260 @@
...
+	for _, tc := range testData {
+		url := fmt.Sprintf("http://localhost:%v%v", s.Ports().ListenerPort, tc.path)
+		gotResp, err := client.DoWithHeaders(url, tc.method, "test-body", tc.headers)
+
+		if tc.httpCallError == nil {
+			if err != nil {
+				t.Fatal(err)
+			}
+		} else {
+			if !strings.Contains(err.Error(), tc.httpCallError.Error()) {
```
TAOXUY, line 102: err can be nil, add err == nil checker

tests/integration_test/http_method_override_test.go
```diff
@@ -0,0 +1,260 @@
...
+	for _, tc := range testData {
+		url := fmt.Sprintf("http://localhost:%v%v", s.Ports().ListenerPort, tc.path)
+		gotResp, err := client.DoWithHeaders(url, tc.method, tc.body, tc.headers)
+
+		if tc.httpCallError == nil {
+			if err != nil {
+				t.Fatal(err)
+			}
+		} else {
+			if !strings.Contains(err.Error(), tc.httpCallError.Error()) {
```
TAOXUY, line 162: ditto

tests/integration_test/http_method_override_test.go
```diff
@@ -0,0 +1,260 @@
...
+		resp, err := client.DoWithHeaders(tc.url, tc.method, tc.message, tc.requestHeader)
+		if tc.httpCallError == nil {
+			if err != nil {
+				t.Fatalf("Test (%s): failed, %v", tc.desc, err)
+			}
+			if !strings.Contains(string(resp), tc.wantResp) {
+				t.Errorf("Test (%s): failed,  expected: %s, got: %s", tc.desc, tc.wantResp, string(resp))
+			}
+		} else {
+			if !strings.Contains(err.Error(), tc.httpCallError.Error()) {
```
TAOXUY, line 249: ditto

src/envoy/utils/http_header_utils.cc
```diff
@@ -38,13 +38,17 @@ absl::string_view extractHeader(const Envoy::Http::HeaderMap& headers,
   return readHeaderEntry(entry);
 }
 
-absl::string_view getRequestHTTPMethodWithOverride(
-    absl::string_view originalMethod, const Envoy::Http::HeaderMap& headers) {
+bool handleHttpMethodOverride(Envoy::Http::HeaderMap& headers) {
```
TAOXUY, line 41: Add unit test for this util method?

src/envoy/utils/http_header_utils.cc
nareddyt: Ah thanks. The catch is you need to use `RequestHeaderMap` instead of the generic `HeaderMap` to make use of these `O(1)` operations. I can do that.

src/envoy/utils/http_header_utils.cc
nareddyt: I think `HeaderMap` was recently re-written to not have this problem. ASan should catch this as well, as we have c++ unit tests that cover this.

src/envoy/utils/http_header_utils.cc
nareddyt: Done. One complication is that this function is outside a filter. We use `ENVOY_LOG_MISC` instead.

src/envoy/utils/http_header_utils.cc
nareddyt: Done

tests/integration_test/http_method_override_test.go
nareddyt: Done

tests/integration_test/http_method_override_test.go
nareddyt: Done

tests/integration_test/http_method_override_test.go
nareddyt: Done

tests/integration_test/http_method_override_test.go
nareddyt: Good idea, it's confusing. Replaced with a success case only.

tests/integration_test/http_method_override_test.go
nareddyt: Yes, the backend will receive the body. This is spec complaint. Added a test.

src/envoy/http/path_matcher/filter.cc
```diff
@@ -49,8 +49,13 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
     return Envoy::Http::FilterHeadersStatus::StopIteration;
   }
 
-  std::string method(utils::getRequestHTTPMethodWithOverride(
-      headers.Method()->value().getStringView(), headers));
+  bool usedMethodOverride = utils::handleHttpMethodOverride(headers);
+  if (usedMethodOverride) {
```
JLXIA, line 53: Just. if (utils::handleHttpMethodOverride(headers)) and add a comment

src/envoy/utils/http_header_utils.cc
```diff
@@ -38,13 +41,22 @@ absl::string_view extractHeader(const Envoy::Http::HeaderMap& headers,
   return readHeaderEntry(entry);
 }
 
-absl::string_view getRequestHTTPMethodWithOverride(
-    absl::string_view originalMethod, const Envoy::Http::HeaderMap& headers) {
+bool handleHttpMethodOverride(Envoy::Http::RequestHeaderMap& headers) {
   const auto* entry = headers.get(kHttpMethodOverrideHeader);
   if (entry) {
```
JLXIA, line 46: if (entry == nullptr) {
   return false
}

...

src/envoy/http/path_matcher/filter.cc
nareddyt: Done

src/envoy/utils/http_header_utils.cc
nareddyt: Done

PR #109, opened 2020-04-03
b/153198370: Add service control check and jwt authn for websocket

TAOXUY:


tests/integration_test/websocket_test.go
```diff
@@ -36,40 +38,86 @@ func TestWebsocket(t *testing.T) {
...
+		schema               string
+		wantResp             string
+		wantScRequests       []interface{}
+		wantSkipScRequestNum int
 	}{
 		{
-			desc:         "Websocket call succeed",
+			desc:         "Websocket call succeed with service control check and jwt authn",
 			path:         "/websocketecho",
+			query:        "key=api-key&access_token=" + testdata.FakeCloudTokenMultiAudiences,
```
qiwzhang, line 53: Can we use header for the JWT token?

PR #110, opened 2020-04-04
Apply retry and network_fail_open for failed server response properly

qiwzhang:
Specifically:
4xx response:  not retry and not to apply network_fail_open policy
5xx response:  apply retry and network_fail_open policy

for b/152387957:

### DIFF
--- a/src/envoy/http/service_control/client_cache.cc
+++ b/src/envoy/http/service_control/client_cache.cc
@@ -315,6 +315,9 @@ CancelFunc ClientCache::callCheck(
           // All 5xx server error codes have been mapped to Code::UNAVAILABLE.
           // network_fail_open only applies to 5xx server error codes.
           if (network_fail_open_ && status.error_code() == Code::UNAVAILABLE) {
+            ENVOY_LOG(debug,
+                      "service control check fails, but the request is allowed "
+                      "due to network_fail_open policy.");
             on_done(Status::OK, response_info);
           } else {
             on_done(status, response_info);
--- a/src/envoy/http/service_control/client_cache.cc
+++ b/src/envoy/http/service_control/client_cache.cc
@@ -300,24 +300,29 @@ CancelFunc ClientCache::callCheck(
                   "Service Control cache query: Check");
 
   auto* response = new CheckResponse;
-  client_->Check(request, response,
-                 [this, response, on_done](const Status& status) {
-                   CheckResponseInfo response_info;
-                   if (status.ok()) {
-                     Status converted_status = ::espv2::api_proxy::
-                         service_control::RequestBuilder::ConvertCheckResponse(
-                             *response, config_.service_name(), &response_info);
-                     on_done(converted_status, response_info);
-                   } else {
-                     if (network_fail_open_) {
-                       on_done(Status::OK, response_info);
-                     } else {
-                       on_done(status, response_info);
-                     }
-                   }
-                   delete response;
-                 },
-                 check_transport);
+  client_->Check(
+      request, response,
+      [this, response, on_done](const Status& status) {
+        CheckResponseInfo response_info;
+        if (status.ok()) {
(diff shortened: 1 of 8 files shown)


src/envoy/http/service_control/client_cache.cc
```diff
@@ -300,24 +300,29 @@ CancelFunc ClientCache::callCheck(
...
+          Status converted_status = ::espv2::api_proxy::service_control::
+              RequestBuilder::ConvertCheckResponse(
+                  *response, config_.service_name(), &response_info);
+          on_done(converted_status, response_info);
+        } else {
+          // Envoy::Grpc::httpToGrpcStatus() is called at http_call.cc at
+          // HttpCallImpl::onSuccess to map http_code to grpc_code.
+          // All 5xx server error codes have been mapped to Code::UNAVAILABLE.
+          // network_fail_open only applies to 5xx server error codes.
+          if (network_fail_open_ && status.error_code() == Code::UNAVAILABLE) {
```
TAOXUY, line 317: Check [5XX errors](https://github.com/cloudendpoints/esp/blob/ca70265e9ef6a560eed3bf0ff1cf7cc24c004a16/src/api_manager/service_control/aggregated.cc#L101)

nareddyt: This is already done in `http_call.cc`

> All 5xx server error codes have been mapped to Code::UNAVAILABLE.

I do think checking the error code twice is a little messy, but it's required: `HttpCall` checks for retry logic, `ClientCache` checks for network fail open logic. So it seems fine to me.

nareddyt, line 317: Please add a info or warning log about network fail open to make it obvious when debugging.

src/envoy/http/service_control/http_call.cc
```diff
@@ -101,8 +102,13 @@ class HttpCallImpl : public HttpCall,
 
         ENVOY_LOG(debug, "http call response status code: {}, body: {}",
                   status_code, body);
-        on_done_(Status(Code::INTERNAL, "Failed to call service control"),
-                 body);
+        std::string error_msg =
+            absl::StrCat("Calling ServiceControl failed with: ", status_code);
```
nareddyt, line 106: Since this error message will be exposed to API consumers, let's try to reduce confusion and say "Google Service Control API" instead of "ServiceControl".

src/envoy/http/service_control/http_call.cc
```diff
@@ -101,8 +102,13 @@ class HttpCallImpl : public HttpCall,
 
         ENVOY_LOG(debug, "http call response status code: {}, body: {}",
```
nareddyt, line 103: Slightly unrelated to this PR, but can you move this log line to be before `attemptRetry`? Otherwise, the logs will be out of order when the stack unwinds on retries.

src/envoy/http/service_control/http_call.cc
qiwzhang: Done

src/envoy/http/service_control/http_call.cc
qiwzhang: Done

src/envoy/http/service_control/http_call.cc
```diff
@@ -95,14 +96,20 @@ class HttpCallImpl : public HttpCall,
...
         }
 
-        ENVOY_LOG(debug, "http call response status code: {}, body: {}",
-                  status_code, body);
-        on_done_(Status(Code::INTERNAL, "Failed to call service control"),
-                 body);
+        std::string error_msg = absl::StrCat(
+            "Calling Google Service Control API failed with: ", status_code);
+        if (!body.empty()) {
+          error_msg += absl::StrCat(" and body: " + body);
```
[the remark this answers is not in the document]
qiwzhang: Sorry, will use another PR to fix it

PR #111, opened 2020-04-06
Handle duplicated `snake_names` in filter config

nareddyt:
This should never occur in practice, but guards against users who manually edit the service config or proto JSON mapping.

### DIFF
--- a/src/envoy/http/path_matcher/filter_config.cc
+++ b/src/envoy/http/path_matcher/filter_config.cc
@@ -14,6 +14,7 @@
 
 #include "src/envoy/http/path_matcher/filter_config.h"
 #include "common/common/empty_string.h"
+#include "common/protobuf/utility.h"
 
 namespace espv2 {
 namespace envoy {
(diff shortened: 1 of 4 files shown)


nareddyt:
@qiwzhang I changed the logic a little, PTAL again when free

src/envoy/http/path_matcher/filter_config.cc
```diff
@@ -41,8 +42,26 @@ FilterConfig::FilterConfig(
   path_matcher_ = pmb.Build();
 
   for (const auto& segment_name : proto_config_.segment_names()) {
-    snake_to_json_map_.emplace(segment_name.snake_name(),
-                               segment_name.json_name());
+    const auto& result = snake_to_json_map_.emplace(segment_name.snake_name(),
```
qiwzhang, line 45: what will happen if we don't check duplication?

I don't like to add complicated code to solve no-problem

TAOXUY: I agree the duplication may rarely happen in our repo but my only concern is if  we are going to opensource it even contribute it to envoy, shall we consider to have this after the filter gets decoupled?

qiwzhang: Well,  the checking can be done in configmgr. It will have more direct feedback to the users.   checking here is not easy  for users to find out such error.

qiwzhang: even we are going to upstream this filter,  it will NOT be in this shape,  not in this exact code.

nareddyt: Config manager only treats this as a list, so we would need to build a map to check it. Since we are building a map here anyways, I figured it would make sense to do it here.

Also, I think after Jilin changed `start_proxy` to run config manager and envoy as subprocesses, both errors provide direct feedback to users.

qiwzhang: Yes for static config used by api-gateway. 

Not for ADS used by configmgr.   When envoy receives an invalid config,  it will not crash envoy,  it will just wait for next update from configmgr.

nareddyt: Ok, let's do this in the golang code then. I'll close this PR and open another one.

nareddyt:
As discussed above, we will implement this in the go code.

PR #112, opened 2020-04-06
use absl::StrAppend for error message in http_call

qiwzhang:
absl::StrAppend is much faster than string +=


nareddyt:
Test failure probably means the enabling of the service hasn't propagated across GCP yet. Please consider increasing this sleep time: https://github.com/GoogleCloudPlatform/esp-v2/blob/a74de59c63a76886dabebc9394b2ea2a3475aea3/tests/e2e/scripts/cloud-run/deploy.sh#L313

PR #116, opened 2020-04-07
fix release script and remove useless file

TAOXUY:


scripts/release/release-stable.sh
```diff
@@ -50,8 +50,7 @@ function tag_stable_image() {
   local image=$1
 
   gcloud container images add-tag "${image}:${VERSION}" \
-    "${image}:${MINOR_BASE_VERSION}" "${image}:${MAJOR_BASE_VERSION}" \
-    --project ${APIPROXY_RELEASE_PROJECT} --quiet
```
nareddyt, line 54: FYI this is also used in `release-show.sh`

TAOXUY:
> PR title doesn't match, are we removing any files?

The file "-" under root path. It is hard to notice so it was introduced.

PR #118, opened 2020-04-07
Handle duplicated `snake_names` in path matcher filter config

nareddyt:
This should never occur in practice, but guards against users who manually edit the service config or proto JSON mapping.

IMO, the implementation in #111 is cleaner. But this will surface the error message to users quicker.

Also increase wait time on flaky test.

### DIFF
--- a/src/go/configinfo/service_info.go
+++ b/src/go/configinfo/service_info.go
@@ -123,7 +123,7 @@ func NewServiceInfoFromServiceConfig(serviceConfig *confpb.Service, id string, o
 	}
 
 	serviceInfo.processAccessToken()
-	if err := serviceInfo.ProcessTypes(); err != nil {
+	if err := serviceInfo.processTypes(); err != nil {
 		return nil, err
 	}
 	serviceInfo.addGrpcHttpRules()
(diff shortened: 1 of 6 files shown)


src/go/configinfo/service_info_test.go
```diff
@@ -1926,3 +1928,108 @@ func TestProcessApisForGrpc(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			wantSegments: []*pmpb.SegmentName{
+				{
+					SnakeName: "foo_bar",
+					JsonName:  "fooBar",
+				},
```
qiwzhang, line 1990: can we only generate one pair?

src/go/configinfo/service_info.go
```diff
@@ -607,18 +609,31 @@ func (s *ServiceInfo) extractApiKeyLocations(method *methodInfo, parameters []*c
 	method.ApiKeyLocations = append(method.ApiKeyLocations, headerNames...)
 }
 
-func (s *ServiceInfo) processTypes() {
-	// Create snake name to JSON name mapping.
+func (s *ServiceInfo) ProcessTypes() error {
+	// Create snake name to JSON name mapping (and validate against duplicates).
```
JLXIA, line 613: why capitalize? It should called inside service_info.go only

nareddyt: Right, thanks. Got confused by test structure

src/go/configinfo/service_info_test.go
nareddyt: Sure, good idea

PR #119, opened 2020-04-08
Support whether allowing underscore in HTTP headers

JLXIA:
1,   Update go-control-plane in order to include the update HttpProtoOption
2,  Add flag underscores_in_headers
3,  Add integration tests


src/go/configmanager/config_manager_test.go
```diff
@@ -43,6 +42,9 @@ import (
 	routerpb "github.com/envoyproxy/go-control-plane/envoy/config/filter/http/router/v2"
 	transcoderpb "github.com/envoyproxy/go-control-plane/envoy/config/filter/http/transcoder/v2"
 	hcmpb "github.com/envoyproxy/go-control-plane/envoy/config/filter/network/http_connection_manager/v2"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
```
nareddyt, line 45: Nit: Place this import in the proper section and `make format`

JLXIA: Done

PR #122, opened 2020-04-13
Covert non-5xx sidestream errors to 500 Internal Server Error

nareddyt:
This makes it clear that deployment errors (ie: SC Check fail due to permissions) are not caused by the client request.

### DIFF
--- a/src/envoy/http/service_control/client_cache.cc
+++ b/src/envoy/http/service_control/client_cache.cc
@@ -309,24 +309,27 @@ CancelFunc ClientCache::callCheck(
               RequestBuilder::ConvertCheckResponse(
                   *response, config_.service_name(), &response_info);
           on_done(converted_status, response_info);
-        } else {
+        } else if (status.error_code() == Code::UNAVAILABLE) {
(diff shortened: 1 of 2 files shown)


src/envoy/http/service_control/client_cache.cc
```diff
@@ -320,7 +320,12 @@ CancelFunc ClientCache::callCheck(
                       "due to network_fail_open policy.");
             on_done(Status::OK, response_info);
           } else {
-            on_done(status, response_info);
+            // This is not a client request error, so translate non-5xx error
+            // codes to 500 Internal Server Error. Error message contains
+            // details on the original error (including the original HTTP status
+            // code).
+            Status converted(Code::INTERNAL, status.error_message());
```
qiwzhang, line 327: I am doing similar changes in v1.  I discover that
 for 5xx error, it is better to preserve it.  only convert 4xx errors to INTERNAL
5xx errors will go here if it is network_fail_close.

nareddyt: Good point, fixed. Seems like the tests do catch this case.

PR #123, opened 2020-04-15
Only pass Interupt and SIGTERM signals to GCS Runner.

toddbeckman-g:
GCS Runner shuts down too much. Cloud Run sends periodic signals to the container, which GCS Runner always interprets as a terminator. This causes Envoy to shut down and the container to restart very often. Instead, only shut down when the container receives an actual termination signal.

TESTED=Manual testing. Feature not covered by automated testing.


nareddyt:
That test is flakey, we'll just merge it in

PR #125, opened 2020-04-17
Update README, adding more supported features

JLXIA:


README.md
```diff
@@ -72,6 +72,14 @@ ESPv2 also writes access logs for each request, providing insight into HTTP head
 * [Application Observability](https://cloud.google.com/endpoints/docs/grpc/tracing):
 Understand and debug problems with your API or ESPv2 by viewing traces and application logs.
 
+* Multi Transport Protocols: ESPv2 supports HTTP/1.1, HTTP/2, websockets, gRPC and gRPC-web.
+
+* Security: ESPv2 supports TLS for downstream and upstream, also mTLS for upstream connections, for both HTTP and gRPC.
+
+* Dynamic Load Balancing: ESPv2 can load balance traffic across multiple remote upstream services.
```
qiwzhang, line 79: Hmm, I am not sure about this one.  

We support this:
/foo  -> backend: foo
/bar -> backend: bar

But this is not load balancing. 

Load balancing is:
Two backends: foo and bar.   a request can route to either of them.  In envoy config, foo and bar are in the same "cluster".    

I don't think we support such load balancing

nareddyt: I was thinking about this, technically we support load balancing via dns for remote backends.

Agreed this might be confusing

JLXIA: Updated.

PR #127, opened 2020-04-20
[Draft] Add new Error Translator filter

nareddyt:
Notes:
- Implementation as discussed in the design doc.
- This PR only adds the code and config for the filter. None of this code will be built into the main envoy binary or enabled by Config Manager yet.

Still a work in progress, not ready for review.

### DIFF
--- a/src/envoy/http/error_translator/filter.cc
+++ b/src/envoy/http/error_translator/filter.cc
@@ -1,0 +1,89 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "src/envoy/http/error_translator/filter.h"
+
+#include <string>
+
+#include "common/http/headers.h"
+#include "external/envoy/include/envoy/stream_info/_virtual_includes/stream_info_interface/envoy/stream_info/stream_info.h"
+#include "src/envoy/utils/filter_state_utils.h"
+
+namespace espv2 {
+namespace envoy {
+namespace http_filters {
+namespace error_translator {
+
+using ::Envoy::Http::FilterHeadersStatus;
+
+Filter::Filter(FilterConfigSharedPtr config) : config_(config) {}
+
+// TODO(nareddyt): encodeHeaders or encodeData?
+// Probably need both for `handleEnvoyFilterError`
+FilterHeadersStatus Filter::encodeHeaders(Envoy::Http::ResponseHeaderMap&,
+                                          bool) {
+  if (utils::hasErrorFilterState(
+          *encoder_callbacks_->streamInfo().filterState())) {
+    // Enhanced handling of espv2 filter errors.
+    handleEspv2FilterError();
+    return FilterHeadersStatus::Continue;
+  }
+
+  if (encoder_callbacks_->streamInfo().responseCodeDetails() !=
+      Envoy::StreamInfo::ResponseCodeDetails::get().ViaUpstream) {
+    // Basic handling of non-espv2 upstream envoy filter errors.
+    handleEnvoyFilterError();
+    return FilterHeadersStatus::Continue;
+  }
+
+  // Normal reply, not an error.
+  return FilterHeadersStatus::Continue;
+}
+
+void Filter::handleEspv2FilterError() {
+  google::rpc::Status error = utils::getErrorFilterState(
+      *encoder_callbacks_->streamInfo().filterState());
+
+  if (config_->shouldScrubDebugDetails()) {
+    // This is a copy, will not affect the filter state.
+    error.clear_details();
+  }
+
+  // TODO(nareddyt): respond with JSON.
+}
+
+void Filter::handleEnvoyFilterError() {
+  /*
+   * TODO(nareddyt)
+   *
+   * - Read the envoy filter error text from:
+   * -> Body for HTTP/JSON response
+   * -> `grpc-message` header for gRPC response (maybe this can be ignored?)
+   *
+   * - Also need to get the status code from:
+   * -> HTTP status header for HTTP/JSON response
+   * -> `grpc-status` header for gRPC response (maybe this can be ignored?)
+   *
+   * - Create a basic google.rcp.Status from the message and code.
+   *
+   * - Set the google.rpc.Status in the filter state so access log can log it.
+   *
+   * - Respond with JSON.
+   */
+}
+
+}  // namespace error_translator
+}  // namespace http_filters
+}  // namespace envoy
+}  // namespace espv2
--- a/src/envoy/http/error_translator/filter.cc
+++ b/src/envoy/http/error_translator/filter.cc
@@ -53,15 +53,13 @@ FilterHeadersStatus Filter::encodeHeaders(
 
   if (Envoy::Grpc::Common::hasGrpcContentType(headers)) {
     is_grpc_response_ = true;
-  }
 
-  if (isEspv2FilterError() && is_grpc_response_) {
-    // gRPC error is already in filter state. No need to translate body, as it
-    // is not used for gRPC errors. Hence, nothing to do.
-    return FilterHeadersStatus::Continue;
-  }
+    if (isEspv2FilterError()) {
+      // gRPC error is already in filter state. No need to translate body, as it
+      // is not used for gRPC errors. Hence, nothing to do.
+      return FilterHeadersStatus::Continue;
+    }
 
-  if (is_grpc_response_) {
     // We keep the response body as is, but modify the filter state.
     // This will be used by the access log (SC Report).
     // This handles the trailers-only response by `sendLocalReply`.
--- a/src/envoy/http/error_translator/filter.cc
+++ b/src/envoy/http/error_translator/filter.cc
@@ -18,9 +18,12 @@
 
 #include "absl/types/optional.h"
 #include "common/buffer/buffer_impl.h"
+#include "common/common/assert.h"
+#include "common/grpc/common.h"
 #include "common/grpc/status.h"
 #include "common/http/headers.h"
 #include "src/envoy/utils/filter_state_utils.h"
+#include "src/google/protobuf/util/json_util.h"
 
 namespace espv2 {
 namespace envoy {
@@ -30,7 +33,16 @@ namespace error_translator {
 using ::Envoy::Http::FilterDataStatus;
 using ::Envoy::Http::FilterHeadersStatus;
 
-Filter::Filter(FilterConfigSharedPtr config) : config_(config) {}
+Filter::Filter(FilterConfigSharedPtr config) : config_(config) {
+  type_helper_ = std::make_unique<google::grpc::transcoding::TypeHelper>(
+      Envoy::Protobuf::util::NewTypeResolverForDescriptorPool(
+          Envoy::Grpc::Common::typeUrlPrefix(), &descriptor_pool_));
+
+  print_options_.add_whitespace = true;
+  print_options_.always_print_primitive_fields = true;
+  print_options_.always_print_enums_as_ints = false;
+  print_options_.preserve_proto_field_names = false;
+}
 
 FilterHeadersStatus Filter::encodeHeaders(
     Envoy::Http::ResponseHeaderMap& headers, bool) {
@@ -39,37 +51,13 @@ FilterHeadersStatus Filter::encodeHeaders(
     return FilterHeadersStatus::Continue;
   }
 
-  if (headers.ContentType() != nullptr &&
-      headers.ContentType()->value().getStringView() ==
-          Envoy::Http::Headers::get().ContentTypeValues.Grpc) {
+  if (Envoy::Grpc::Common::hasGrpcContentType(headers)) {
     is_grpc_response_ = true;
   }
 
   if (isEspv2FilterError() && is_grpc_response_) {
-    // Do not translate for gRPC, body does not matter.
-    // gRPC error is already in filter state as well, so nothing to do.
-    return FilterHeadersStatus::Continue;
-  }
-
-  if (isEspv2FilterError()) {
-    // We need to modify the response body, but keep filter state as is.
-    ENVOY_LOG(debug, "Translating HTTP espv2 error headers");
-
-    // Store a copy and scrub details if needed.
-    error_ = utils::getErrorFilterState(
-        *encoder_callbacks_->streamInfo().filterState());
-    if (config_->shouldScrubDebugDetails()) {
-      error_.clear_details();
-    }
-
-    // Translate to JSON.
-    error_json_ = errorToJson(error_);
-
-    // Update content headers.
-    headers.setContentLength(error_json_.size());
-    headers.setReferenceContentType(
-        Envoy::Http::Headers::get().ContentTypeValues.Json);
-
+    // gRPC error is already in filter state. No need to translate body, as it
+    // is not used for gRPC errors. Hence, nothing to do.
     return FilterHeadersStatus::Continue;
   }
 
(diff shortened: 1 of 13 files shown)


src/envoy/http/error_translator/filter.h
```diff
@@ -0,0 +1,50 @@
...
+namespace http_filters {
+namespace error_translator {
+
+class Filter : public Envoy::Http::PassThroughEncoderFilter,
+               public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
+ public:
+  Filter(FilterConfigSharedPtr config);
+
+  // Envoy::Http::StreamDecoderFilter
+  Envoy::Http::FilterHeadersStatus encodeHeaders(
```
qiwzhang, line 36: I think we need encodeData()

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,89 @@
...
+namespace http_filters {
+namespace error_translator {
+
+using ::Envoy::Http::FilterHeadersStatus;
+
+Filter::Filter(FilterConfigSharedPtr config) : config_(config) {}
+
+// TODO(nareddyt): encodeHeaders or encodeData?
+// Probably need both for `handleEnvoyFilterError`
+FilterHeadersStatus Filter::encodeHeaders(Envoy::Http::ResponseHeaderMap&,
```
qiwzhang, line 34: This is tricky,  we need to study more envoy code here.  

In general,  in encodeHeader we should
if there is our rpcStatus, we should Stop, otherwise continue

in encodeData() or encodeTrailer() we should 
* generate our JSON body, and change headers content-type and content-length and continue

But not sure if this works for the data from "sendLocalReply",  it should

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  // TODO(nareddyt): respond with JSON.
+}
+
+void Filter::handleEnvoyFilterError() {
+  /*
+   * TODO(nareddyt)
+   *
+   * - Read the envoy filter error text from:
```
qiwzhang, line 70: For handleEnvoyError, we need a reliable way to identify the data from"sendLocalRelay",  and
if content-type is "plain/text" since the default "sendLocalReply"  is using this type, then,

* read the body, and wrap it to JSON as
{ 
  code: http_status
  message:  the body
}

we could not use grpc-status and grpc-message since it is on the trailers.  It is not easy to modify the body data in encodeTrailer() call

src/envoy/http/error_translator/filter.cc
qiwzhang: We need to skip is_grpc:

Just read this [code](https://cs.corp.google.com/piper///depot/google3/third_party/envoy/src/source/common/http/utility.cc?rcl=307319856&g=0&l=435):

We should not write response body to grpc request.  

we can use content-type to detect grpc.

src/envoy/http/error_translator/filter.cc
qiwzhang: We can follow this envoy code:

https://github.com/envoyproxy/envoy/blob/master/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L732

src/envoy/http/error_translator/filter.cc
nareddyt: Done as described above.

src/envoy/http/error_translator/filter.h
nareddyt: Done

src/envoy/http/error_translator/filter.cc
nareddyt: Since the trailers-only response uses `encodeHeaders`, I decided to also add logic for this. I think it's important, otherwise SC Report will be difficult. With the new logic here, we can just do an SC report using all the details in the filter state.

Let me know if that makes it too complex, we can remove it.

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+FilterHeadersStatus Filter::encodeHeaders(
+    Envoy::Http::ResponseHeaderMap& headers, bool) {
+  if (isUpstreamResponse()) {
+    // Do not translate any response sent from the upstream.
+    return FilterHeadersStatus::Continue;
+  }
+
+  if (headers.ContentType() != nullptr &&
+      headers.ContentType()->value().getStringView() ==
+          Envoy::Http::Headers::get().ContentTypeValues.Grpc) {
```
qiwzhang, line 44: we can call this function

https://cs.corp.google.com/piper///depot/google3/third_party/envoy/src/source/common/grpc/common.cc?type=cs&rcl=307518108&g=0&l=28

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+    headers.setReferenceContentType(
+        Envoy::Http::Headers::get().ContentTypeValues.Json);
+
+    return FilterHeadersStatus::Continue;
+  }
+
+  if (is_grpc_response_) {
+    // We keep the response body as is, but modify the filter state.
+    // This will be used by the access log (SC Report).
+    // This handles the trailers-only response by `sendLocalReply`.
```
qiwzhang, line 79: Is this error from our ESPv2 filters,  if so,  we already have rpc.Status in the filterState.  Do we need to do this?

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+    // Do not translate body for gRPC. It is not used.
+    return FilterDataStatus::Continue;
+  }
+
+  if (isEspv2FilterError()) {
+    ENVOY_LOG(debug, "Translating HTTP espv2 error body");
+    GOOGLE_DCHECK(!error_.message().empty());
+    GOOGLE_DCHECK(!error_json_.empty());
+
+    if (!end_stream) {
```
qiwzhang, line 138: end_stream should be always true.  We can assert it.

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  if (is_grpc_response_) {
+    // Do not translate body for gRPC. It is not used.
+    return FilterDataStatus::Continue;
+  }
+
+  if (isEspv2FilterError()) {
+    ENVOY_LOG(debug, "Translating HTTP espv2 error body");
+    GOOGLE_DCHECK(!error_.message().empty());
```
qiwzhang, line 135: Use envoy

RELEASE_ASSERT(X, DETAILS)

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+    if (!end_stream) {
+      // Keep buffering data until we have the full response.
+      // TODO(nareddyt): Is there a better way to discard the remaining data? We
+      // don't care about it.
+      return FilterDataStatus::StopIterationAndBuffer;
+    }
+
+    // Replace the body with the JSON error.
+    Envoy::Buffer::OwnedImpl json_body(error_json_);
+    body.move(json_body);
```
qiwzhang, line 147: is "move" the same as "swap"?  we need "swap" to override it.

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+  // Craft the error. Translate to canonical status codes.
+  const unsigned int http_code =
+      encoder_callbacks_->streamInfo().responseCode().value_or(500);
+  error_.set_code(Envoy::Grpc::Utility::httpToGrpcStatus(http_code));
+
+  const std::string body_string = body.toString();
+  error_.set_message(body_string);
+
+  // Set in the filter state for the access log (SC Report).
+  utils::setErrorFilterState(*encoder_callbacks_->streamInfo().filterState(),
```
qiwzhang, line 171: we can use the original rpc.Status in the filterState.   Why do we create a new one here?

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,211 @@
...
+
+  // Update content headers.
+  headers_->setContentLength(error_json_.size());
+  headers_->setReferenceContentType(
+      Envoy::Http::Headers::get().ContentTypeValues.Json);
+
+  // Replace the body with the JSON error.
+  Envoy::Buffer::OwnedImpl json_body(error_json_);
+  body.move(json_body);
+  return FilterDataStatus::Continue;
```
qiwzhang, line 185: If we can make this work (where encodeHeaders() return Stop, and encodeData() update response_headers and override response_body),  I like both cases to go this flow;   espv2_errors and non_espv2_errors.

Such, the code is simpler.

src/envoy/http/error_translator/filter.cc
qiwzhang: There is one case that we can write rpc.Status to our filterState:

non_espv2  && non_grpc:

the error message is in the response body,  we need to convert it from plain text to JSON,

In order to send this error message in the Report, we can write it to stateFilter too.

src/envoy/http/error_translator/filter.cc
nareddyt: Done

src/envoy/http/error_translator/filter.cc
nareddyt: Good point on it adding latency. Can we put it here for now and move it out in a different PR? Otherwise, I'm worried about the added complexity. Anyways, this code is not used yet.

Re comment #2 above, yup. I plan to keep this case writing the filter state, this is the end of `encodeData`.

src/envoy/http/error_translator/filter.cc
nareddyt: Done

src/envoy/http/error_translator/filter.cc
nareddyt: Done

src/envoy/http/error_translator/filter.cc
nareddyt: I assume it is like the move constructor, but it's not clear to me by looking at the interface. I do not see any way to swap or clear a buffer.

I will verify via unit tests.

src/envoy/http/error_translator/filter.cc
nareddyt: This is the non-espv2 && non-grpc case. There is no `google.rpc.Status` in the filter state, we create one based on the HTTP body in `encodeData()` here.

src/envoy/http/error_translator/filter.cc
nareddyt: Good point, it's a little easier to read now.

src/envoy/http/error_translator/filter.h
```diff
@@ -43,22 +46,24 @@ class Filter : public Envoy::Http::PassThroughEncoderFilter,
...
-  google::rpc::Status error_;
-
-  // Caches the error converted to JSON.
-  std::string error_json_;
-
   // Save the response headers so they can be modified in `encodeData`.
   Envoy::Http::ResponseHeaderMap* headers_;
+
+  // Required for resolving error details (protobuf.Any types).
+  Envoy::Protobuf::DescriptorPool descriptor_pool_;
```
qiwzhang, line 62: pool is just like a cache.  It make sense to store it in a global config so the cache can be shared. 
Here if we store it in filter, it could not be shared.

src/envoy/http/error_translator/filter.cc
```diff
@@ -30,7 +33,16 @@ namespace error_translator {
 using ::Envoy::Http::FilterDataStatus;
 using ::Envoy::Http::FilterHeadersStatus;
 
-Filter::Filter(FilterConfigSharedPtr config) : config_(config) {}
+Filter::Filter(FilterConfigSharedPtr config) : config_(config) {
+  type_helper_ = std::make_unique<google::grpc::transcoding::TypeHelper>(
+      Envoy::Protobuf::util::NewTypeResolverForDescriptorPool(
+          Envoy::Grpc::Common::typeUrlPrefix(), &descriptor_pool_));
+
+  print_options_.add_whitespace = true;
```
qiwzhang, line 41: print options can come from filter config?

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,194 @@
...
+
+Filter::Filter(FilterConfigSharedPtr config) : config_(config) {
+  type_helper_ = std::make_unique<google::grpc::transcoding::TypeHelper>(
+      Envoy::Protobuf::util::NewTypeResolverForDescriptorPool(
+          Envoy::Grpc::Common::typeUrlPrefix(), &descriptor_pool_));
+
+  print_options_.add_whitespace = true;
+  print_options_.always_print_primitive_fields = true;
+  print_options_.always_print_enums_as_ints = false;
+  print_options_.preserve_proto_field_names = false;
```
qiwzhang, line 44: above 3 fields can just use default.

src/envoy/http/error_translator/filter.cc
```diff
@@ -0,0 +1,194 @@
...
+    return FilterHeadersStatus::Continue;
+  }
+
+  if (Envoy::Grpc::Common::hasGrpcContentType(headers)) {
+    is_grpc_response_ = true;
+
+    if (isEspv2FilterError()) {
+      // gRPC error is already in filter state. No need to translate body, as it
+      // is not used for gRPC errors. Hence, nothing to do.
+      return FilterHeadersStatus::Continue;
```
qiwzhang, line 60: need to return StopIteration, otherwise it will be too late to modify response headers in encodeData()

Actually,  by reading envoy sendLocalReply() code,   if is_grpc,  it will not call encodeData() since it is trailer-only.

src/envoy/utils/filter_state_utils.cc
```diff
@@ -25,13 +27,20 @@ using ::Envoy::Router::StringAccessor;
 using ::Envoy::Router::StringAccessorImpl;
 using ::Envoy::StreamInfo::FilterState;
 
+// FilterState container needed to store the google.rpc.Status error proto.
+struct RpcStatusWrapper : public Envoy::StreamInfo::FilterState::Object {
+  google::rpc::Status status_;
+
+  Envoy::ProtobufTypes::MessagePtr serializeAsProto() const override {
```
qiwzhang, line 34: I don't think we need this. It is only used for access_log to write filerState

src/envoy/utils/filter_state_utils.cc
```diff
@@ -25,13 +27,20 @@ using ::Envoy::Router::StringAccessor;
 using ::Envoy::Router::StringAccessorImpl;
 using ::Envoy::StreamInfo::FilterState;
 
+// FilterState container needed to store the google.rpc.Status error proto.
+struct RpcStatusWrapper : public Envoy::StreamInfo::FilterState::Object {
+  google::rpc::Status status_;
```
qiwzhang, line 32: use class, and hike status_ in the private and create a copy constructor.

src/envoy/utils/filter_state_utils.cc
```diff
@@ -44,6 +53,24 @@ absl::string_view getStringFilterState(
...
+  state->status_ = status;
+  filter_state.setData(kErrorRpcStatus, std::move(state),
+                       Envoy::StreamInfo::FilterState::StateType::ReadOnly);
+}
+
+bool hasErrorFilterState(Envoy::StreamInfo::FilterState& filter_state) {
+  return filter_state.hasData<RpcStatusWrapper>(kErrorRpcStatus);
+}
+
+google::rpc::Status getErrorFilterState(
```
qiwzhang, line 68: not to make a copy in "get" function,   and return "const &"

if caller wants to make a copy,  it can copy it on its own.

src/envoy/utils/filter_state_utils.h
```diff
@@ -35,6 +38,14 @@ absl::string_view getStringFilterState(
     const Envoy::StreamInfo::FilterState& filter_state,
     absl::string_view data_name);
 
+void setErrorFilterState(Envoy::StreamInfo::FilterState& filter_state,
+                         const google::rpc::Status& status);
+
+bool hasErrorFilterState(Envoy::StreamInfo::FilterState& filter_state);
```
qiwzhang, line 44: we should be able to pass in const & since we are not modifying it.  same for the next function

src/envoy/utils/filter_state_utils.cc
nareddyt: Done

src/envoy/utils/filter_state_utils.cc
nareddyt: As discussed offline, no need (?)

src/envoy/utils/filter_state_utils.cc
nareddyt: Well we need to keep the object, so may as well keep this function in case a user configured the access log format. It doesn't add any complexity

PR #128, opened 2020-04-22
Simplify auth code used to generate a JWT from a service account

nareddyt:
- Remove symmetric key logic, service accounts are private keys.
- Remove JSON parsing, can just pass string directly.
- Change interface to accept cpp constructs.


src/api_proxy/auth/auth_token.cc
```diff
@@ -15,188 +15,36 @@
...
-  }
-  return grpc_base64_encode(res, res_len, 1, 0);
-}
+  const char* token = grpc_jwt_encode_and_sign(&json_key, audience.c_str(),
+                                         TOKEN_LIFETIME, nullptr);
+  grpc_auth_json_key_destruct(&json_key);
 
-string DotConcat(const string &str1, const string &str2) {
-  return str1 + "." + str2;
+  return absl::optional<std::string>{token};
```
qiwzhang, line 45: I think you need to free token memory

nareddyt: Good catch. Also confirmed the ASAN presubmit caught this. Done

src/api_proxy/auth/auth_token.cc
```diff
@@ -15,188 +15,42 @@
...
+absl::optional<std::string> get_auth_token(const std::string& json_secret,
+                                           const std::string& audience) {
+  grpc_auth_json_key json_key =
+      grpc_auth_json_key_create_from_string(json_secret.c_str());
 
-  char *header = GenerateJoseHeader("HS256");
-  if (header == nullptr) {
-    gpr_log(GPR_ERROR, "Unable to create JOSE header");
-    return nullptr;
+  if (grpc_auth_json_key_is_valid(&json_key) == 0) {
```
TAOXUY, line 41: Are they equivalent? Is json_key.type checked?
- if (strcmp(json_key.type, GRPC_AUTH_JSON_TYPE_INVALID) != 0) { 
- if (grpc_auth_json_key_is_valid(&json_key) == 0) {

nareddyt: Yup: https://github.com/grpc/grpc/blob/dd8259f0eefcfc0743646bdc278a16b2cfad8b17/src/core/lib/security/credentials/jwt/json_token.cc#L66

PR #129, opened 2020-04-22
b/154758301: add access_log and access_log_format

TAOXUY:
Flag parity with ESP

It can be like [this](https://paste.googleplex.com/6400135833583616).

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -414,6 +414,15 @@ environment variable or by passing "-k" flag to this script.
         '''
     )
     parser.add_argument(
+        '--access_log_format',
+        help='''
+        String format to specify the format of access log. For the detailed
+        rules, please refer to the following document.
+        https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log#format-strings
+        '''
+    )
+
+    parser.add_argument(
         '--disable_tracing',
         action='store_true',
         default=False,
@@ -652,6 +661,9 @@ def enforce_conflict_args(args):
             # for non gcp case, disable tracing if tracing project id is not provided.
             args.disable_tracing = True
 
+    if not args.access_log and args.access_log_format:
+        return "Flag --access_log_format has to be used together with --access_log."
+
     if args.ssl_port and args.ssl_server_cert_path:
         return "Flag --ssl_port is going to be deprecated, please use --ssl_server_cert_path only."
     if args.tls_mutual_auth and args.ssl_client_cert_path:
@@ -824,6 +836,10 @@ def gen_proxy_config(args):
     if args.access_log:
         proxy_conf.extend(["--access_log",
                            args.access_log])
+    if args.access_log_format:
+        proxy_conf.extend(["--access_log_format",
+                           args.access_log_format])
+
     if args.disable_tracing:
         proxy_conf.append("--disable_tracing")
 
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -408,6 +408,12 @@ environment variable or by passing "-k" flag to this script.
         Must be >= 0 and the default is 5 if not set.
         ''')
     parser.add_argument(
+        '--access_log',
+        help='''
+        Path to a local file to which the access log entries will be written.
+        '''
+    )
+    parser.add_argument(
         '--disable_tracing',
         action='store_true',
         default=False,
@@ -815,6 +821,9 @@ def gen_proxy_config(args):
     if args.underscores_in_headers:
         proxy_conf.append("--underscores_in_headers")
 
+    if args.access_log:
+        proxy_conf.extend(["--access_log",
+                           args.access_log])
     if args.disable_tracing:
         proxy_conf.append("--disable_tracing")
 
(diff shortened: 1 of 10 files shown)


src/go/configgenerator/listener_generator.go
```diff
@@ -246,6 +222,56 @@ func makeListener(serviceInfo *sc.ServiceInfo) (*v2pb.Listener, error) {
...
+			RouteConfig: route,
+		},
+		UseRemoteAddress:  &wrapperspb.BoolValue{Value: opts.EnvoyUseRemoteAddress},
+		XffNumTrustedHops: uint32(opts.EnvoyXffNumTrustedHops),
+	}
+
+	if opts.AccessLog != "" {
+		fileAccessLog, _ := ptypes.MarshalAny(&facpb.FileAccessLog{
+			Path:            opts.AccessLog,
+			AccessLogFormat: nil,
```
qiwzhang, line 244: What is the default format?   can we have flag for users to change the format?

tests/integration_test/access_log_test.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil
+	}
+	return os.Remove(path)
+}
+
+func TestAccessLog(t *testing.T) {
+	t.Parallel()
+
+	accessLog := platform.GetFilePath(platform.AccessLog)
+	if err := tryRemoveFile(accessLog); err != nil {
```
JLXIA, line 55: Do you need to remove the file before and after the test?

tests/integration_test/access_log_test.go
```diff
@@ -0,0 +1,97 @@
...
+		"--rollout_strategy=fixed", "--access_log=" + accessLog}
+
+	s := env.NewTestEnv(comp.TestAccessLog, platform.EchoSidecar)
+
+	if err := s.Setup(args); err != nil {
+		t.Fatalf("fail to setup test env, %v", err)
+	}
+
+	reqCnt := uint32(10)
+	for i := uint32(0); i < reqCnt; i++ {
```
JLXIA, line 70: Do you want to do it in parallel?

tests/integration_test/access_log_test.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	s.TearDown()
+
+	gotReqCnt, err := lineCounter(accessLog)
+	if err != nil {
+		t.Fatalf("fail to get line count in access file: %v", err)
+	}
+	if reqCnt != gotReqCnt {
```
JLXIA, line 90: what is wrote into the access log?  only one line for each req?

tests/env/components/ports.go
```diff
@@ -29,7 +29,8 @@ import (
 
 // All integration tests should be listed here to get their test ids
 const (
-	TestAsymmetricKeys uint16 = iota
+	TestAccessLog uint16 = iota
```
JLXIA, line 32: not need

TAOXUY: Keep the alphabeta order.

tests/integration_test/access_log_test.go
TAOXUY: It is according to your format

tests/integration_test/access_log_test.go
TAOXUY: I was thinking to keep them but they are lots of temporary data in the log, which is hard to keep a sample to compare, so I remove those log items and it is too simple to be worth being kept,  compared to the envoy's doc.

docker/generic/start_proxy.py
```diff
@@ -407,6 +407,23 @@ def make_argparser():
         Set the retry times for service control Report request.
         Must be >= 0 and the default is 5 if not set.
         ''')
+    parser.add_argument(
+        '--access_log',
```
qiwzhang, line 411: does "/dev/stdout" work for access_log?

If so,  should we default to /dev/stdout so that Envoy access log will show up in the stack driver logs.  I know it is showing something when deployed in Cloud Run, but I guess it is not from Envoy.   I am wondering if it shows anything when deployed in GKE.

TAOXUY: Yes, it works. Those should be the request log from [Cloud Run](https://cloud.google.com/run/docs/logging?hl=en) but they are very simple [ones](https://screenshot.googleplex.com/OYNUNKxPbWn).

I think we need be careful about this since:
- it will make duplicate log entries on Cloud Run
- the report logs in endpoints basically cover most needs though this access_log can log the headers you want which is cool

Yeah, we can think about it later.

nareddyt: Agreed that this is basically a duplication of our SC Reports...

tests/integration_test/access_log_test.go
```diff
@@ -0,0 +1,95 @@
...
+	return os.Remove(path)
+}
+
+func makeOneRequest(t *testing.T, s *env.TestEnv) {
+	wantResp := `{"message":"hello"}`
+	url := fmt.Sprintf("http://localhost:%v/echo?key=api-key", s.Ports().ListenerPort)
+	resp, err := client.DoPost(url, "hello")
+
+	if err != nil {
+		t.Errorf("got unexpected error: %s", err)
```
JLXIA, line 44: Fatalf

PR #130, opened 2020-04-23
Add filter state to handle ESPv2 errors

nareddyt:
Split off from PR #127, see that for context.

### DIFF
--- a/src/api_proxy/auth/auth_token.cc
+++ b/src/api_proxy/auth/auth_token.cc
@@ -21,7 +21,7 @@
 #include "src/api_proxy/auth/grpc_internals.h"
 
 extern "C" {
-  #include "grpc/support/alloc.h"
+#include "grpc/support/alloc.h"
 };
 
 namespace espv2 {
--- a/src/envoy/utils/filter_state_utils.cc
+++ b/src/envoy/utils/filter_state_utils.cc
@@ -29,7 +29,6 @@ using ::Envoy::StreamInfo::FilterState;
 
 // FilterState container needed to store the google.rpc.Status error proto.
 class RpcStatusWrapper : public Envoy::StreamInfo::FilterState::Object {
-
  public:
   RpcStatusWrapper(const google::rpc::Status& status) : status_(status) {}
   RpcStatusWrapper(google::rpc::Status&& status) : status_(std::move(status)) {}
@@ -80,7 +79,8 @@ bool hasErrorFilterState(const Envoy::StreamInfo::FilterState& filter_state) {
   return filter_state.hasData<RpcStatusWrapper>(kErrorRpcStatus);
 }
 
-const google::rpc::Status& getErrorFilterState(const Envoy::StreamInfo::FilterState& filter_state) {
+const google::rpc::Status& getErrorFilterState(
+    const Envoy::StreamInfo::FilterState& filter_state) {
   const auto& state =
       filter_state.getDataReadOnly<RpcStatusWrapper>(kErrorRpcStatus);
   return state.getStatus();
--- a/src/envoy/utils/filter_state_utils.cc
+++ b/src/envoy/utils/filter_state_utils.cc
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 #include "src/envoy/utils/filter_state_utils.h"
+#include <memory>
 
 #include "common/common/empty_string.h"
 #include "common/router/string_accessor_impl.h"
+#include "envoy/stream_info/filter_state.h"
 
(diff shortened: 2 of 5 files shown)


src/api_proxy/auth/auth_token.cc
```diff
@@ -21,7 +21,7 @@
 #include "src/api_proxy/auth/grpc_internals.h"
 
 extern "C" {
-  #include "grpc/support/alloc.h"
+#include "grpc/support/alloc.h"
```
qiwzhang, line 24: this should not in this pr

src/envoy/utils/filter_state_utils.cc
```diff
@@ -44,6 +54,25 @@ absl::string_view getStringFilterState(
   return filter_state.getDataReadOnly<StringAccessor>(data_name).asString();
 }
 
+void setErrorFilterState(Envoy::StreamInfo::FilterState& filter_state,
+                         const google::rpc::Status& status) {
+  auto state = std::make_unique<RpcStatusWrapper>();
```
qiwzhang, line 59: if the wrapper has a copy constructor, these 2 lines can be combined into one

src/envoy/utils/filter_state_utils.cc
```diff
@@ -44,6 +54,25 @@ absl::string_view getStringFilterState(
   return filter_state.getDataReadOnly<StringAccessor>(data_name).asString();
 }
 
+void setErrorFilterState(Envoy::StreamInfo::FilterState& filter_state,
+                         const google::rpc::Status& status) {
```
qiwzhang, line 58: maybe pass in rpc::Status&& so it can be either swap-ed or moved?  try to see if we can avoid copy

src/envoy/utils/filter_state_utils.cc
```diff
@@ -44,6 +54,25 @@ absl::string_view getStringFilterState(
...
+  auto state = std::make_unique<RpcStatusWrapper>();
+  state->status_ = status;
+  filter_state.setData(kErrorRpcStatus, std::move(state),
+                       Envoy::StreamInfo::FilterState::StateType::ReadOnly);
+}
+
+bool hasErrorFilterState(Envoy::StreamInfo::FilterState& filter_state) {
```
qiwzhang, line 65: use "const FilterState&" for this function and next.

src/api_proxy/auth/auth_token.cc
nareddyt: This was from `clang-format`. Let's fix it in this PR

src/envoy/utils/filter_state_utils.cc
nareddyt: Done

src/envoy/utils/filter_state_utils.cc
nareddyt: Done, but I had to duplicate code. What do you think?

src/envoy/utils/filter_state_utils.cc
nareddyt: Done

src/envoy/utils/filter_state_utils.cc
```diff
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 #include "src/envoy/utils/filter_state_utils.h"
+#include <memory>
```
TAOXUY, line 16: Can we leave space above to cluster headers?

src/envoy/utils/filter_state_utils.h
```diff
@@ -35,6 +38,17 @@ absl::string_view getStringFilterState(
     const Envoy::StreamInfo::FilterState& filter_state,
     absl::string_view data_name);
 
+void setErrorFilterState(Envoy::StreamInfo::FilterState& filter_state,
+                         const google::rpc::Status& status);
+
+void setErrorFilterState(Envoy::StreamInfo::FilterState& filter_state,
+                         google::rpc::Status&& status);
+
+bool hasErrorFilterState(const Envoy::StreamInfo::FilterState& filter_state);
```
TAOXUY, line 47: add const in the end?

src/envoy/utils/filter_state_utils.h
```diff
@@ -35,6 +38,17 @@ absl::string_view getStringFilterState(
...
+
+const google::rpc::Status& getErrorFilterState(
+    const Envoy::StreamInfo::FilterState& filter_state);
```
TAOXUY, line 50: ditto

PR #132, opened 2020-04-23
Enforce max path length limit

nareddyt:
Based on feedback from the security review, we should enforce a limit (even if GCLB already does). This will prevent stack overflows on non-gcp deployments and fix oss-fuzz failures.

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21510&q=esp-v2


src/envoy/http/path_matcher/filter.cc
```diff
@@ -31,6 +31,10 @@ namespace http_filters {
 namespace path_matcher {
 namespace {
 
+// Half of the max header value size Envoy allows.
+// 4x the standard browser request size.
+uint32_t PATH_MAX_SIZE = 8192;
```
qiwzhang, line 36: please use constexpr

src/envoy/http/path_matcher/filter.cc
```diff
@@ -47,6 +51,9 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
   } else if (!headers.Path()) {
     rejectRequest(Envoy::Http::Code(400), "No path in request headers.");
     return Envoy::Http::FilterHeadersStatus::StopIteration;
+  } else if (headers.Path()->value().size() > PATH_MAX_SIZE) {
+    rejectRequest(Envoy::Http::Code(400), "Path is too long.");
```
qiwzhang, line 55: also print our limit in the error message.

src/envoy/http/path_matcher/filter.cc
nareddyt: Done

src/envoy/http/path_matcher/filter.cc
nareddyt: Done

PR #133, opened 2020-04-24
b/154877853: add dns_resolver_address

TAOXUY:
Add the flag --dns_resolver_address for --[dns](https://github.com/cloudendpoints/esp/blob/0704a6f23eb4685b0e6b88e54be5245de8964fd1/start_esp/start_esp.py#L740) in esp.


tests/env/components/ports.go
```diff
@@ -222,7 +224,8 @@ func NewPorts(testId uint16) *Ports {
 		DiscoveryPort:             base + 3,
 		AdminPort:                 base + 4,
 		FakeStackdriverPort:       base + 5,
-		JwtRangeBase:              base + 6,
+		DnsResolverPort:           base + 6,
```
nareddyt, line 227: Also need to change `portNum` on line 140

tests/integration_test/dns_resolver_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/GoogleCloudPlatform/esp-v2/tests/env"
+	comp "github.com/GoogleCloudPlatform/esp-v2/tests/env/components"
+	"github.com/GoogleCloudPlatform/esp-v2/tests/env/platform"
+)
+
+func fullDns(dns string) string {
+	return dns + "."
+}
+
+func TestDnsResolver(t *testing.T) {
```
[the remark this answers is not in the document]
nareddyt: Also test how Envoy handles the case where DNS resolve is unavailable. Does it prevent startup? Or fail during API traffic?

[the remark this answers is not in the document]
TAOXUY: That is a good point. Add the failed case where envoy will reject request by `503 unhealthy upstream` after failing to resolve the domain name.

tests/env/components/ports.go
TAOXUY: Thanks!

docker/generic/start_proxy.py
```diff
@@ -496,6 +496,18 @@ def make_argparser():
         the location of the service account credentials JSON file. If the option is
         omitted, the proxy contacts the metadata service to fetch an access token.
         '''.format(creds_key=GOOGLE_CREDS_KEY))
+
+    parser.add_argument(
+        '--dns_resolver_address',
+        help='''
+        The IP address or the IP address followed with port number used to
```
qiwzhang, line 503: maybe just:

IP address of the DNS resolver

src/go/configgenerator/cluster_generator.go
```diff
@@ -196,6 +198,24 @@ func makeJwtProviderClusters(serviceInfo *sc.ServiceInfo) ([]*v2pb.Cluster, erro
 	return providerClusters, nil
 }
 
+func parseDnsResolverAddress(address string) (string, uint32, error) {
+	arr := strings.Split(address, ":")
```
qiwzhang, line 202: can we re-use util.ParseURL() 

or 
create a new function ParseHostPort() so the code can be shared

src/go/configgenerator/cluster_generator.go
```diff
@@ -205,6 +225,26 @@ func makeBackendCluster(opt *options.ConfigGeneratorOptions, brc *sc.BackendRout
 		LoadAssignment:       util.CreateLoadAssignment(brc.Hostname, brc.Port),
 	}
 
+	if opt.DnsResolverAddress != "" {
+		host, port, err := parseDnsResolverAddress(opt.DnsResolverAddress)
+		if err != nil {
+			return nil, fmt.Errorf("fail to parse dnsResolverAddress: %v", err)
+		}
+
+		c.DnsResolvers = []*corepb.Address{
```
qiwzhang, line 234: It seems that Envoy support multiple resolvers.   Should we support multiple?

My thought:  Most likely, users only need to use one.  it will be very very rare need to have 2.  
But since it is easy to support multiple,  why don't we just support it?
my concern is if we design this flag to only support one,  later if we want to support multiple with the same flag, it may break, not backward compatible.

src/go/configgenerator/cluster_generator_test.go
```diff
@@ -579,6 +579,134 @@ func TestMakeBackendRoutingCluster(t *testing.T) {
...
+					},
+				},
+			},
+			BackendAddress:     "http://127.0.0.1:80",
+			DnsResolverAddress: "127.0.0.1:53",
+			wantedCluster: &v2pb.Cluster{
+				Name:                 "bookstore.endpoints.project123.cloud.goog_local",
+				ConnectTimeout:       ptypes.DurationProto(20 * time.Second),
+				ClusterDiscoveryType: &v2pb.Cluster_Type{v2pb.Cluster_LOGICAL_DNS},
+				LoadAssignment:       util.CreateLoadAssignment("127.0.0.1", 80),
```
qiwzhang, line 613: This cluster config looks weird,  cluster address is already IP,  it doesn't need DNS resolver. 

maybe we can put in a check,  if host is ready IP,  not need to add DNS resolver.

src/go/configmanager/flags/flags.go
```diff
@@ -60,6 +60,12 @@ var (
 	SslMaximumProtocol = flag.String("ssl_maximum_protocol", "", "Maximum TLS protocol version for Downstream connections.")
 	RootCertsPath      = flag.String("root_certs_path", util.DefaultRootCAPaths, "Path to the root certificates to make TLS connection.")
 	EnableHSTS         = flag.Bool("enable_strict_transport_security", false, "Enable HSTS (HTTP Strict Transport Security).")
+	DnsResolverAddress = flag.String("dns_resolver_address", "", `The IP address or the IP address followed with port number used to
```
qiwzhang, line 63: can we simplify the help text?

src/go/configgenerator/cluster_generator.go
qiwzhang, line 234: We need to add dns_resovler for all clusters;  service control,  jwks_url,  service_management, iam

src/go/configgenerator/cluster_generator_test.go
TAOXUY: Why not leave envoy to handle IP address by itself?

src/go/configgenerator/cluster_generator.go
TAOXUY: add to all clusters

TAOXUY: Done

src/go/configgenerator/cluster_generator.go
TAOXUY: This method is so hard to share so still put in dns_resolver file

docker/generic/start_proxy.py
TAOXUY: Use `address of DNS resolver` since users can assign PORT also.

TAOXUY:
Done with the functional part and integration test. Will add unit test later. 

Now once the flag is set, the dns resolvers will be added to all clusters. For those clusters using IP addresses, it fine since:
- envoy will know it is IP and won't try to resolve it.
- If we don't configure it, the default [dns resolver](https://github.com/envoyproxy/envoy/blob/0b0213fdc38eb0d0346659cf90bd4c502cadb00c/api/envoy/api/v2/cluster.proto#L674) will be configured anyway.

PTAL

docker/generic/start_proxy.py
```diff
@@ -498,12 +498,14 @@ def make_argparser():
...
     parser.add_argument(
-        '--dns_resolver_address',
+        '--dns_resolver_addresses',
         help='''
-        The IP address or the IP address followed with port number used to
-        resolve the domain name of backend. It should be in format of IP_ADDR
-        or IP_ADDR:PORT. For the IP_ADDR case, the default DNS port 52 will be
-        used.
+        The addresses of dns resolvers. Each address should be in format of
+        IP_ADDR or IP_ADDR:PORT and they are separated by '.';. For the IP_ADDR
```
qiwzhang, line 504: the example in the help text make me believe I can specify two ip addresses by
--dns_resolver_addresses=a.a.a.a  --dns_resolver_addresses=b.b.b.b

Is this supported,  if so,  it will be great. 

If no, please don't use this in the example.  

There is a typo in separated by '.'

TAOXUY: Done

TAOXUY:
PTAL

docker/generic/start_proxy.py
```diff
@@ -496,6 +496,18 @@ def make_argparser():
         the location of the service account credentials JSON file. If the option is
         omitted, the proxy contacts the metadata service to fetch an access token.
         '''.format(creds_key=GOOGLE_CREDS_KEY))
+
+    parser.add_argument(
+        '--dns_resolver_addresses',
```
JLXIA, line 501: also add --dns for espv1 compability.

TAOXUY: Done

PR #137, opened 2020-04-28
Token refresh logic for invalid expiry times

nareddyt:
- If the token has a very low refresh time, do not init ready until the retry finishes.
- Add extra unit tests for timestamp and expiry time parsing.
- Add extra unit test for retry on low refresh time.

### DIFF
--- a/src/envoy/token/token_subscriber.cc
+++ b/src/envoy/token/token_subscriber.cc
@@ -74,20 +74,18 @@ void TokenSubscriber::handleSuccessResponse(
     absl::string_view token, const std::chrono::seconds& expires_in) {
   active_request_ = nullptr;
 
-  ENVOY_LOG(debug, "{}: Got token with expiry duration: {} , {} sec",
-            debug_name_, token, expires_in.count());
-
-  // If the token will expire soon, then don't signal ready.
-  if (expires_in <= kRefreshBuffer) {
-    refresh();
-    return;
-  }
-
-  refresh_timer_->enableTimer(expires_in - kRefreshBuffer);
-
   // Signal that we are ready for initialization.
+  ENVOY_LOG(debug, "{}: Got token and expiry duration: {} , {} seconds",
+            debug_name_, token, expires_in.count());
   callback_(token);
   init_target_->ready();
+
+  if (expires_in <= kRefreshBuffer) {
+    // Handle low expiry time by retrying immediately.
+    refresh();
+  } else {
+    refresh_timer_->enableTimer(expires_in - kRefreshBuffer);
+  }
 }
 
 void TokenSubscriber::refresh() {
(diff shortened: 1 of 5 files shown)


src/envoy/token/token_subscriber.cc
```diff
@@ -75,16 +76,17 @@ void TokenSubscriber::handleSuccessResponse(
 
   ENVOY_LOG(debug, "{}: Got token with expiry duration: {} , {} sec",
             debug_name_, token, expires_in.count());
-  callback_(token);
 
-  // Use the buffer to set the next refresh time.
+  // If the token will expire soon, then don't signal ready.
```
qiwzhang, line 80: My concern is the service token behaviors:  If we try to fetch the token again 5 seconds before it is expired,   server may send use the same token,  server will only try to generate a new token when it is actual expired.

In this case, we are calling token server repeatedly until a new token.     Definitely it is a bad thing to do.  

I don't have a good plan to fix such corner case.  

Or this is not the case,   token server has a larger grace period,  it will give us a new token if we call it 5 seconds before it expired.

nareddyt: I am not aware of any such caching by IAM. I went through documentation and code, did not see anything.

Also note this command returns a different access token every time: 
```
gcloud auth print-access-token --impersonate-service-account <SA_ACCOUNT_EMAIL>
```

This may be calling a different endpoint.

In any case, would you prefer removing the retry() call directly and treating it as a failure (so we do a retry after the 2 second timer)?

nareddyt: If we do this, this will also prevent against stack overflows if a retry keeps occuring (ie. if the short expiry time token is cached). But I doubt this will happen in practice.

qiwzhang: I am sure GCP metadata server cache it.

nareddyt: Good idea

nareddyt: As discussed offline, we found out the token is cached. But a new one is generated within > 30 minutes of the token expiry time.

To resolve, we will allow tokens with low expiry time, but still retry.

src/envoy/token/token_subscriber_test.cc
```diff
@@ -340,6 +340,47 @@ TEST_F(TokenSubscriberTest, TokenSantizationCheckFails) {
   ASSERT_FALSE(init_ready_);
 }
 
+TEST_F(TokenSubscriberTest, TokenPastExpiryCheckFails) {
+  // Setup fake remote request.
+  Envoy::Http::RequestHeaderMapPtr req_headers(
+      new Envoy::Http::TestRequestHeaderMapImpl());
+  EXPECT_CALL(*info_, prepareRequest(token_url_))
+      .Times(1)
```
qiwzhang, line 348: It seems that all tests use the same test flow,  just wondering if we can use same code to share the flow. the parameters can be passed in.    I feel it will make test code more readable. 

This can be a separate PR.

PR #138, opened 2020-04-28
b/154830645: Verify second port available in GCS Runner

toddbeckman-g:
A second "Loopback" Listener can be used to utilize access logs on Service Control Filter's Check and Report calls and JWT Authn Filter's async token refresh. This requires opening a second port. GCSRunner needs to verify that a second port is available. This port can be specified with the LOOPBACK_PORT variable.

Note that only the PORT variable should be opened to public traffic.

Changes in this PR:

- Add support for FileAccessLog to JSON Marshal code. **Note: this change cannot be safely downgraded due to new configs being unable to parse with the former code.**
- Added LOOPBACK_PORT environment variable.
- Updated transformation code to support multiple listeners. **Note: this change cannot be safely downgraded due to the formerly hard-coded requirement of 1 listener.**
- If `loopback_listener` is provided, set the port number. If not, do nothing (this is required to allow compatibility with previous configs). One of `http_listener` or `https_listener` is still required.

Refactoring changes:

- Removed the ReplacePort field, as it added no functionality.

TESTED=Unit tests here and manual smoke test with full stack.

### DIFF
--- a/src/go/gcsrunner/main/runner.go
+++ b/src/go/gcsrunner/main/runner.go
@@ -28,6 +28,9 @@
 // Additionally, an optional `PORT` variable may be provided to override
 // where Envoy listens to traffic. This will be used only if the original config
 // specifies the port `8080`.
+//
+// Optionally, `LOOPBACK_PORT` may be used to configure Envoy configurations which
+// have an extra listener with the name "loopback_listener" to listen on this port.
 package main
 
 import (
@@ -48,7 +51,6 @@ const (
 	fetchGCSObjectInitialInterval = 10 * time.Second
 	fetchGCSObjectTimeout         = 5 * time.Minute
 	terminateEnvoyTimeout         = time.Minute
-	replaceListenerPort           = 8080
 )
 
 var (
@@ -59,16 +61,17 @@ var (
 
 func main() {
 	flag.Parse()
-
-	port := os.Getenv("PORT")
-	if port == "" {
-		port = "8080"
-	}
-	portNum, err := strconv.ParseUint(port, 10, 32)
+	port, err := envNum("PORT", 8080)
 	if err != nil {
-		glog.Fatal("PORT variable must be a valid port number.")
+		glog.Fatalf("Failed to get PORT number: %v", err)
 	}
-
+	loopbackPort, err := envNum("LOOPBACK_PORT", 8090)
+	if err != nil {
+		glog.Fatalf("Failed to get LOOPBACK_PORT number: %v", err)
+	}
+	if port == loopbackPort {
+		glog.Fatalf("PORT and LOOPBACK_PORT cannot be the same, got: (%d == %d)", port, loopbackPort)
+	}
 	bucketName := os.Getenv("BUCKET")
 	if bucketName == "" {
 		glog.Fatal("Must specify the BUCKET environment variable.")
@@ -100,8 +103,8 @@ func main() {
 	if err := gcsrunner.FetchConfigFromGCS(gcsrunner.FetchConfigOptions{
 		BucketName:                    bucketName,
 		ConfigFileName:                configFileName,
-		WantPort:                      uint32(portNum),
-		ReplacePort:                   replaceListenerPort,
+		WantPort:                      port,
+		LoopbackPort:                  loopbackPort,
 		FetchGCSObjectInitialInterval: fetchGCSObjectInitialInterval,
 		FetchGCSObjectTimeout:         fetchGCSObjectTimeout,
 		WriteFilePath:                 envoyConfigPath,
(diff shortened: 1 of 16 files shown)


src/go/gcsrunner/transform_config_test.go
```diff
@@ -254,6 +242,14 @@ var validConfigInput = []byte(`{
             ]
           }
         ]
+      },
+      {
+        "name": "loopback_listener",
```
qiwzhang, line 247: I don't think this is the real loopback listener.  this is just for test-only. is that right?

Who will be generating such loopback listener?   what will it be look like?

toddbeckman-g: The config here is just enough to test the code; it won't actually work.

I have a changelist with the loopback listener config generation code (b/153103031). It will be on our end, after calling configgenerator and before deploying it where GCS Runner to fetch it. Do we want an example of what this might look like, for the sake of OSS inclusiveness? I could probably document this package properly.

qiwzhang: OSS may not need to use it, but we will like to see how the loopback listener is configured.

src/go/gcsrunner/main/runner.go
```diff
@@ -68,7 +67,17 @@ func main() {
...
 	}
-
+	loopbackPort := os.Getenv("LOOPBACK_PORT")
+	if loopbackPort == "" {
+		loopbackPort = "8090"
+	}
+	lpNum, err := strconv.ParseUint(loopbackPort, 10, 32)
+	if err != nil {
+		glog.Fatal("LOOPBACK_PORT variable must be a valid port number.")
+	}
```
qiwzhang, line 77: It seem "PORT" and "LOOPBACK_PORT" are using the same code.  Can we use a function to share them?  I think it will make the code more readable.

src/go/gcsrunner/transform_config.go
```diff
@@ -101,29 +98,59 @@ func transformConfigBytes(config []byte, opts FetchConfigOptions) ([]byte, error
...
+		// This requires the use of another unused port.
+		case "loopback_listener":
+			if err := transformLoopbackListener(l, opts); err != nil {
+				return fmt.Errorf("failed to transform Loopback Listener")
+			}
+		}
 	}
-	if err := doListenerTransform(listeners[0], opts); err != nil {
+
+	if !httpListenerTransformed {
```
nareddyt, line 127: Should this also be checked for loopback listener? Sounds like you will always need this in your deployment

src/go/gcsrunner/transform_config.go
```diff
@@ -101,29 +98,59 @@ func transformConfigBytes(config []byte, opts FetchConfigOptions) ([]byte, error
...
+		return fmt.Errorf("expected at least 1 listener, got: 0")
+	}
+	// Require http_listener or https_listener, as the absence of these incidates
+	// a broken config.
+	// Note that loopback_listener is not required.
+	httpListenerTransformed := false
+	for _, l := range listeners {
+		switch l.GetName() {
+		// The default listener created by this project's configgenerator package.
+		case "http_listener", "https_listener":
```
qiwzhang, line 111: should we call it "ingress_listener", vs loopback_listner?  really don't care about http, https or grpc

src/go/gcsrunner/transform_config.go
toddbeckman-g: In the future possibly, but at the moment this will cause all configs without the new loopback listener to break. This will result in forward- and backward-incompatible changes, and it's virtually impossible to lock the versions together.

The migration plan here is to enable this runner to use both the old and the new configurations before updating the management plane to incorporate the loopback listener feature.

src/go/gcsrunner/transform_config.go
toddbeckman-g: The reason for this is because it's hard-coded into configgenerator. `ingress_listener` does make sense, but whatever name we decide, we might want to make this a constant shared between configgenerator and this package.

qiwzhang: If we don't have backward compatible issue, I like to make such change.

TAOXUY, line 111: Can you move "http_listener" and "https_listener" in listener_generator to util and make them shared?

TAOXUY:
Can we not amend to previous commit and force push? 
It is better to keep the commit history for review.

src/go/gcsrunner/transform_config_test.go
```diff
@@ -201,37 +179,52 @@ func TestReplaceListenerPort(t *testing.T) {
...
 		t.Errorf("doServiceControlTransform was not called")
 	}
 }
 
 var validConfigInput = []byte(`{
-	"static_resources": {
+"static_resources": {
     "listeners": [
       {
+      "name": "ingress_listener",
```
TAOXUY, line 224: In terms of your adding old listener name, maybe also test them here?

PR #139, opened 2020-04-28
[DO NOT SUBMIT] Test if access tokens are cached in IMDS

nareddyt:


qiwzhang:
My idea was to write a shell script to repeatedly call "curl" to get access token from metadata server and print its token and expiration time, just dump it to a log,  and run for a couple hours,  we will know

PR #141, opened 2020-04-29
Forward `Authorization` header in JWT Authn filter

### LINKED ISSUE #140: ESPv2 not forwarding original Authorization header when an x-google-backend is specified
Noticed this behaviour recently.

I'm using ESPv2 as a gateway to connect multiple cloud run services. As such, I have multiple x-google-backends specified in my API spec.

As per the docs, if ESP needs to overwrite the original Authorization header, it will forward the original one in the X-Forwarded-Authorization header.

However, when I log the headers received by my cloud run services this forwarded header is missing.

Any ideas?

Thanks :)

nareddyt:
Notes:
- Always forward `Authorization` header in frontend auth.
- This will work with the pre-existing backend auth `X-Forwarded-Authorization` logic.

Testing:
- Verified `Authorization` and `X-Forwarded-Authorization` headers for backend auth enabled or disabled (frontend auth always enabled).
- Also verify `X-Endpoint-API-UserInfo` header, seems we have no tests for this.

Fixes #140

### DIFF
--- a/tests/integration_test/jwt_auth_integration_test.go
+++ b/tests/integration_test/jwt_auth_integration_test.go
@@ -20,11 +20,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
 	"github.com/GoogleCloudPlatform/esp-v2/tests/endpoints/bookstore_grpc/client"
 	"github.com/GoogleCloudPlatform/esp-v2/tests/env"
 	"github.com/GoogleCloudPlatform/esp-v2/tests/env/platform"
 	"github.com/GoogleCloudPlatform/esp-v2/tests/env/testdata"
+	"github.com/GoogleCloudPlatform/esp-v2/tests/utils"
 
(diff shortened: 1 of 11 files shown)


tests/integration_test/jwt_auth_integration_test.go
```diff
@@ -239,3 +242,93 @@ func TestInvalidOpenIDConnectDiscovery(t *testing.T) {
...
+				"Authorization": "Bearer " + testdata.Es256Token,
+			},
+			wantHeaders: map[string]string{
+				"Authorization":           "Bearer " + testdata.Es256Token,
+				"X-Endpoint-API-UserInfo": testdata.Es256TokenPayloadBase64,
+			},
+		},
+		{
+			desc: "Frontend auth preserves `Authorization` and creates `X-Endpoint-API-UserInfo`." +
+				"Backend auth then modifies `Authorization` and creates `X-Forwarded-Authorization`.",
```
nareddyt, line 304: cc @fareeda0, this integration test should validate the request header flow you described in #140

PR #144, opened 2020-04-30
Destruct invalid json keys in grpc auth lib

nareddyt:
This is actually not needed, as `grpc_auth_json_key_create` will call `grpc_auth_json_key_destruct` if the key is invalid. But this improves readability (was confusing during security review). Calling it twice is safe.

### DIFF
--- a/src/api_proxy/auth/auth_token.cc
+++ b/src/api_proxy/auth/auth_token.cc
@@ -39,7 +39,8 @@ absl::optional<std::string> get_auth_token(const std::string& json_secret,
       grpc_auth_json_key_create_from_string(json_secret.c_str());
 
   if (grpc_auth_json_key_is_valid(&json_key) == 0) {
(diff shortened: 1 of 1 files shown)


src/api_proxy/auth/auth_token.cc
```diff
@@ -39,6 +39,7 @@ absl::optional<std::string> get_auth_token(const std::string& json_secret,
       grpc_auth_json_key_create_from_string(json_secret.c_str());
 
   if (grpc_auth_json_key_is_valid(&json_key) == 0) {
+    grpc_auth_json_key_destruct(&json_key);
```
qiwzhang, line 42: If it is not needed,  I prefer to add comment here, not to call the destruct function.

nareddyt: Done

PR #146, opened 2020-05-01
b/155213423: add latency in ServiceControl statistics

TAOXUY:


tests/integration_test/statistics_test.go
```diff
@@ -0,0 +1,134 @@
...
+			},
+			wantHistograms: map[string][]float64{
+				// For overhead_time, only verify the maximum(0 will be ignored).
+				"http.ingress_http.service_control.overhead_time": {0, 0, 0, 0, 0, 0, 0, 0, 0, 5},
+				"http.ingress_http.service_control.backend_time":  {1000, 1025, 1050, 1075, 1090, 1095, 1099, 1099.5, 1099.9, 1100},
+				"http.ingress_http.service_control.request_time":  {1000, 1025, 1050, 1075, 1090, 1095, 1099, 1099.5, 1099.9, 1100},
+			},
+		},
+		{
+			desc:        "Success after 10s due to ESPv2 default response timeout being 15s",
```
tomboc, line 66: Tests have the same desc. Is it intentional?

tests/integration_test/statistics_test.go
```diff
@@ -0,0 +1,134 @@
...
+				t.Errorf("Test (%s): failed, expected histogram %v not in the got histograms: %v", tc.desc, wantHistogramName, histograms)
+				break
+			} else if len(wantHistogramVals) != len(getHistogramVals) {
+				t.Errorf("Test (%s): failed, different value number for histogram %v, expected vals: %v , got vals: %v", tc.desc, wantHistogramName, wantHistogramVals, getHistogramVals)
+			} else {
+				for i, wantHistogramVal := range wantHistogramVals {
+					if wantHistogramVal == 0 {
+						continue
+					}
+					if wantHistogramVal/bufferPercent < getHistogramVals[i] || wantHistogramVal*bufferPercent > getHistogramVals[i] {
```
tomboc, line 126: Why does the check perform the division and multiplication here?

TAOXUY: Put it in the function to make it self-explanative.

tests/integration_test/statistics_test.go
```diff
@@ -0,0 +1,138 @@
...
+				t.Errorf("Test (%s): failed, expected histogram %v not in the got histograms: %v", tc.desc, wantHistogramName, histograms)
+				break
+			} else if len(wantHistogramVals) != len(getHistogramVals) {
+				t.Errorf("Test (%s): failed, different value number for histogram %v, expected vals: %v , got vals: %v", tc.desc, wantHistogramName, wantHistogramVals, getHistogramVals)
+			} else {
+				for i, wantHistogramVal := range wantHistogramVals {
+					if wantHistogramVal == 0 {
+						continue
+					}
+					if !roughEqual(getHistogramVals[i], wantHistogramVal, bufferPercent) {
```
qiwzhang, line 130: Not quite understand what we are comparing with.   If we are comparing the latency in ms,  it could be very flaky. very hard to control "bufferPercent"

I think just we just need to make sure number of buckets exist with non-zero data,   not need to actual check the data.

Just like Report,  we don't verify latency values in the report. It can vary dramatically.

TAOXUY: We can still do comparison here because we make the server delay response at least 1s to make the base very large and with  `bufferPercent` it can mitigate the overhead variation mainly introduced by our proxy, like under different build mode.

qiwzhang: SG

tests/integration_test/statistics_test.go
```diff
@@ -40,7 +40,7 @@ func TestStatistics(t *testing.T) {
 	if err := s.Setup(utils.CommonArgs()); err != nil {
 		t.Fatalf("fail to setup test env, %v", err)
 	}
-	const bufferPercent = 0.5
+	const bufferPercent = 0.8
```
qiwzhang, line 43: call it latencyMargin?

tests/integration_test/statistics_test.go
```diff
@@ -60,8 +60,8 @@ func TestStatistics(t *testing.T) {
 				"http.ingress_http.service_control.allowed":                        1,
 			},
 			wantHistograms: map[string][]float64{
-				// For overhead_time, only verify the maximum(0 will be ignored).
-				"http.ingress_http.service_control.overhead_time": {0, 0, 0, 0, 0, 0, 0, 0, 0, 5},
+				// For overhead_time, don't exact comparison since overhead time varies a log in different build mode.
```
qiwzhang, line 63: the latency is different each run.  Here the backend server introduces a fixed big 1 second latency, so the overall latency should be around 1 second.  This test compares the latency with expected 1 second with certain error margin.

TAOXUY: Done.

tests/endpoints/echo/client/client.go
```diff
@@ -26,78 +26,32 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/esp-v2/tests/env/testdata"
+	"github.com/GoogleCloudPlatform/esp-v2/tests/utils"
+	tpoauth2 "github.com/GoogleCloudPlatform/esp-v2/third_party/golang_internal"
```
nareddyt, line 30: Nit: Pre-existing go code in our repository prefers putting all named imports in a separate block, not merged into unnamed imports

tests/utils/stats.go
```diff
@@ -0,0 +1,86 @@
...
+	Values []Point `json:"values"`
+}
+
+// Point is the struct to decode the values of computed_quantiles.
+type Point struct {
+	Cumulative float64 `json:"cumulative,omitempty"`
+	Interval   float64 `json:"interval,omitempty"`
+}
+
+func ESpv2FiltersStatsPath() string {
```
nareddyt, line 56: Nit: Make this a const variable

tests/integration_test/statistics_test.go
```diff
@@ -0,0 +1,148 @@
...
+			},
+		},
+		{
+			desc:        "backend respond in 2s",
+			reqDuration: time.Second * 2,
+			wantCounts: map[string]int{
+				"http.ingress_http.backend_auth.token_added":                       2,
+				"http.ingress_http.backend_routing.append_path_to_address_request": 2,
+				"http.ingress_http.path_matcher.allowed":                           2,
+				"http.ingress_http.service_control.allowed":                        2,
```
nareddyt, line 81: So each test case is influenced by the statistics of the previous test case? This seems difficult to test.

Perhaps reset counters between each test? https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--reset_counters

If that does not reset the histograms, then just have our integration test suite run `Setup` for each test entry?

tests/endpoints/echo/client/client.go
nareddyt: Still needs to be fixed

PR #151, opened 2020-05-08
b/155212712: add statistical counter for check/allocateQuota/report call status

TAOXUY:


src/envoy/http/service_control/filter_stats.cc
```diff
@@ -0,0 +1,114 @@
...
+
+using ::google::protobuf::util::Status;
+using ::google::protobuf::util::error::Code;
+
+namespace espv2 {
+namespace envoy {
+namespace http_filters {
+namespace service_control {
+
+void ServiceControlFilterStats::collectCheckStatus(
```
nareddyt, line 27: Can we find a less redundant way of doing this? Templating would certainly help

How does Envoy do this? They collect stats per http code, right?

src/envoy/http/service_control/filter_stats.h
```diff
@@ -29,7 +30,41 @@ namespace service_control {
 // clang-format off
 #define ALL_SERVICE_CONTROL_FILTER_STATS(COUNTER, HISTOGRAM)     \
   COUNTER(allowed)                                    \
-  COUNTER(denied) \
+  COUNTER(denied)                          \
+  COUNTER(check_count_0) \
```
nareddyt, line 34: See if you can reduce code duplication. At the very least, you can use another macro to generate 0-16 for check vs report.

Also consider better naming instead of 0-16.

src/envoy/http/service_control/client_cache.cc
```diff
@@ -304,12 +307,25 @@ CancelFunc ClientCache::callCheck(
       request, response,
       [this, response, on_done](const Status& status) {
         CheckResponseInfo response_info;
```
nareddyt, line 309: FYI there is a bug here that I will create a PR for. I'll also re-order the logic here, it should be easier for you to track stats then.

src/envoy/http/service_control/client_cache.cc
```diff
@@ -304,12 +307,25 @@ CancelFunc ClientCache::callCheck(
...
+
+        // Transportation is successful so parse the checkResponse and use its
+        // status in callback.
         if (status.ok()) {
           Status converted_status = ::espv2::api_proxy::service_control::
               RequestBuilder::ConvertCheckResponse(
                   *response, config_.service_name(), &response_info);
+          ServiceControlFilterStats::collectCheckStatus(
```
nareddyt, line 317: I don't think this implements what is actually asked for in b/155209585. I believe they want to track each HTTP call to SC Check / Report, not the number of client API calls.

I am working on b/155209585 for what you seem to have implemented here.

TAOXUY: It is not b/155209585. b/155212712 simply wants to tracks the statuses of all the calls to SC.

nareddyt: Sorry I made a typo (edited original comment now). I don't think you implemented the ask in b/155212712.

For example here: https://github.com/GoogleCloudPlatform/esp-v2/pull/151/files#diff-dd87d549727ba1b173020a7adbdc168aR328

This will run for every consumer API call, not for every call to SC Check. If there are 5 API calls and only 1 Check call (because of the cache), you will record the metric 5 times instead of just once.

I may be misunderstanding your comment.

TAOXUY: Good catch. Done.

src/envoy/http/service_control/filter_stats.cc
TAOXUY: I don't see how templating can help cut down line of code. Could you give an example if you don't mind?

As I see in envoy's repo, they don't have the case of collecting all the status.

src/envoy/http/service_control/filter_stats.h
TAOXUY: For macro, 
- I thought it is not recommended in Google and for here it will be used once, so do we really need create a macro? What do you think? @qiwzhang 

For naming, I have double checked with 	
Tomaszand this is what they want.

qiwzhang: I prefer to name the counter properly,  it will be easier to read

src/envoy/http/service_control/filter_stats.h
```diff
@@ -42,6 +77,15 @@ namespace service_control {
 struct ServiceControlFilterStats {
   ALL_SERVICE_CONTROL_FILTER_STATS(GENERATE_COUNTER_STRUCT,
                                    GENERATE_HISTOGRAM_STRUCT)
+  // Collect check call status.
+  static void collectCheckStatus(
+      ServiceControlFilterStats& filter_stats,
+      const ::google::protobuf::util::error::Code& code);
+
+  // Collect report call status.
+  static void collectReportStatus(
```
qiwzhang, line 86: How about this idea?  
1) we split ALL_SERVICE_CONTROL_FILTER_STATS into multiple list
SERVICE_CONROL_FILTER_STATS with old stats
SERVICE_CONTROL_X_STATS for all counters, x could be check, report and quota

3) struct ServiceControlFilterStats   {
   ALL_SERVICE_CONTROL_FILTER_STATS()
  SERVICE_CONTROL_CHECK_STATS
  SERVICE_CONTROL_REPORT_STATS
  SERVICE_CONTROL_QUOTA_STATS
}

Will this work?

TAOXUY: I have tried. It seems not work  when you call
{ALL_SERVICE_CONTROL_FILTER_STATS(
        POOL_COUNTER_PREFIX(scope, final_prefix),
        POOL_HISTOGRAM_PREFIX(scope, final_prefix)),
SERVICE_CONTROL_CHECK_STATS(
        POOL_COUNTER_PREFIX(scope, final_prefix),
        POOL_HISTOGRAM_PREFIX(scope, final_prefix)),
SERVICE_CONTROL_REPORT_STATS(
        POOL_COUNTER_PREFIX(scope, final_prefix),
        POOL_HISTOGRAM_PREFIX(scope, final_prefix)),
}

src/envoy/http/service_control/filter_stats.h
TAOXUY: Done with use status name.

As for reuse macro, talking with Teju offline, seems not so trivial. Make it as a TODO.

src/envoy/http/service_control/filter_stats.h
qiwzhang: I think it can be done.  Please try it again.  In the worse case, we create 4 struct for stats.  You can think of each of them is just one `struct xxxStats`.   This base class hold 4 of them.  It can be done.

The problem is we need to pass them into handler and client_cache.  So it is easier to wrap 4 of them into a new struct.

TAOXUY: Done

src/envoy/http/service_control/filter_stats.h
```diff
@@ -27,47 +27,71 @@ namespace service_control {
...
-  COUNTER(report_count_ABORTED)                              \
-  COUNTER(report_count_OUT_OF_RANGE)                         \
-  COUNTER(report_count_UNIMPLEMENTED)                        \
-  COUNTER(report_count_INTERNAL)                             \
-  COUNTER(report_count_UNAVAILABLE)                          \
-  COUNTER(report_count_DATA_LOSS)                            \
+  HISTOGRAM(overhead_time, Milliseconds)
+
+// TODO(taoxuy): add macro function to init [check|quota|report]_count_STATUS
+#define SERVICE_CONTROL_CHECK_STATS(COUNTER, HISTOGRAM) \
```
qiwzhang, line 40: can we remove HISTOGRAM in this macro since it is not used?

TAOXUY: done

src/envoy/http/service_control/filter_stats.cc
```diff
@@ -0,0 +1,160 @@
...
+    case Code::DATA_LOSS:filter_stats.check_count_DATA_LOSS_.inc();
+      return;
+    case Code::UNAUTHENTICATED:filter_stats.check_count_UNAUTHENTICATED_.inc();
+      return;
+    default:
+      return;
+  }
+}
+
+void ServiceControlFilterStats::collectQuotaStatus(
```
qiwzhang, line 88: I believe the code for Check,Report, Quota can be shared without using macro and template.  
How about this?

For the stats struct, we can have:

Struct ServiceControlFilterStats {
         FilterStats   fiilter_;  // defined from SERVICE_CONTROL_CHECK_STATS
         StatusStats  check_;   // SERVICE_CONTROL_CHECK_STATS
         StatusStats  quota_;   // SERVICE_CONTROL_REPORT_STATS
         StatusStats  report_;
};

Then you only need to define one static function

static void collectStatusStats(  StatusStats& stats,   const Code& code) {}

In the check on_done lambda function, just call  collectStatusStats(filter_stats.check_, code);

qiwzhang: You may have to only define one macro

SERVICE_CONTROL_CODE_STATS

And use different prefix for Check, Quota, Report here

POOL_COUNTER_PREFIX(scope, final_prefix)

TAOXUY: DONE, PTAL.

src/envoy/http/service_control/client_cache.cc
```diff
@@ -120,6 +120,17 @@ class EnvoyPeriodicTimer
   Envoy::Event::TimerPtr timer_;
 };
 
+template <class Response>
+Status processScCallTransportStatus(const Status& status, Response* resp,
+                                    const std::string& body) {
+  if (status.ok()) {
+    if (!resp->ParseFromString(body)) {
+      return Status(Code::INVALID_ARGUMENT, std::string("Invalid response"));
```
nareddyt, line 128: Append the body in error response? Or log a warning

tests/integration_test/statistics_test.go
```diff
@@ -133,3 +136,128 @@ func TestStatistics(t *testing.T) {
...
+	serviceName := "bookstore-service"
+	configID := "test-config-id"
+	args := []string{"--service=" + serviceName, "--service_config_id=" + configID,
+		"--rollout_strategy=fixed"}
+
+	tests := []struct {
+		desc           string
+		reqCnt         int
+		checkRespCode  int
+		checkRespBody  *scpb.CheckResponse
```
nareddyt, line 152: Remove: `checkRespBody` not needed

nareddyt:
I can start working on b/155209585 once this is merged. Thanks!

TAOXUY:
Add `destruct_mode_` in client_cache to avoid segfault of accessing stats when filter is destructed.

Please take another look.

src/envoy/http/service_control/client_cache.cc
```diff
@@ -292,8 +302,10 @@ CancelFunc ClientCache::callCheck(
                                   const std::string& body) {
           Status final_status = processScCallTransportStatus<CheckResponse>(
               status, response, body);
-          ServiceControlFilterStats::collectCallStatus(filter_stats_.check_,
-                                                       final_status.code());
+          if (!destruct_mode_) {
+            ServiceControlFilterStats::collectCallStatus(filter_stats_.check_,
```
qiwzhang, line 306: How about moving collectCallStatus as ClientCache class member,  and do the destruct_mode checking in the function?

PR #153, opened 2020-05-08
Sync check response proto

nareddyt:
Low probability that any of these responses will occur, but let's keep it in sync.

Testing: unit tests


qiwzhang:
Do we need to update googleapis SHA [here](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/bazel/googleapis.bzl#L20)

nareddyt:
No, I double checked and the proto hasn't changed in 7 months. We just need to support the new codes that we didn't support previously.

qiwzhang:
I see.

PR #154, opened 2020-05-13
Assert path matcher query params are valid in fuzz test

nareddyt:
They are extracted from the path header, so they should be valid. But just add a check to confirm.

### DIFF
--- a/src/envoy/http/path_matcher/filter_fuzz_test.cc
+++ b/src/envoy/http/path_matcher/filter_fuzz_test.cc
@@ -4,6 +4,7 @@
 #include "test/mocks/server/mocks.h"
 
 #include "src/envoy/http/path_matcher/filter.h"
+#include "src/envoy/utils/filter_state_utils.h"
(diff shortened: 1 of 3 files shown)


src/envoy/http/path_matcher/filter_fuzz_test.cc
```diff
@@ -61,6 +62,12 @@ DEFINE_PROTO_FUZZER(
     // Run data against the filter.
     ASSERT_NO_THROW(doTest(filter, input));
 
+    // Ensure the query param filter state is valid.
+    absl::string_view query_params = utils::getStringFilterState(
+        *mock_decoder_callbacks.stream_info_.filter_state_,
+        utils::kQueryParams);
+    ASSERT_TRUE(Envoy::Http::validHeaderString(query_params));
```
qiwzhang, line 69: Are you always setting QueryParams?

nareddyt: No, but `utils::getStringFilterState` will return empty string when it's not set. Empty string is valid.

PR #155, opened 2020-05-13
b/155209585 #1: Refactor callback for testability and implement `denied_control_plane_fault`

nareddyt:
When HTTP call is unsuccessful, SC Check returns HTTP 500, or SC Check returns a check response with a 5xx-like status:
- For network fail open: Increment `allowed_control_plane_fault`
- For network fail closed: Increment `refactor_control_plane_fault`

The code changes for above are very simple (2 lines to increment a counter). Most of this diff is from splitting the callback into another function that we can unit test. Added unit tests for blocked/allowed requests and the stats.

### DIFF
--- a/src/envoy/http/service_control/client_cache.h
+++ b/src/envoy/http/service_control/client_cache.h
@@ -62,6 +62,8 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
       const ::google::api::servicecontrol::v1::ReportRequest& request);
 
  private:
+  friend class test::ClientCacheCheckResponseTest;
+
   // Ownership of CheckResponse is passed to this function.
   // The function will always call CheckDoneFunc.
   void handleCheckResponse(
@@ -117,8 +119,6 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
 
   // Used to retrieve the current time for tracing.
   Envoy::TimeSource& time_source_;
-
-  friend class test::ClientCacheCheckResponseTest;
 };
 
 }  // namespace service_control
--- a/src/envoy/http/service_control/client_cache.h
+++ b/src/envoy/http/service_control/client_cache.h
@@ -30,8 +30,10 @@ namespace envoy {
 namespace http_filters {
 namespace service_control {
 
-using ::google::api::servicecontrol::v1::CheckResponse;
-using ::google::protobuf::util::Status;
+// Forward declare friend class to test private functions.
+namespace test {
+class ClientCacheCheckResponseTest;
+}
 
 // The class to cache check and batch report.
 class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
@@ -52,11 +54,6 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
       const ::google::api::servicecontrol::v1::CheckRequest& request,
       Envoy::Tracing::Span& parent_span, CheckDoneFunc on_done);
 
-  // Ownership of CheckResponse is passed to this function.
-  // The function will always call CheckDoneFunc.
-  void handleCheckResponse(const Status& http_status, CheckResponse* response,
-                           CheckDoneFunc on_done);
-
   void callQuota(
       const ::google::api::servicecontrol::v1::AllocateQuotaRequest& request,
       QuotaDoneFunc on_done);
@@ -65,6 +62,13 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
       const ::google::api::servicecontrol::v1::ReportRequest& request);
 
  private:
+  // Ownership of CheckResponse is passed to this function.
+  // The function will always call CheckDoneFunc.
+  void handleCheckResponse(
+      const ::google::protobuf::util::Status& http_status,
+      ::google::api::servicecontrol::v1::CheckResponse* response,
+      CheckDoneFunc on_done);
+
   void initHttpRequestSetting(
       const ::google::api::envoy::http::service_control::FilterConfig&
           filter_config);
@@ -113,6 +117,8 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
 
   // Used to retrieve the current time for tracing.
   Envoy::TimeSource& time_source_;
+
+  friend class test::ClientCacheCheckResponseTest;
 };
 
 }  // namespace service_control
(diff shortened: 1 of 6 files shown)


src/envoy/http/service_control/client_cache.h
```diff
@@ -49,6 +52,11 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
       const ::google::api::servicecontrol::v1::CheckRequest& request,
       Envoy::Tracing::Span& parent_span, CheckDoneFunc on_done);
 
+  // Ownership of CheckResponse is passed to this function.
+  // The function will always call CheckDoneFunc.
+  void handleCheckResponse(const Status& http_status, CheckResponse* response,
```
nareddyt, line 57: Technically, this should not be part of the public interface. But I need to put it here for unit testing. Is this ok @qiwzhang?

I could move these into a separate class, but that seems like overkill for just 1 function.

src/envoy/http/service_control/client_cache_test.cc
```diff
@@ -0,0 +1,179 @@
...
+  };
+
+  CheckResponse* resp = new CheckResponse();
+  CheckError* check_error = resp->mutable_check_errors()->Add();
+  check_error->set_code(CheckError::NAMESPACE_LOOKUP_UNAVAILABLE);
+
+  const Status http_status(Code::OK, "");
+  cache_->handleCheckResponse(http_status, resp, on_done);
+
+  CheckAndReset(stats_base_.stats().filter_.allowed_control_plane_fault_, 1);
```
nareddyt, line 109: Do we need to add integration tests for this change? I feel stats is a trivial component of envoy that we can expect to work, and we have unit tests to ensure they are incremented. We already have integration tests for some stats, I don't see the point of integration testing every single stat we add.

What do you think @qiwzhang, @TAOXUY?

src/envoy/http/service_control/client_cache_test.cc
```diff
@@ -0,0 +1,179 @@
...
+        service_config_, filter_config_, stats_base_.stats(), cm_, time_source_,
+        dispatcher_, token_fn_, token_fn_);
+  }
+
+  void CheckAndReset(Envoy::Stats::Counter& counter, const int expected_value) {
+    EXPECT_EQ(counter.value(), expected_value);
+    counter.reset();
+  }
+
+  void TearDown() override {
```
TAOXUY, line 56: Seem unnecessary

src/envoy/http/service_control/client_cache_test.cc
TAOXUY: agree

src/envoy/http/service_control/client_cache_test.cc
```diff
@@ -0,0 +1,179 @@
...
+TEST_F(ClientCacheCheckResponseTest, Http5xxAllowed) {
+  CheckDoneFunc on_done = [](const Status& status, const CheckResponseInfo&) {
+    EXPECT_EQ(status.code(), Code::OK);
+  };
+
+  CheckResponse* resp = new CheckResponse();
+  const Status http_status(Code::UNAVAILABLE, "");
+  cache_->handleCheckResponse(http_status, resp, on_done);
+
+  CheckAndReset(stats_base_.stats().filter_.allowed_control_plane_fault_, 1);
```
TAOXUY, line 84: The main part of test can be put into a test separate function to be shared, which can even make it to be table-driven. Input with responseStatus, expectStatus and statsCounter.

TAOXUY: Same for ClientCacheCheckResponseNetworkFailClosedTest.

src/envoy/http/service_control/client_cache.h
```diff
@@ -30,6 +30,9 @@ namespace envoy {
 namespace http_filters {
 namespace service_control {
 
+using ::google::api::servicecontrol::v1::CheckResponse;
+using ::google::protobuf::util::Status;
```
qiwzhang, line 34: Per Google style guide,  not allow to use namespace "using" in header files.  It is easy to cause problem.

src/envoy/http/service_control/client_cache.h
qiwzhang: it is better not to pollute the public interfaces. 
see if we can move into "private" and use "friend class" to allow test class to access it.  
need to move test class out of annoymous name space.

src/envoy/http/service_control/client_cache.h
nareddyt: Done

src/envoy/http/service_control/client_cache_test.cc
nareddyt: I feel this is unnecessary, it will only save a few lines of code. And CheckResponse still needs to be constructed manually in each test.

Keep in mind table driven tests are go style, not c++ style. This is not recommended by gtest. We should be using test fixture parameters instead for gtest (which we do not in our codebase...)

src/envoy/http/service_control/client_cache_test.cc
nareddyt: This is needed to ensure we didn't set any other stats other than the ones in the test.

src/envoy/http/service_control/client_cache.h
nareddyt: Done with friend class.

src/envoy/http/service_control/client_cache.h
```diff
@@ -113,6 +117,8 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
 
   // Used to retrieve the current time for tracing.
   Envoy::TimeSource& time_source_;
+
+  friend class test::ClientCacheCheckResponseTest;
```
qiwzhang, line 121: Per style guide,  friend class should be right after "private"

nareddyt: Done. I did not see this in the style guide. But it seems to be a common recommendation.

https://google.github.io/styleguide/cppguide.html#Friends

src/envoy/http/service_control/client_cache_test.cc
TAOXUY: I think we are using test fixture parameters. 

It still helps to make code look clearer if we can extract the helper function. You can make `new response` in SetUp.

nareddyt: Done, thanks it's clearer now.

I did not move the `new response` to `SetUp`, I feel this hides what is being tested. Each test is clearer when it creates the response and passes it to the test function.

PR #156, opened 2020-05-15
b/155209585 #2: Implement `denied_consumer_blocked` and `denied_consumer_error` stats

nareddyt:
Implementation details:
- Modify `CheckResponseInfo` to have a summarized `ErrorType` instead of booleans.
- Increment the correct stat based on the `ErrorType`.

Testing:
- Unit tests to ensure `ErrorType` is set correctly.
- Unit tests to ensure correct stat is incremented for each `ErrorType`.

### DIFF
--- a/src/api_proxy/service_control/check_response_test.cc
+++ b/src/api_proxy/service_control/check_response_test.cc
@@ -29,174 +29,236 @@ namespace {
 
 Status ConvertCheckErrorToStatus(gasv1::CheckError::Code code,
                                  const char* error_detail,
-                                 const char* service_name) {
+                                 const char* service_name,
+                                 CheckResponseInfo* info) {
   gasv1::CheckResponse response;
   gasv1::CheckError* check_error = response.add_check_errors();
-  CheckRequestInfo info;
   check_error->set_code(code);
   check_error->set_detail(error_detail);
-  return RequestBuilder::ConvertCheckResponse(response, service_name, nullptr);
+  return RequestBuilder::ConvertCheckResponse(response, service_name, info);
 }
 
-Status ConvertCheckErrorToStatus(gasv1::CheckError::Code code) {
+Status ConvertCheckErrorToStatus(gasv1::CheckError::Code code,
+                                 CheckResponseInfo* info) {
   gasv1::CheckResponse response;
   std::string service_name;
   response.add_check_errors()->set_code(code);
-  return RequestBuilder::ConvertCheckResponse(response, service_name, nullptr);
+  return RequestBuilder::ConvertCheckResponse(response, service_name, info);
 }
 
 }  // namespace
 
 TEST(CheckResponseTest, AbortedWithInvalidArgumentWhenRespIsKeyInvalid) {
-  Status result = ConvertCheckErrorToStatus(CheckError::API_KEY_INVALID);
+  CheckResponseInfo info;
+  Status result = ConvertCheckErrorToStatus(CheckError::API_KEY_INVALID, &info);
   EXPECT_EQ(Code::INVALID_ARGUMENT, result.code());
+  EXPECT_EQ(info.error_type, CheckResponseErrorType::API_KEY_INVALID);
 }
 
 TEST(CheckResponseTest, AbortedWithInvalidArgumentWhenRespIsKeyExpired) {
-  Status result = ConvertCheckErrorToStatus(CheckError::API_KEY_EXPIRED);
+  CheckResponseInfo info;
+  Status result = ConvertCheckErrorToStatus(CheckError::API_KEY_EXPIRED, &info);
   EXPECT_EQ(Code::INVALID_ARGUMENT, result.code());
+  EXPECT_EQ(info.error_type, CheckResponseErrorType::API_KEY_INVALID);
 }
 
 TEST(CheckResponseTest,
      AbortedWithInvalidArgumentWhenRespIsBlockedWithNotFound) {
-  Status result = ConvertCheckErrorToStatus(CheckError::NOT_FOUND);
+  CheckResponseInfo info;
+  Status result = ConvertCheckErrorToStatus(CheckError::NOT_FOUND, &info);
   EXPECT_EQ(Code::INVALID_ARGUMENT, result.code());
+  EXPECT_EQ(info.error_type, CheckResponseErrorType::CONSUMER_ERROR);
 }
 
 TEST(CheckResponseTest,
      AbortedWithInvalidArgumentWhenRespIsBlockedWithKeyNotFound) {
-  Status result = ConvertCheckErrorToStatus(CheckError::API_KEY_NOT_FOUND);
+  CheckResponseInfo info;
+  Status result =
+      ConvertCheckErrorToStatus(CheckError::API_KEY_NOT_FOUND, &info);
   EXPECT_EQ(Code::INVALID_ARGUMENT, result.code());
+  EXPECT_EQ(info.error_type, CheckResponseErrorType::API_KEY_INVALID);
 }
 
 TEST(CheckResponseTest,
      AbortedWithPermissionDeniedWhenRespIsBlockedWithServiceNotActivated) {
-  Status result = ConvertCheckErrorToStatus(
-      CheckError::SERVICE_NOT_ACTIVATED, "Service not activated.", "api_xxxx");
+  CheckResponseInfo info;
+  Status result =
+      ConvertCheckErrorToStatus(CheckError::SERVICE_NOT_ACTIVATED,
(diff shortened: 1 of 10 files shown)


src/api_proxy/service_control/request_builder.cc
```diff
@@ -1371,62 +1371,82 @@ Status RequestBuilder::ConvertCheckResponse(
   const CheckError& error = check_response.check_errors(0);
   switch (error.code()) {
     case CheckError::NOT_FOUND:
+      check_response_info->error_type = CheckResponseErrorType::CONSUMER_ERROR;
```
TAOXUY, line 1374: Probably make it as a separate function containing the mapping logic for this.

src/api_proxy/service_control/request_info.h
```diff
@@ -103,18 +103,19 @@ struct CheckRequestInfo : public OperationInfo {
...
-  // If service is activated.
-  bool service_is_activated;
-  // Consumer project id
-  std::string consumer_project_id;
+  CheckResponseErrorType error_type = CheckResponseErrorType::UNKNOWN;
 
-  // By default api_key is valid and service is activated.
-  // They only set to false by the check response from server.
-  CheckResponseInfo() : is_api_key_valid(true), service_is_activated(true) {}
+  std::string consumer_project_id;
```
TAOXUY, line 118: keep the comment?

src/envoy/http/service_control/client_cache.cc
```diff
@@ -336,6 +337,20 @@ void ClientCache::handleCheckResponse(const Status& http_status,
...
+      case CheckResponseErrorType::CONSUMER_BLOCKED:
+        filter_stats_.filter_.denied_consumer_blocked_.inc();
+        break;
+      case CheckResponseErrorType::CONSUMER_ERROR:
+      case CheckResponseErrorType::SERVICE_NOT_ACTIVATED:
+      case CheckResponseErrorType::API_KEY_INVALID:
+        filter_stats_.filter_.denied_consumer_error_.inc();
+        break;
+      default:
+        break;
```
TAOXUY, line 351: This function is long enough so put this part into a function?

src/api_proxy/service_control/request_info.h
```diff
@@ -103,18 +103,19 @@ struct CheckRequestInfo : public OperationInfo {
...
+
 // Stores the information subtracted from the check response.
 struct CheckResponseInfo {
-  // If the request have a valid api key.
-  bool is_api_key_valid;
-  // If service is activated.
-  bool service_is_activated;
-  // Consumer project id
-  std::string consumer_project_id;
+  CheckResponseErrorType error_type = CheckResponseErrorType::UNKNOWN;
```
TAOXUY, line 116: Look like the default value would be `UNKNOWN` which doesn't fit the case of no error. 

Maybe use a unique_ptr to represent if it contains an error? Not sure about this.

src/api_proxy/service_control/request_builder.cc
```diff
@@ -1371,62 +1371,82 @@ Status RequestBuilder::ConvertCheckResponse(
...
       return Status(Code::INVALID_ARGUMENT,
                     "Client project not found. Please pass a valid project.");
     case CheckError::RESOURCE_EXHAUSTED:
```
qiwzhang, line 1377: Are you going to add counter denied_consumer_quota  here

Here from the feature request
service_conrol.denied_consumer_quota
       + Check response with RESOURCE_EXHAUSTED error code

src/envoy/http/service_control/client_cache_test.cc
```diff
@@ -148,6 +148,52 @@ TEST_F(ClientCacheCheckResponseNetworkFailClosedTest, Sc5xxBlocked) {
...
+    CheckDoneFunc on_done = [&](const Status&, const CheckResponseInfo&) {};
+    const Status http_status(Code::OK, Envoy::EMPTY_STRING);
+    cache_->handleCheckResponse(http_status, got_response, on_done);
+  }
+};
+
+TEST_F(ClientCacheCheckResponseErrorTypeTest, ConsumerBlocked) {
+  CheckResponse* response = new CheckResponse();
+  CheckError* check_error = response->mutable_check_errors()->Add();
+  check_error->set_code(CheckError::CLIENT_APP_BLOCKED);
```
qiwzhang, line 164: it seems that you could move above 3 lines into runTest()

src/api_proxy/service_control/request_info.h
nareddyt: Comment is no longer accurate / doesn't add any information

src/envoy/http/service_control/client_cache.cc
nareddyt: I think it's fine, it's a simple switch statement. I would like all the SC Check related stats incremented in one function (see a few lines below, where we increment stats for 5xx). Otherwise, it's difficult to reason about which code paths changed which stats.

src/api_proxy/service_control/request_info.h
nareddyt: Good point, protos suffer from the same problem. Let's follow https://aip.dev/126 and change `UNKNOWN` to `ERROR_TYPE_UNSPECIFIED`.

src/api_proxy/service_control/request_builder.cc
nareddyt: Yes, I will do that in another PR, the feature request is not done yet. 

I have some questions I need to clarify first. Specifically: 
- Shouldn't this counter be incremented in `AllocateQuota` calls, not `Check` calls?
- What is the difference between `CheckResponse::RESOURCE_EXHUASTED` and  `AllocateQuotaResponse::RESOURCE_EXHUASTED`?

src/envoy/http/service_control/client_cache_test.cc
nareddyt: Done

src/api_proxy/service_control/request_builder.cc
qiwzhang: Good question.   In ESF used CheckRequest,  it has quota request.  so CheckResponse::RESOURCE_EXHUASTED is for ESF.    For ESP,  we don't use it,  we call separate API AllocateQuota,   so we should only get AllocateQuotaResponse::RESOURCE_EXHUASTED,  not CheckResponse::RESOURCE_EXHUASTED

nareddyt: Makes sense, thanks

src/api_proxy/service_control/request_builder.cc
```diff
@@ -1371,62 +1371,82 @@ Status RequestBuilder::ConvertCheckResponse(
...
+      check_response_info->error_type = CheckResponseErrorType::CONSUMER_ERROR;
       return Status(Code::PERMISSION_DENIED, "Project has been deleted.");
     case CheckError::PROJECT_INVALID:
+      check_response_info->error_type = CheckResponseErrorType::CONSUMER_ERROR;
       return Status(Code::INVALID_ARGUMENT,
                     "Client project not valid. Please pass a valid project.");
     case CheckError::BILLING_DISABLED:
+      check_response_info->error_type = CheckResponseErrorType::CONSUMER_ERROR;
       return Status(Code::PERMISSION_DENIED,
                     std::string("API ") + service_name +
```
TAOXUY, line 1432: strCat?

src/api_proxy/service_control/check_response_test.cc
```diff
@@ -29,174 +29,236 @@ namespace {
...
-  EXPECT_EQ("", result.error_message());
+  EXPECT_EQ(info.error_type, CheckResponseErrorType::CONSUMER_ERROR);
 }
 
 TEST(CheckResponseTest, WhenResponseIsBlockedWithApiTargetBlocked) {
-  Status result = ConvertCheckErrorToStatus(CheckError::API_TARGET_BLOCKED);
+  CheckResponseInfo info;
+  Status result =
+      ConvertCheckErrorToStatus(CheckError::API_TARGET_BLOCKED, &info);
   EXPECT_EQ(Code::PERMISSION_DENIED, result.code());
```
TAOXUY, line 204: Can we also make a `runTest` for this part?

nareddyt: Done. 

I wonder if these tests are actually needed. In fact, the mappings defined in the test code could just be used in the production code to replace the switch/case, would be easier to read :)

src/api_proxy/service_control/request_builder.cc
nareddyt: Done

PR #158, opened 2020-05-18
Variety of fixes for file-based tests in google3

nareddyt:
Summary of changes:
- Static bootstrap test: Move all internal static bootstrap tests to a subdirectory of `examples/`. This is more maintainable.
- Config manager test: Use `platform.GetFilePath` to load test data.
- Config manager test: Decouple test from ` /etc/ssl/certs/ca-certificates.crt` because it is env dependent. Instead, copy the root CA certs file into `testdata` and use that.
- Marshal test: Don't assert the proto name, as the name changes in google3.
- Token generator test: Remove testing with a file, instead copy code to work with direct data. This is not ideal, but we cannot write to files in google3 during a test (easily).

Needed to unblock config versioning.


src/go/configmanager/config_manager_test.go
```diff
@@ -1803,6 +1801,7 @@ func runTest(t *testing.T, opts options.ConfigGeneratorOptions, f func(*testEnv)
 
 	metadataFetcher := metadata.NewMockMetadataFetcher(mockMetadataServer.URL, time.Now())
 
+	opts.RootCertsPath = platform.GetFilePath(platform.RootCaCerts)
```
JLXIA, line 1804: this is unit test, I don't think it really reads the file

tests/env/platform/files.go
```diff
@@ -58,10 +58,20 @@ const (
...
 	GrpcEchoEnvoyConfig
+
+	// Other configurations for testing
+	FixedDrServiceConfig
 )
 
+// go tests are not executed from the root fo the repository.
+// So the file path will change based on which test is using the file.
+// i.e. unit tests are more deeply nested that integration tests.
 var fileMap = map[RuntimeFile]string{
```
JLXIA, line 69: I remember the reason of adding this file is because we want to run integration test in g3, which is not used now. Should we get ride of this?

src/go/configmanager/config_manager_test.go
nareddyt: It does read the file, which is not hermetic.

https://github.com/GoogleCloudPlatform/esp-v2/blob/master/src/go/configmanager/config_manager.go#L312

tests/env/platform/files.go
nareddyt: We also use it to run our unit tests in google3. A variety of them depend on files, which this PR tries to fix. So we need to keep it.

src/go/configmanager/config_manager_test.go
JLXIA: -> platform.TestRootCaCerts

src/go/util/token_generator.go
```diff
@@ -31,19 +31,28 @@ var (
...
 
 	return generateAccessToken(data)
 }
 
+// A test-friendly version of `GenerateAccessTokenFromFile`
+func generateAccessTokenFromData(saData []byte) (string, time.Duration, error) {
+	if token, duration := activeAccessToken(); token != "" {
+		return token, duration, nil
+	}
+
```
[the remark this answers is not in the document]
nareddyt: No it was not working. Reading from the file is ok, but writing to it does not work. I just disabled the test in Google3 manually. But with the new config versioning implementation, disabling tests is too complex.

src/go/configmanager/config_manager_test.go
nareddyt: Done

src/go/util/token_generator.go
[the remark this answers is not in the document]
nareddyt: Agreed that adding code for that looks bad. Unfortunately, I'm not sure of any other workaround (other than to refactor the production code and unit test individual components, not the whole thing).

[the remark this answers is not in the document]
TAOXUY: I see. In this case, wrap `generateAccessTokenFromFile` by `GenerateAccessTokenFromFile` with comments, instead of copying code.

Only testing `GenerateAccessTokenFromFile` is fine.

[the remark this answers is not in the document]
nareddyt: That will change production code, it will require reading the SA file every time you check the cache. The only option is to split the unit test into one for the cache and one for the generation. What do you think?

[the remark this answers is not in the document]
TAOXUY: SG

PR #161, opened 2020-05-19
Improve readability/validation of backend routing config and filter

nareddyt:
During security review, there were some concerns about when backend rules are created and what the default path translation behavior is.

To address these:
- Change path translation into an enum, similar to service config BackendRule.
- Better validation/clarification of the enum in Config Manager and backend routing filter config.
- Split `decodeHeaders` into functions based on the path translation. This is easier to read and clarifies the cases where query params are needed.

Breaking change in filter config, but API Config Versioning should handle this. 

### DIFF
--- a/api/envoy/http/backend_routing/config.proto
+++ b/api/envoy/http/backend_routing/config.proto
@@ -22,15 +22,17 @@ message BackendRoutingRule {
   // Operation name, also known as selector.
   string operation = 1 [(validate.rules).string.min_bytes = 1];
 
-  // True if CONSTANT_ADDRESS, otherwise is APPEND_PATH_TO_ADDRESS.
+  // Corresponds to the path translation options in Google Cloud Endpoints.
   // https://cloud.google.com/endpoints/docs/openapi/openapi-extensions#understanding_path_translation
-  bool is_const_address = 2;
+  enum PathTranslation {
+    PATH_TRANSLATION_UNSPECIFIED = 0;
+    CONSTANT_ADDRESS = 1;
+    APPEND_PATH_TO_ADDRESS = 2;
+  }
+
+  PathTranslation path_translation = 2;
 
   // Prefix used when re-writing the path for the backend.
-  // Example for CONSTANT_ADDRESS:
-  //     https://us-central1-my-project-id.cloudfunctions.net/helloGET
-  // Example for APPEND_PATH_TO_ADDRESS:
-  //     https://my-project-id.appspot.com
   string path_prefix = 3 [(validate.rules).string = {
     well_known_regex: HTTP_HEADER_VALUE,
     strict: false
@@ -38,7 +40,7 @@ message BackendRoutingRule {
 }
 
 message FilterConfig {
-  // A list of backend routing rules for those selectors which needs path
+  // A list of backend routing rules for those selectors that needs path
   // translation.
   repeated BackendRoutingRule rules = 1;
 }
--- a/api/envoy/http/backend_routing/config.proto
+++ b/api/envoy/http/backend_routing/config.proto
@@ -19,6 +19,8 @@ package google.api.envoy.http.backend_routing;
 import "validate/validate.proto";
 
 message BackendRoutingRule {
+  reserved 2;
+
   // Operation name, also known as selector.
   string operation = 1 [(validate.rules).string.min_bytes = 1];
 
@@ -30,7 +32,7 @@ message BackendRoutingRule {
     APPEND_PATH_TO_ADDRESS = 2;
   }
 
-  PathTranslation path_translation = 2;
+  PathTranslation path_translation = 4;
 
   // Prefix used when re-writing the path for the backend.
   string path_prefix = 3 [(validate.rules).string = {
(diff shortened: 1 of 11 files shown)


api/envoy/http/backend_routing/config.proto
```diff
@@ -22,23 +22,25 @@ message BackendRoutingRule {
...
+  // Corresponds to the path translation options in Google Cloud Endpoints.
   // https://cloud.google.com/endpoints/docs/openapi/openapi-extensions#understanding_path_translation
-  bool is_const_address = 2;
+  enum PathTranslation {
+    PATH_TRANSLATION_UNSPECIFIED = 0;
+    CONSTANT_ADDRESS = 1;
+    APPEND_PATH_TO_ADDRESS = 2;
+  }
+
+  PathTranslation path_translation = 2;
```
qiwzhang, line 33: Please don't re-use same index for backward compatibility

src/envoy/http/backend_routing/filter.h
```diff
@@ -37,6 +37,16 @@ class Filter : public Envoy::Http::PassThroughDecoderFilter,
                                                  bool) override;
 
  private:
+  std::string translateConstPath(
```
TAOXUY, line 40: These two can be anonymous function

src/envoy/http/backend_routing/filter.cc
```diff
@@ -52,50 +53,74 @@ FilterHeadersStatus Filter::decodeHeaders(
...
+  if (!extracted_query_params.empty()) {
+    // Append extracted variable bindings.
+    absl::StrAppend(&new_path, "&", extracted_query_params);
+  }
+  return new_path;
+}
+
+std::string Filter::translateAppendPath(const BackendRoutingRule* rule,
+                                        absl::string_view original_path) {
+  // Just append the original request path to the configured path prefix.
```
TAOXUY, line 118: Put the comments in front of function name?

api/envoy/http/backend_routing/config.proto
nareddyt: Done

src/envoy/http/backend_routing/filter.h
nareddyt: You mean a lambda? I don't see the benefit of that.

src/envoy/http/backend_routing/filter.cc
nareddyt: Done

src/envoy/http/backend_routing/filter.h
TAOXUY: Sorry, function in anonymous space.

nareddyt: Oh, `translateConstPath` uses the filter state member variable. So it must be a member function. We should keep both here for consistency.

src/envoy/http/backend_routing/filter.cc
```diff
@@ -52,50 +54,76 @@ FilterHeadersStatus Filter::decodeHeaders(
...
   return FilterHeadersStatus::Continue;
 }
 
+// Replace the original path with the constant path prefix.
+// If any variable bindings were extracted, append them as query params.
+std::string Filter::translateConstPath(absl::string_view prefix,
+                                       absl::string_view original_path) {
+  const auto& filter_state = *decoder_callbacks_->streamInfo().filterState();
+  absl::string_view extracted_query_params =
+      utils::getStringFilterState(filter_state, utils::kQueryParams);
```
TAOXUY, line 94: Since this is the only place to use class member, can we inject extracted_query_params, so that move `translateConstPath` and `translateAppendPath` functions out of class interface to keep it simple? Just my 2 cents, what do you think?@qiwzhang

nareddyt: Is there any specific benefit to moving these to an anonymous namespace? If not, I feel it's fine to keep them as private functions in the class. Also it helps with readability to have all the functions declared in the header file. Up to you

api/envoy/http/backend_routing/config.proto
```diff
@@ -19,26 +19,30 @@ package google.api.envoy.http.backend_routing;
 import "validate/validate.proto";
 
 message BackendRoutingRule {
+  reserved 2;
```
qiwzhang, line 22: Are we going to make this a breaking change? if so,  it is fine.
we need to have a doc to list all deprecated fields.

If we need support backward compatible,  we may have to keep it and support it.

we also need a doc to list all fields that waiting to be deprecated.  so next break change can deprecate all of them.

src/envoy/http/backend_routing/filter.cc
qiwzhang: private member function is fine as I see reading FilterState is specific to ConstPath,  it should be here at this function.

api/envoy/http/backend_routing/config.proto
nareddyt: Yes this will be a breaking change. With API Config Versioning, we don't have to support both. 

What sort of doc are you thinking of having? I don't think API Gateway needs it, as they will use the correct version of config generator to generate the new config for the new envoy.

qiwzhang: Maybe just config_breaking_change.md in the top folder to list out the config version and which fields causing the breakage.

nareddyt: Ack

PR #163, opened 2020-05-21
b/155209585 #3: Implement `denied_consumer_quota` and `denied_producer_error`

nareddyt:
Implementation notes:
- Rename enum and re-use for check and quota responses.
- Add in a `QuotaResponseInfo` to serve as an intermediate translation of the codes.
- Move logic to increment stats into its own function to share between check and quota.
- Move logic to handle Quota responses into it's own function for testability.

Testing done:
- Similar to #156, add tests for the quota response.

### DIFF
--- a/src/envoy/http/service_control/client_cache.cc
+++ b/src/envoy/http/service_control/client_cache.cc
@@ -26,8 +26,9 @@ using ::google::api::envoy::http::service_control::FilterConfig;
 using ::google::protobuf::util::Status;
 using ::google::protobuf::util::error::Code;
 
-using ::espv2::api_proxy::service_control::CheckResponseErrorType;
 using ::espv2::api_proxy::service_control::CheckResponseInfo;
+using ::espv2::api_proxy::service_control::QuotaResponseInfo;
+using ::espv2::api_proxy::service_control::ScResponseErrorType;
 using ::google::api::servicecontrol::v1::AllocateQuotaRequest;
 using ::google::api::servicecontrol::v1::AllocateQuotaResponse;
 using ::google::api::servicecontrol::v1::CheckRequest;
@@ -289,6 +290,23 @@ ClientCache::ClientCache(
       config_.service_name(), config_.service_config_id(), options);
 }
 
+void ClientCache::handleScResponseErrorStats(ScResponseErrorType error_type) {
+  switch (error_type) {
+    case ScResponseErrorType::CONSUMER_BLOCKED:
+      filter_stats_.filter_.denied_consumer_blocked_.inc();
+      break;
+    case ScResponseErrorType::CONSUMER_ERROR:
+    case ScResponseErrorType::SERVICE_NOT_ACTIVATED:
+    case ScResponseErrorType::API_KEY_INVALID:
+      filter_stats_.filter_.denied_consumer_error_.inc();
+      break;
+    case ScResponseErrorType::CONSUMER_QUOTA:
+      filter_stats_.filter_.denied_consumer_quota_.inc();
+    default:
+      break;
+  }
+}
+
 CancelFunc ClientCache::callCheck(const CheckRequest& request,
                                   Envoy::Tracing::Span& parent_span,
                                   CheckDoneFunc on_done) {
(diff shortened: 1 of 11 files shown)


src/envoy/http/service_control/client_cache.cc
```diff
@@ -381,37 +388,50 @@ void ClientCache::handleCheckResponse(const Status& http_status,
...
+  if (http_status.ok()) {
+    QuotaResponseInfo response_info;
+    Status quota_status = ::espv2::api_proxy::service_control::RequestBuilder::
+        ConvertAllocateQuotaResponse(*response, config_.service_name(),
+                                     &response_info);
+
+    handleScResponseErrorStats(response_info.error_type);
+    on_done(quota_status);
+  } else {
+    // Most likely an auth error in ESPv2 or API producer deployment.
```
nareddyt, line 429: @qiwzhang: This branch should never occur, status should always be OK for Quota calls. I'll remove it in a different PR, I would like to add some integration tests to verify.

src/envoy/http/service_control/client_cache.cc
```diff
@@ -381,37 +388,50 @@ void ClientCache::handleCheckResponse(const Status& http_status,
     }
   } else {
     if (translate_non_5xx) {
```
qiwzhang, line 390: translate_non_5xx is kind of confusing for me.   how about final_status_is_from_http? is this correct?

src/envoy/http/service_control/client_cache.cc
```diff
@@ -381,37 +388,50 @@ void ClientCache::handleCheckResponse(const Status& http_status,
...
-        }
-        delete response;
-      });
+  client_->Quota(request, response,
+                 [this, response, on_done](const Status& status) {
+                   handleQuotaResponse(status, response, on_done);
+                 });
+}
+
+void ClientCache::handleQuotaResponse(const Status& http_status,
```
qiwzhang, line 417: we should call it handleQuotaOnDone.   This is called from or after cache.  It is different with QuotaResponse which I feel it is from quota_transport which is the actual response from the server.   Regardless, please add this comment.

e.g. response proto message it is from the cache,  except the first cache miss,  which is an empty one.

src/envoy/http/service_control/client_cache.cc
qiwzhang: this only happen when cache is disabled.  But we don't support disabling quota cache.  With cache,  the rate limiting is not very precious,   e.g. if limit is 300 p/m,   we may go 500 p/m.  So maybe some users may perfer long latency, but precious limit.

src/envoy/http/service_control/filter_stats.h
```diff
@@ -33,6 +33,8 @@ namespace service_control {
   COUNTER(denied_control_plane_fault)    \
   COUNTER(denied_consumer_blocked)       \
   COUNTER(denied_consumer_error)         \
+  COUNTER(denied_consumer_quota)         \
+  COUNTER(denied_producer_error)         \
```
qiwzhang, line 37: Can we add a doc to describe these statistics?  I don't want people to read code to understand how they are collected.

src/envoy/http/service_control/client_cache.cc
nareddyt: Ack. I feel no customers have such strict requirements, but we can follow-up later.

src/envoy/http/service_control/client_cache.cc
nareddyt: Correct. We only want to translate HTTP 4xx errors, not SC Check 4xx errors. I think I can simplify this actually, no need for the variable.

src/envoy/http/service_control/client_cache.cc
nareddyt: Done.

src/envoy/http/service_control/filter_stats.h
nareddyt: Sure, I'll add to the README.

src/envoy/http/service_control/client_cache.h
```diff
@@ -65,6 +67,12 @@ class ClientCache : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
  private:
   friend class test::ClientCacheCheckResponseTest;
   friend class test::ClientCacheCheckResponseErrorTypeTest;
+  friend class test::ClientCacheQuotaResponseTest;
+  friend class test::ClientCacheQuotaResponseErrorTypeTest;
+
+  // Increments the corresponding stat for the given error type.
+  void handleScResponseErrorStats(
```
TAOXUY, line 74: Maybe call `collectScResponseErrorStats`

nareddyt: Done

PR #164, opened 2020-05-26
Handle errors and test stats in Backend Routing

nareddyt:
Error handling:
- Errors should occur very rarely in the Backend Routing filter, but we must handle those cases for security review.
- Reject the requests when error occurs. This will prevent forwarding requests to localhost. It will also present clearer error messages to users.
- If no operation is found, still allow pass through.

Stats:
- Add some more stats, documented in README.
- Add unit tests for pre-existing stats.

Also cleanup uses of `ASSERT_EQ` to `EXPECT_EQ`.

### DIFF
--- a/src/envoy/http/backend_routing/README.md
+++ b/src/envoy/http/backend_routing/README.md
@@ -29,6 +29,8 @@ This filter records statistics.
 ### Counters
 
 - `denied`: Number of API Consumer requests that are denied due to invalid headers, etc.
+- `pass_through`: Number of API Consumer requests that are allowed through
+ without modification. Occurs when the request is not for dynamic routing.
 - `append_path_to_address_request`: Number of API Consumer requests that are
  accepted and translated as APPEND_PATH_TO_ADDRESS.
 - `constant_address_request`: Number of API Consumer requests that are
--- a/src/envoy/http/backend_routing/README.md
+++ b/src/envoy/http/backend_routing/README.md
@@ -28,9 +28,10 @@ This filter records statistics.
 
 ### Counters
 
-- `denied`: Number of API Consumer requests that are denied due to invalid headers, etc.
-- `pass_through`: Number of API Consumer requests that are allowed through
- without modification. Occurs when the request is not for dynamic routing.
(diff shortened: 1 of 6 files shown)


src/envoy/http/backend_routing/filter.cc
```diff
@@ -39,20 +47,25 @@ FilterHeadersStatus Filter::decodeHeaders(
   // NOTE: this shouldn't happen in practice because Path Matcher filter would
   // have already rejected the request.
   if (operation.empty()) {
-    ENVOY_LOG(debug, "No operation found from DynamicMetadata");
-    return FilterHeadersStatus::Continue;
+    rejectRequest(Envoy::Http::Code::InternalServerError,
+                  "No operation found from DynamicMetadata",
+                  RcDetails::get().BadRequest);
+    return FilterHeadersStatus::StopIteration;
```
qiwzhang, line 53: I think the design of this filter is:  passive.   If I don't not handle it, pass through.  In this way, it is easier to config this filter, it doesn't need to know all request patterns.  It only needs to know the request patterns that it need to act.

It makes this filter safer to be added.

nareddyt: I see, then let's pass through for this case and error for the other cases (that shouldn't happen).

Also seems like the integration tests didn't catch my error. I will attempt to make a follow-up PR that tests this, if we want to support both sidecar and dynamic routing modes at the same time.

src/envoy/http/backend_routing/README.md
```diff
@@ -20,4 +20,18 @@ This filter is designed to strongly integrate with the following filters:
...
\ No newline at end of file
+for inline documentation.
+
+## Statistics
+
+This filter records statistics.
+
+### Counters
+
+- `denied`: Number of API Consumer requests that are denied due to invalid headers, etc.
```
qiwzhang, line 31: I prefer to have one stat for each error,  can we have
denied_by_no_path
denied_by_no_operation
allowed_by_no_configured_rules

nareddyt: Done

PR #165, opened 2020-05-26
Update envoy to 2020-05-26

nareddyt:
Changes for fuzz tests:
- Rearrange code to reduce exception throwing (better performance).
- Reduce code duplication by using fuzzing library built into Envoy.
- Fuzzer supports generating protobuf data, modify corpus for new fields.

Changes for integration tests:
- Update sleep time due to new draining policies.
- Update fake DNS resolver to respond with correct return codes.

Changes for coverage:
- Update coverage script to use `bazel coverage`.
- Only run coverage on periodic jobs, as it does not work with the remote cache.

Changes to prow jobs:
- https://github.com/GoogleCloudPlatform/oss-test-infra/pull/368
- https://github.com/GoogleCloudPlatform/oss-test-infra/pull/369
- https://github.com/GoogleCloudPlatform/oss-test-infra/pull/371

TODOs:
- Re-enable coverage job for presubmits (nareddyt)


WORKSPACE
```diff
@@ -39,9 +39,9 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 # 3) Check if envoy_build_config/extensions_build_config.bzl is up-to-date.
 # Try to match it with the one in source/extensions and comment out unneeded extensions.
 
-ENVOY_SHA1 = "99706022206acb99bd8093c654fc84f6cd5a877e"  # 2020-04-06
+ENVOY_SHA1 = "74290ef76a76fbbf50f072dc33438791f93f68c7"  # 2020-05-23
```
qiwzhang, line 42: My pr https://github.com/envoyproxy/envoy/pull/11309 just merged.  Could you include it in?

nareddyt: Ack

nareddyt:
Seems like we need to update our prow images with `python3-distutils`. Will do tomorrow

nareddyt:
As expected, some integration tests failed for JWT Authn update.

Unsure why E2E tests are failing deployment. I imagine something wrong with the new image

nareddyt:
Hi @TAOXUY, PTAL at the dns integration test: https://github.com/GoogleCloudPlatform/esp-v2/pull/165/files#diff-d2a60decb0b332ac7a412bb489efd4c0

tests/integration_test/dns_resolver_test.go
```diff
@@ -25,63 +25,85 @@ import (
...
 		desc            string
+		backendHost     string
 		isResolveFailed bool
 		wantResp        string
 		wantError       string
 	}{
 		{
-			desc:     "resolve domain name successfully",
-			wantResp: `{"message":"hello"}`,
+			desc:        "resolve non-FQDN domain name successfully",
```
TAOXUY, line 43: IIUC, PQDN is tested here so maybe replace non-FQDN with PQDN.
http://www.tcpipguide.com/free/t_AbsoluteFullyQualifiedandRelativePartiallyQualifie.htm

tests/integration_test/dns_resolver_test.go
```diff
@@ -25,63 +25,85 @@ import (
...
+			backendHost: "dns-resolver-test-backend",
```
TAOXUY, line 44: For PQDN, it will append domain to the host. Can we get the domain in the test and put that in dns_resolver?

Like if we can somehow get `corp.google.com` is local domain, we put `dns-resolver-test-backend.corp.google.com` in dns_resolver and let envoy to discover it. Current case is not obvious.

tests/integration_test/dns_resolver_test.go
```diff
@@ -31,56 +31,79 @@ func toFqdnWithRoot(dns string) string {
...
+			wantResp:    `{"message":"hello"}`,
+		},
+		{
+			desc:        "resolve FQDN domain name successfully",
+			backendHost: "dns-resolver-test-backend.example.com",
+			wantResp:    `{"message":"hello"}`,
+		},
+		{
+			desc:        "resolve workstation FQDN domain name successfully",
+			backendHost: "dns-resolver-test-backend.corp.google.com",
```
qiwzhang, line 54: confused,  will this fail when the test is run in prow, or in the hosts with other domain names?

If no,  it is just a FQDN,  it could be any FQDN, then is this case cover by the one above with "dns-resolver-test-backend.example.com"

qiwzhang: Same question for the next test

nareddyt: Previously yet. Even when we used a FQDN like `dns-resolver-test-backend.corp.google.com`, it would work on our workstations but fail on prow. That is because the hostname in the prow job is `test-pods.svc.cluster.local`, so it would check for `dns-resolver-test-backend.corp.google.com.est-pods.svc.cluster.local`.

I put those cases here to make sure this test never fails on both our workstations and prow.

tests/integration_test/dns_resolver_test.go
nareddyt: We can, but that's not what we should be testing, right? We should ensure that Envoy will make calls to discover the PQDN instead of the FQDN. The other tests with FQDN test the case you are describing.

tests/integration_test/dns_resolver_test.go
nareddyt: Done

tests/integration_test/dns_resolver_test.go
TAOXUY: IIUC, all the dns requests sent from envoy are all FQDN by appending root dot to the full dns guess. I was trying to say whether we can test the process of envoy trying all the possible FQDN by appending domains like
- `dns-resolver-test-backend.`
- `dns-resolver-test-backend.XXX`
- `dns-resolver-test-backend.YYY`
-  `dns-resolver-test-backend.test-instance-domain.com.`
...

It can be achieved by setting `dns-resolver-test-backend` as backend and `dns-resolver-test-backend.test-instance-domain.com.`(we need know the domain of machine running the test) in the dns resolver. The backend host in other test cases look already have the full name so they don't cover this case.

nareddyt: I don't see the benefit of testing this, this is part of the DNS spec. As long as both sides are using a spec-compliant DNS resolver, it should be fine.

Also based on my testing, it doesn't just use the host name. Here are some example queries from Envoy on my workstation:
```
;dns-resolver-test-backend.prod.google.com
;dns-resolver-test-backend.prodz.google.com.
;dns-resolver-test-backend.google.com.
;dns-resolver-test-backend.svl.corp.google.com.
;dns-resolver-test-backend.
```

I don't know where it gets the first 2 from...

TAOXUY: In this case, this testCase has no difference with other testCase. We are expecting resolving `dns-resolver-test-backend.`, which is already a FQDN.

nareddyt: We spoke offline and decided it was worth testing it, since envoy seems to make different requests based on if it has the host name.

PR #169, opened 2020-05-29
b/157743299: Migrate configs from api/v2 to api/v3

TAOXUY:
- Migrate all the v2 proto to v3 proto if v3 is available, involving both configmanager and gcsrunner, so it would be a api **breaking** change.
- add `api/VERSION` to denote the next api version needed to be imported into google3


nareddyt:
@TAOXUY Quick comment: Can you make a `VERSION` file under the `api/` directory and place the text "v2" inside? This corresponds to the next config version that will be synced into google3. Since this is a breaking API change, we want to keep track of it by incrementing the version.

TAOXUY:
> @TAOXUY Quick comment: Can you make a `VERSION` file under the `api/` directory and place the text "v2" inside? This corresponds to the next config version that will be synced into google3. Since this is a breaking API change, we want to keep track of it by incrementing the version.

Hi Teju, I see in the cloudesf/copybara, you already put a `api_version` to denote the version. I think they represent something so is it possible to make it one place and make copybara rely on this `VERSION`?

nareddyt:
I need to research if it's possible to pass the version arg to copybara, but I don't think so. Worst case, the oncaller can just read the version from this file and replace the variables in copybara. I'll look into this next week.

TAOXUY:
> I need to research if it's possible to pass the version arg to copybara, but I don't think so. Worst case, the oncaller can just read the version from this file and replace the variables in copybara. I'll look into this next week.

Oncaller can also read from copybara and increment by 1, which is basically same, right?

nareddyt:
Yes, but @qiwzhang prefered to have the file in github with the actual protos. Then the oncaller doesn't need to check if there are any breaking API changes manually.

TAOXUY:
> Yes, but @qiwzhang prefered to have the file in github with the actual protos. Then the oncaller doesn't need to check if there are any breaking API changes manually.

OK, got your point. I can help update our oncall manual after.

examples/auth/envoy_config.json
```diff
@@ -1,13 +1,12 @@
 {
   "admin": {},
   "node": {
-    "id": "ESPv2",
-    "cluster": "ESPv2_cluster"
+    "cluster": "ESPv2_cluster",
```
nareddyt, line 4: How did you alphabetize the entries? Please add to the README in this directory

examples/dynamic_routing/envoy_config.json
```diff
@@ -1,341 +1,349 @@
...
+                }
+              }
+            ]
+          }
+        ],
+        "name": "ingress_listener"
+      }
+    ]
+  }
+}
```
nareddyt, line 349: Nit: Newlines at end of all files

go.sum
```diff
@@ -15,6 +15,7 @@ github.com/cncf/udpa/go v0.0.0-20200313221541-5f7e5dd04533 h1:8wZizuKuZVu5COB7Es
 github.com/cncf/udpa/go v0.0.0-20200313221541-5f7e5dd04533/go.mod h1:WmhPx2Nbnhtbo57+VJT5O0JRkEi1Wbu0z5j0R8u5Hbk=
 github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
+github.com/envoyproxy/go-control-plane v0.9.5 h1:lRJIqDD8yjV1YyPRqecMdytjDLs2fTXq363aCib5xPU=
```
nareddyt, line 18: Why was this changed if there is no change to `go.mod`?

src/go/configgenerator/listener_generator.go
```diff
@@ -30,18 +30,19 @@ import (
 	commonpb "github.com/GoogleCloudPlatform/esp-v2/src/go/proto/api/envoy/http/common"
 	pmpb "github.com/GoogleCloudPlatform/esp-v2/src/go/proto/api/envoy/http/path_matcher"
 	scpb "github.com/GoogleCloudPlatform/esp-v2/src/go/proto/api/envoy/http/service_control"
-	v2pb "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	corepb "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
-	listenerpb "github.com/envoyproxy/go-control-plane/envoy/api/v2/listener"
-	routepb "github.com/envoyproxy/go-control-plane/envoy/api/v2/route"
-	acpb "github.com/envoyproxy/go-control-plane/envoy/config/filter/accesslog/v2"
+
```
nareddyt, line 33: Nit: Remove newline, let gofmt sort them

examples/auth/envoy_config.json
TAOXUY: I think `python -m json.tool` can alphabetize entries.

PR #175, opened 2020-06-01
Additional validation and filter config cleanup for Backend Routing

nareddyt:
Config Manager changes:
- For empty prefix in APPEND: Don't generate rules, they are no-op.
- For empty prefix in CONST: Generate rules with path prefix '/'.
- For URL parser: Don't parse query params. None of the control plane config should allow query params anyways.

Filter changes:
- More validation of proto config (as described above) and tests for this.
- Tests to show duplicated query param behavior.


src/envoy/http/backend_routing/filter_config.h
```diff
@@ -57,6 +58,14 @@ class FilterConfig : public Envoy::Logger::Loggable<Envoy::Logger::Id::filter> {
         throw Envoy::ProtoValidationException(
             "Path translation for BackendRouting rule must be specified", rule);
       }
+      if (rule.path_prefix() == Envoy::EMPTY_STRING) {
```
qiwzhang, line 61: can these be specified in proto validation rules?

src/envoy/http/backend_routing/filter_config_test.cc
```diff
@@ -0,0 +1,108 @@
...
+namespace http_filters {
+namespace backend_routing {
+namespace {
+
+class BackendRoutingConfigTest : public ::testing::Test {
+ protected:
+  void createConfig(absl::string_view filter_config) {
+    google::api::envoy::http::backend_routing::FilterConfig proto_config;
+    ASSERT_TRUE(google::protobuf::TextFormat::ParseFromString(
+        std::string(filter_config), &proto_config));
```
qiwzhang, line 39: can we use Envoy::test_common,  loadYaml fuction

src/envoy/http/backend_routing/filter_config.h
qiwzhang: Beside,  coding wise, prefer   rule.path_prefix().empty()

src/go/configgenerator/listener_generator.go
```diff
@@ -795,8 +800,15 @@ func makeBackendRoutingFilter(serviceInfo *sc.ServiceInfo) (*hcmpb.HttpFilter, e
 			}
 
 			if method.BackendInfo != nil {
-				newRule.PathPrefix = method.BackendInfo.Uri
+				newRule.PathPrefix = method.BackendInfo.Path
 			}
+
+			// If the path prefix is empty, then the filter is a no-op for this operation.
+			// It doesn't need to be in the filter config, skip it.
```
qiwzhang, line 807: If it is skipped,  will we reject this request as 404?

api/VERSION
```diff
@@ -1 +1 @@
-v2
+v3
```
TAOXUY, line 1: api_version in google3 is `v1` and here is `v2`, so no need to increment before next import?

nareddyt: I worry this will be confusing to have to check the version every time. It could result in difficulties if we ever revert a CL that changes this number. 

As discussed with Wayne, let's always try to increment the number in each CL. This should be OK for API Gateway.

src/go/configgenerator/listener_generator.go
nareddyt: Skipping the backend routing filter will skip rewriting the path. But we still configure a host rewrite in the router config.

So no, it will not result in a 404.

src/envoy/http/backend_routing/filter_config_test.cc
nareddyt: We are using protobuf TextFormat across all of our filter config tests, not JSON or YAML. Let's keep it consistent, I don't see any benefit to switching to JSON. The code is already simple anyways.

src/envoy/http/backend_routing/filter_config.h
nareddyt: Good point. Done

src/envoy/http/backend_routing/filter_config_test.cc
qiwzhang: ok

src/go/configgenerator/listener_generator.go
qiwzhang: ack

api/VERSION
TAOXUY: It makes sense from our side. Just one thing, in google3 it would be like v1 -> v4 -> v7 -> v14 and can we confirm with API Gateway? @toddbeckman-g

qiwzhang: For each breaking change, we should increment this version so that github history of this file will show all the breaking changes.

toddbeckman-g: It doesn't look like this PR has any breaking API changes which affect API Gateway, so I think no increment is necessary.

To answer your question though, it doesn't really matter what the version number here is, as long as we can differentiate between them, so it is okay to skip version numbers.

TAOXUY: Thanks!

nareddyt: It's a breaking API change. Older filter configs generate no-op backend routing rules. With this change, envoy will not accept configs with no-op rules.

nareddyt: Let me clarify. Backend routing filter only re-writes the path portion of a URI. We use envoy route config to re-write the host portion.

So even for API Gateway, some operations will not need backend routing filter. If the only change we need to make is the hostname but not the path, requests can skip over this filter.

That is the breaking API change. Beforehand, those requests with no path re-write would go into the backend routing filter and be translated as APPEND requests. Now, they skip over the logic in the filter.

toddbeckman-g: Ah I see now. Then yes, I agree that this is a breaking change. Sorry for the confusion.

PR #176, opened 2020-06-02
Fix grpc-web: move grpc-web filter in front of transcoder

qiwzhang:

Current grpc-web feature is broken.  Need to move grpc-web filter before grpc transcoder filter.
If grpc transcoder is at front,  it will try to transcode a grpc-web request, it will fail.
If grpc-web filter is at front, it will convert content type from application/grpc-web to application/grpc, 
then pass to grpc transcoder, it will bypass it.

Changes:
* add some POST + /SERVICE_NAME/METHOD additional binding to grpc bookstore methods,  these methods are used in the grpc-web test.  Adding them is to re-produce the bug so that transcoder will try to transcode grpc-web request which will fail.
* move grpc-web filter in configgenerate code.


tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto
```diff
@@ -36,6 +36,15 @@ service Bookstore {
     //   curl http://DOMAIN_NAME/v1/shelves
     option (google.api.http) = {
       get: "/v1/shelves"
+      // This is added to test grpc-web, to make sure the method is enabled
+      // for transcoder filter, and it should bypass this method.
+      // grpc-web client is using POST + /service_name/method_name.
+      // It could be auto-generated by setting auto_mapping. But auto_mapping
+      // will not apply if a method already has http.rules options defined.
+      additional_bindings {
```
JLXIA, line 44: Is there any integration test using this rule, and can capture the grpc-web failure?

qiwzhang: Yes,  test/grpc_integration_test.go  its GRPCWeb test is using all these 3 additional_binding rules.  They are sending POST + /service_name/service_method HTTP request.  and pass to grpc transcoder.

qiwzhang:
We do have tests/grpc_integration_test.go its GRPCWeb test,  but it did not catch this bug due to
the method is not registered in the transcoder path_matcher,  so transcoder bypass these methods.  Then grpc-web works.  The problem is if it passes to grpc transcoder, and if it tries to transcode it , it will fail. that is what the users encountered.

So first, I need to re-produce the problem in our integration tests,  I just need to register these POST + /service_name/service_method in the grpc trancoder.  Once it is registered,  the integration test fails, so I reproduced the problem.

Then, I switched the grpc-web filter order to fix it,  and integration test passed.

nareddyt:
I see. But previously, why did the test pass when transcoder bypasses the methods?

If the transcoder bypasses the methods, then shouldn't the test fail because downstream is HTTP and upstream is gRPC? The `wantResp` shouldn't match.

nareddyt:
Nevermind, I understand now. The test framework deserializes the proto body into the JSON for a grpc web client.

TAOXUY:
> I see. But previously, why did the test pass when transcoder bypasses the methods?
> 
> If the transcoder bypasses the methods, then shouldn't the test fail because downstream is HTTP and upstream is gRPC? The `wantResp` shouldn't match.

As the PR description, grpc-web filter transcodes request to grpc already and putting in front is to skip grpc-transcoder.

qiwzhang:
grpc-web protocol is almost the same as grpc protocol, only content-type need to be converted, the body is already in the serialized proto message, when directly pass to backend grpc, it will work.
For grpc-web-text,  the serialized proto message is based64 encoded,  grpc-web filter just base64 decoded.

PR #182, opened 2020-06-03
Cleanup / validation / tests for Backend Routing

nareddyt:
- `oneofs` are required to be set.
- `jwt_audience` should be a valid header string, other envoy will crash.
- Test case for missing audience / operation.
- Cleanup empty string.

### DIFF
--- a/api/envoy/http/common/base.proto
+++ b/api/envoy/http/common/base.proto
@@ -65,7 +65,8 @@ message HttpUri {
       [(validate.rules).duration.required = true];
 }
 
-/* TODO(taoxuy): replace with envoy internal DataSource proto */
+// Duplicate DataSource from envoy for self containment.
+// https://aip.dev/213
 message DataSource {
   oneof specifier {
     option (validate.required) = true;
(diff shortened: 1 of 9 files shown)


api/envoy/http/common/base.proto
```diff
@@ -65,7 +65,8 @@ message HttpUri {
       [(validate.rules).duration.required = true];
 }
 
-/* TODO(taoxuy): replace with envoy internal DataSource proto */
+// Duplicate DataSource from envoy for self containment.
+// https://aip.dev/213
```
qiwzhang, line 69: Maybe you can try to see if we can use Envoy api DataSource directly instead of defining our own.
I know for now, CloudESF api doens't depends on any Envoy api.  Maybe we can change it.

Can be a separate as clean up tech debt

qiwzhang: Same for HttpUri,

nareddyt: @qiwzhang I disagree, it's harder to version our APIs if they're tied to envoy.

From https://aip.dev/213

> APIs **must not** define a set of API-specific common protos which live outside of its versioning structure. This prevents independent movement of particular versions ...

Given that Datasource is a versioned envoy proto, this would be a breaking change if we update from v3->v4, etc. While we can handle breaking changes, it just seems like more maintenance work for us.

qiwzhang: OK, good point. I am convinced. Thanks.

PR #184, opened 2020-06-04
b/157908493: add flag to control production prefix in generated headers

TAOXUY:
- Add --generated_header_prefix to control production prefix in generated headers. It will be used in
  - XXX-api-project-number
  - XXX-api-userinfo
  - XXX-api-consumer-number[later]

- Rename XXX-api-project-id to XXX-api-project-number. Please note it is a breaking change.


api/envoy/http/service_control/config.proto
```diff
@@ -126,4 +125,7 @@ message FilterConfig {
 
   // The Http uri to call service control
   api.envoy.http.common.HttpUri service_control_uri = 8;
+
+  // The prefix added to generated headers
+  string generated_header_prefix = 9 [(validate.rules).string.min_bytes = 1];
```
nareddyt, line 130: Also add the http header value validation that other ones have. Make sure to include `strict: false`.

src/api_proxy/service_control/request_builder.cc
```diff
@@ -596,7 +596,7 @@ Status set_backend_protocol(const SupportedLabel& l,
 Status set_consumer_project(const SupportedLabel& l,
                             const ReportRequestInfo& info,
                             Map<std::string, std::string>* labels) {
-  (*labels)[l.name] = info.check_response_info.consumer_project_id;
+  (*labels)[l.name] = info.check_response_info.consumer_proejct_number;
```
nareddyt, line 599: typo `project`

src/go/commonflags/flags.go
```diff
@@ -34,6 +34,7 @@ var (
 	HttpRequestTimeoutS        = flag.Int("http_request_timeout_s", 5, `Set the timeout in second for all requests. Must be > 0 and the default is 5 seconds if not set.`)
 	Node                       = flag.String("node", "ESPv2", "envoy node id")
 	NonGCP                     = flag.Bool("non_gcp", false, `By default, the proxy tries to talk to GCP metadata server to get VM location in the first few requests. Setting this flag to true to skip this step`)
+	GeneratedHeaderPrefix      = flag.String("generated_header_prefix", "X-Endpoint-", "Set the header prefix for the generated headers. By default, it is `X-Endpoint-``")
```
nareddyt, line 37: 2 tildes at the end of the string, should be 1

src/go/configgenerator/listener_generator.go
```diff
@@ -412,7 +412,7 @@ func makeJwtAuthnFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
 			},
 			FromHeaders:          fromHeaders,
 			FromParams:           fromParams,
-			ForwardPayloadHeader: "X-Endpoint-API-UserInfo",
+			ForwardPayloadHeader: serviceInfo.Options.GeneratedHeaderPrefix + "API-UserInfo",
```
nareddyt, line 415: Nit: Move this suffic into a const in utils or at the top of the file?

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -56,6 +55,9 @@ ServiceControlHandlerImpl::ServiceControlHandlerImpl(
       stream_info_(stream_info),
       time_source_(time_source),
       uuid_(uuid),
+      consumer_project_number_header_(
```
nareddyt, line 58: This is a breaking API change. Please increment api version

TAOXUY:
PTAL

api/envoy/http/service_control/config.proto
TAOXUY: Hi Teju, what is difference between `strict: true/false`? I haven't found doc about this.

nareddyt:
```
// x must conform to a well known regex for HTTP header names (via RFC 7230)
string x = 1 [(validate.rules).string.well_known_regex = HTTP_HEADER_NAME]

// x must conform to a well known regex for HTTP header values (via RFC 7230) 
string x = 1 [(validate.rules).string.well_known_regex = HTTP_HEADER_VALUE];

// x must conform to a well known regex for headers, disallowing \r\n\0 characters.
string x = 1 [(validate.rules).string {well_known_regex: HTTP_HEADER_VALUE, strict: false}];
```

nareddyt: It essentially only validates it doesn't contain `\r\n\0` characters, because these will crash envoy at runtime.

nareddyt: If strict is true, then it will disallow all characters that are not in RFC 7230. But envoy can handle some of these characters anyways

PR #186, opened 2020-06-05
b/158192303: #1 fix data race in config_manager_test.go

TAOXUY:
Unit tests of configmanagers failed with data racing on the [global data](https://github.com/GoogleCloudPlatform/esp-v2/blob/f08675f8efcc78190eabfbe000ed557d99b26c7d/src/go/configmanager/config_manager_test.go#L1809) and they are read by mockServer to serve requests and written by test in the same time.

So, make the raced data protected by mutex.
    
Details:
- add `safeData` to guard data with mutex
- cleanup until functions
- put literal testData into `testdata` folder

### DIFF
--- a/src/go/configmanager/config_manager_test.go
+++ b/src/go/configmanager/config_manager_test.go
@@ -15,13 +15,13 @@ package configmanager
 
 import (
 	"context"
-	"encoding/base64"
 	"flag"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"sort"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -43,23 +43,9 @@ import (
 	smpb "google.golang.org/genproto/googleapis/api/servicemanagement/v1"
 )
 
-const (
-	testProjectName  = "bookstore.endpoints.project123.cloud.goog"
-	testEndpointName = "endpoints.examples.bookstore.Bookstore"
-	testConfigID     = "2017-05-01r0"
-	testProjectID    = "project123"
-	fakeToken        = `{"access_token": "ya29.new", "expires_in":3599, "token_type":"Bearer"}`
-)
-
-var (
-	fakeConfig             []byte
-	fakeScReport           []byte
-	fakeRollouts           []byte
-	fakeProtoDescriptor    = base64.StdEncoding.EncodeToString([]byte("rawDescriptor"))
-	testBackendClusterName = fmt.Sprintf("%s_local", testProjectName)
-)
-
 func TestFetchListeners(t *testing.T) {
+	var fakeConfig, fakeScReport, fakeRollouts safeData
+
 	testData := []struct {
 		desc              string
 		enableTracing     bool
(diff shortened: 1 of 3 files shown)


TAOXUY:
> I'm not convinced about this fix. You're adding a mutex in the tests so the tests pass, hiding the race condition.
> 
> But shouldn't the mutex be added in the config manager library? Doesn't this mean there is a race in config manager?
> 
> I guess the more specific question is: Why does the data race only happen in the tests and not in config manager?

Updated the description. If you open the stacktrace in the bug, you will see the data race happens on these global data.

src/go/configmanager/config_manager_test.go
```diff
@@ -1800,37 +466,36 @@ func runTest(t *testing.T, opts options.ConfigGeneratorOptions, f func(*testEnv)
...
-func initMockRolloutServer(t *testing.T) *httptest.Server {
-	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
-		_, err := w.Write(fakeRollouts)
-		if err != nil {
-			t.Fatal("fail to write rollout config: ", err)
-		}
-	}))
+func (s *safeData) write(src []byte) {
+	s.mutex.Lock()
```
qiwzhang, line 479: I think this will not fix the problem.  For each test, its always write() and read(), they are always sequential. 
The key problem is: multiple tests share the same `fakeConfig`,   each has different config data, and they are running in parallel so their data over-write each over.
The right approach is that each test should use its own object to store its config data for initMockServer.  We don't need to protect it,  as I see it,  we always create an object first (write) and use it (read) later.  As long as each test is using its own storage, they will not conflict with each other.

TAOXUY: I don't think so. If you open the sponge of test history, 
- only `TestFixedModeDynamicRouting` are failed and other two tests are passed
- the log only a sequential execution 
so it is not because of interference between tests. 

This is because the goroutine created by http server to handle request is reading the data and main goroutine is writing the data. This is an [example](https://stackoverflow.com/questions/31792810/simple-race-condition-in-go-http-handler-is-this-really-a-race-condition).

qiwzhang: I though for each test,  the main routine write the data, and pass it to httpServer, its handler only read the data.  read data can be in multiple threads.   which main routine is writing again? if not from different tests.

TAOXUY: That is exactly the case.

qiwzhang: in that case, let us not use a global variable for all test cases,  define one variable for each test case,  it should solve the problem.  not need to have mutex.

TAOXUY: Here, we are testing configmanager under `managed` mode so the config will be read by mockServer and modified by test in the same time.

qiwzhang: I see.  for multiple iterations.

nareddyt: Thanks for the clarifications, I understand now

PR #187, opened 2020-06-05
b/157610728: update googleapis to 06042020

TAOXUY:
Please those removed checkErrorCode are [hidden](https://github.com/googleapis/googleapis/commit/1db88f2a42f6b0911c26eaf83e77377252cf94cc) according to the visibility Service Control team set.


src/api_proxy/service_control/request_builder.cc
```diff
@@ -1383,10 +1383,6 @@ Status RequestBuilder::ConvertCheckResponse(
     case CheckError::RESOURCE_EXHAUSTED:
       check_response_info->error_type = ScResponseErrorType::CONSUMER_QUOTA;
       return Status(Code::RESOURCE_EXHAUSTED, "Quota check failed.");
-    case CheckError::ABUSER_DETECTED:
```
nareddyt, line 1386: Just double checking, there were no new codes added in the proto?

TAOXUY: Checked, nop.

https://github.com/googleapis/googleapis/commit/1db88f2a42f6b0911c26eaf83e77377252cf94cc

PR #188, opened 2020-06-05
b/152334472: Change filter names to be outside the envoy namespace

nareddyt:
Custom filters shouldn't say envoy in them. Instead, use the reverse domain name of the project (com.google.espv2).

Filter names:
- Previous: `envoy.filters.http.path_matcher`
- Now: `com.google.espv2.filters.http.path_matcher`

Proto packages:
- Previous: `::google::api::envoy::http::path_matcher`
- Now: `::espv2::api::envoy::http::path_matcher`


qiwzhang:
Just FYI, filter name is not used any more,  Now it is using "typed_url" to register its factory.

qiwzhang:
since package name is not prefixed with `com.google`,  maybe filter name can start with "espv2" too?

nareddyt:
> since package name is not prefixed with `com.google`, maybe filter name can start with "espv2" too?

I think the current name is OK, it doesn't have to match the proto namespace. Harvey prefers to have full reverse dns name:

> [...] The API documentation encourages non-Envoy core projects to adopt reverse DNS to avoid clashing with Envoy core.

PR #189, opened 2020-06-05
[WIP]: Send error message in JSON foramt in response body

qiwzhang:

This PR config envoy HCM(http connection manager) LocalReplyConfig to send the error message for requests rejected by Envoy in Json format in response body

### DIFF
--- a/src/go/configgenerator/listener_generator.go
+++ b/src/go/configgenerator/listener_generator.go
@@ -43,6 +43,7 @@ import (
 	routerpb "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/router/v3"
 	hcmpb "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
 	durationpb "github.com/golang/protobuf/ptypes/duration"
+	structpb "github.com/golang/protobuf/ptypes/struct"
 	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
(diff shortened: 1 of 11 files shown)


src/go/configgenerator/listener_generator.go
```diff
@@ -237,6 +238,29 @@ func makeHttpConMgr(opts *options.ConfigGeneratorOptions, route *routepb.RouteCo
 		},
 		UseRemoteAddress:  &wrapperspb.BoolValue{Value: opts.EnvoyUseRemoteAddress},
 		XffNumTrustedHops: uint32(opts.EnvoyXffNumTrustedHops),
+		// Converting the error message for requests rejected by Envoy to JSON format:
+		//
+		//    {
+		//       "code": "http-status-code",
```
nareddyt, line 244: In ESPv1, code is the canonical grpc status code, not a http status code. Unfortunately, this will be a breaking change.

I assume there is no way to easily fix this?

qiwzhang: Ack.  Yeap,  it is a breaking change.  But I don't understand why ESPv1 is sending grpc status code for http calls. I think it make more sense to send http status code.

nareddyt: I see. I guess even ESPv1 doesn't follow the same error format that ESF / google APIs do: https://cloud.google.com/apis/design/errors#http_mapping

Seems fine, but we need to document in the migration guide.

PR #193, opened 2020-06-11
SC Config Path - Security review #7 fixes

nareddyt:
Various minor changes based on comments:
- `min_stream_report_interval_ms` should have a realistic minimum value enforced. Do this in our `ConfigParser` instead of proto config because we allow 0 value as default.
- `generated_header_prefix` is a header name, not value. Stricter validation here.
- `metric_cost_vector` should be stored as reference not array. It will never be null. Testing will be easier (test the values match, not the pointers).
- Rename `find_requirement` to match naming in the file. Seems we have some conflicting naming across filters, we can fix later.
- Clarify filter stats.
- Exception invalid access token case (already validated by protoc-gen-validate, so it should never be reached).

### DIFF
--- a/api/envoy/http/service_control/config.proto
+++ b/api/envoy/http/service_control/config.proto
@@ -80,7 +80,7 @@ message Service {
 
   // Minimum amount of time (milliseconds) between sending intermediate
   // reports on a stream.
-  uint64 min_stream_report_interval_ms = 8 [(validate.rules).uint64.gte = 100];
+  uint64 min_stream_report_interval_ms = 8;
 
   // The array of jwt payloads demanded to be logged
   repeated string log_jwt_payloads = 9;
--- a/api/envoy/http/service_control/config.proto
+++ b/api/envoy/http/service_control/config.proto
@@ -80,7 +80,7 @@ message Service {
 
   // Minimum amount of time (milliseconds) between sending intermediate
   // reports on a stream.
-  uint64 min_stream_report_interval_ms = 8;
+  uint64 min_stream_report_interval_ms = 8 [(validate.rules).uint64.gte = 100];
 
   // The array of jwt payloads demanded to be logged
   repeated string log_jwt_payloads = 9;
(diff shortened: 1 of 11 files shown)


api/envoy/http/service_control/config.proto
```diff
@@ -131,8 +131,7 @@ message FilterConfig {
 
   // The prefix added to generated headers
   string generated_header_prefix = 9 [(validate.rules).string = {
-    well_known_regex: HTTP_HEADER_VALUE,
-    strict: false,
+    well_known_regex: HTTP_HEADER_NAME,
```
TAOXUY, line 134: Will header prefix `X-Endpoint-` be regarded as a valid header name?

nareddyt: Yes. Our integration tests still pass, so it's definitely valid.

src/api_proxy/service_control/request_info.h
```diff
@@ -122,7 +122,10 @@ struct CheckResponseInfo {
 struct QuotaRequestInfo : public OperationInfo {
   std::string method_name;
 
-  const std::vector<std::pair<std::string, int>>* metric_cost_vector;
+  const std::vector<std::pair<std::string, int>>& metric_cost_vector;
+
+  QuotaRequestInfo(const std::vector<std::pair<std::string, int>>& metric_costs)
```
qiwzhang, line 127: If you use &&, then you can use std::move() to swap it.

src/envoy/http/service_control/config_parser.h
```diff
@@ -98,7 +112,8 @@ class FilterConfigParser {
       const {
     return config_;
   }
-  const RequirementContext* FindRequirement(absl::string_view operation) const {
+  const RequirementContext* find_requirement(
```
qiwzhang, line 115: I want c++ code under "envoy" to follow envoy style. which is findRequirement

nareddyt: Let me do this in another PR across the entire directory. 

This change makes it consistent with the rest of this file at least.

qiwzhang: OK

PR #198, opened 2020-06-16
b/148454048: Increase IMDS access token timeout to 30s

nareddyt:
- Modify envoy token subscriber to read the `timeout` from filter config. Currently the field is ignored, which is misleading.
- Modify config generator to always read from `--http_request_timeout_s` when generating timeouts for HTTP URIs in the filter config.
- Modify the default value of `--http_request_timeout_s` to `30s`.
- Add integration test for `--http_request_timeout_s` flag calling IAM. This follows the same code/config path as IMDS.


src/go/metadata/metadata_fetcher_test.go
```diff
@@ -316,19 +315,3 @@ func TestFetchGCPAttributes(t *testing.T) {
 	}
 
 }
-
-func TestMetadataFetcherTimeout(t *testing.T) {
```
nareddyt, line 320: Removed this because it's hard to test, we're using a hardcoded 30s. No option to change it to a lower value. Don't want unit test running for 30s.

qiwzhang: Maybe add a variable ImdsDeadLine to be used by metadata fetching, and test code change change it.

qiwzhang: I was thinking about if we should add a flag for users to adjust it.  But I don't see the drawback of setting a big timeout.  So,  default to a 30s timeout should be ok,  not need to add a new flag to adjust it.

src/go/metadata/metadata_fetcher.go
```diff
@@ -61,7 +61,7 @@ var (
 	NewMetadataFetcher = func(opts options.CommonOptions) *MetadataFetcher {
 		return &MetadataFetcher{
 			client: http.Client{
-				Timeout: opts.HttpRequestTimeout,
+				Timeout: util.DefaultImdsDeadline,
```
qiwzhang, line 64: should this be controlled by flag --http_request_timeout_s

src/go/configgenerator/listener_generator.go
```diff
@@ -767,7 +767,7 @@ func makeBackendAuthFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
 			ImdsToken: &commonpb.HttpUri{
 				Uri:     fmt.Sprintf("%s%s", serviceInfo.Options.MetadataURL, util.IdentityTokenSuffix),
 				Cluster: util.MetadataServerClusterName,
-				Timeout: ptypes.DurationProto(serviceInfo.Options.HttpRequestTimeout),
+				Timeout: ptypes.DurationProto(util.DefaultImdsDeadline),
```
qiwzhang, line 770: how about keep the original code to use the flag --http_request_timeout_s

nareddyt: Done. After discussing offline, we decided to just use the flag `--http_request_timeout_s` for all requests, including IMDS. We changed the default to `30s`.

examples/dynamic_routing/envoy_config.json
```diff
@@ -306,7 +306,7 @@
                             "route": {
                               "cluster": "http-bookstore-edf123456-uc.a.run.app:443",
                               "hostRewriteLiteral": "http-bookstore-edf123456-uc.a.run.app",
-                              "timeout": "30s"
```
nareddyt, line 309: These are timeouts to the backend. I changed the values to make them unique, they're unrelated to the `--http_request_timeout_s`.

src/envoy/token/mocks.h
```diff
@@ -28,13 +28,16 @@ class MockTokenSubscriberFactory : public TokenSubscriberFactory {
...
+               const std::string& token_url,
+               const std::chrono::seconds fetch_timeout,
+               UpdateTokenCallback callback),
               (const));
 
   MOCK_METHOD(
       TokenSubscriberPtr, createIamTokenSubscriber,
       (const TokenType& token_type, const std::string& token_cluster,
-       const std::string& token_url, UpdateTokenCallback callback,
+       const std::string& token_url, const std::chrono::seconds fetch_timeout,
```
qiwzhang, line 39: nit:  const std::chrono::seconds&    
add &,  otherwise, it doesn't make sense

qiwzhang: If it is a small struct,  like string_view,   it is better to copy the value so remove "const &"
I believe std::chrono::seconds is a very small struct.

nareddyt: I took a look. I'd prefer to do the change to `absl::string_view` in a different PR. A lot of the files in /src/envoy/token/ use `const std::string&`, and the diff is too big already.

My original code for `std::chrono::seconds` seems correct. 
- Pass `std::chrono::seconds` by value since it just wraps a single variable. No need for reference. 
- Still keep it as `const` because it's not immutable (unlike `string_view`). It has a `++` / `--` operator that we should guard against.

qiwzhang: Sorry,  my requirement is just to remove "const"  for std::chrono::seconds.   I used string_view just as an example.  

Here is my understanding
foo(const T t)  doesn't make sense since t is copied into the function.  It is exactly the same as foo(T t). 

const T& t makes sense, it is referencing

nareddyt: I see, done.

src/envoy/http/backend_auth/config_parser_impl.cc
```diff
@@ -58,7 +58,7 @@ AudienceContext::AudienceContext(
       const std::string& uri = filter_config.iam_token().iam_uri().uri();
       const std::string& cluster =
           filter_config.iam_token().iam_uri().cluster();
-      const std::chrono::seconds fetch_timeout(TimeUtil::DurationToSeconds(
+      std::chrono::seconds fetch_timeout(TimeUtil::DurationToSeconds(
```
qiwzhang, line 61: we should not remove this "const" for local variable.  I only want you to remove "const" for function argument.

nareddyt: Oops, I did a find-replace and didn't notice this in the diff. Fixed.

src/envoy/token/iam_token_info.cc
```diff
@@ -141,7 +141,7 @@ bool IamTokenInfo::parseAccessToken(absl::string_view response,
     return false;
   }
 
-  const std::chrono::seconds& expires_in = std::chrono::seconds(
+  std::chrono::seconds expires_in = std::chrono::seconds(
```
qiwzhang, line 144: we can add "const" for this one.

src/envoy/token/imds_token_info.cc
```diff
@@ -88,8 +88,7 @@ bool ImdsTokenInfo::parseAccessToken(absl::string_view response,
     return false;
   }
 
-  const std::chrono::seconds& expires_in =
-      std::chrono::seconds(expires_seconds);
+  std::chrono::seconds expires_in = std::chrono::seconds(expires_seconds);
```
qiwzhang, line 91: we can add "const" for this one too

PR #200, opened 2020-06-17
b/157610728: add consumer-type and consumer-number headers

TAOXUY:
- Add header of Prefix-Consumer-Type and header of Prefix-Consumer-Number.

- Header Prefix-Project-Number is removed since it's been replaced by Consumer-Number. Though it is a breaking change, we think no many users are using this so it should be fine.


src/api_proxy/service_control/request_builder.cc
```diff
@@ -1364,6 +1366,19 @@ Status RequestBuilder::ConvertCheckResponse(
...
   }
 
+  if (check_response.check_info().consumer_info().consumer_number() > 0) {
+    check_response_info->consumer_number = std::to_string(
+        check_response.check_info().consumer_info().consumer_number());
+  }
+
+  if (check_response.check_info().consumer_info().type() !=
+      CheckResponse_ConsumerInfo_ConsumerType::
+          CheckResponse_ConsumerInfo_ConsumerType_CONSUMER_TYPE_UNSPECIFIED) {
```
nareddyt, line 1376: Would prefer `CheckResponse::ConsumerInfo::ConsumerTypeCase::CONSUMER_TYPE_UNSPECIFIED`. Then you don't need the extra using statement either.

src/api_proxy/service_control/request_builder.cc
```diff
@@ -1364,6 +1366,19 @@ Status RequestBuilder::ConvertCheckResponse(
...
+    check_response_info->consumer_type =
+        CheckResponse_ConsumerInfo_ConsumerType_Name(
```
qiwzhang, line 1378: Is this from proto generated code?   is this a exposed public functions or just internal implementation.  if latter,  proto team can remove it anytime, we will break.

src/api_proxy/service_control/request_info.h
```diff
@@ -117,6 +117,10 @@ struct CheckResponseInfo {
   ScResponseErrorType error_type = ScResponseErrorType::ERROR_TYPE_UNSPECIFIED;
 
   std::string consumer_project_number;
```
nareddyt, line 119: Is this still needed?

TAOXUY: Yes, we have a label still using consumer_project_number
https://github.com/GoogleCloudPlatform/esp-v2/blob/585f37109ee2cf831c97f279f0a3755572b32bb8/src/api_proxy/service_control/request_builder.cc#L595

src/api_proxy/service_control/request_builder.cc
TAOXUY: It is generated. 

https://github.com/googleapis/googleapis/blob/8c53b2cb792234354c13336ac7daee61333deade/google/api/servicecontrol/v1/service_controller.proto#L122

TAOXUY: Follow the convention of nestedProtoDefinition + _name

src/api_proxy/service_control/request_builder.cc
TAOXUY: That is so long.
Also consider both the enum and the util function converting enum to string need that namespace, so it is fine to add `using`.

PR #201, opened 2020-06-18
b/158192303: #2 fix data race in rollout_id_change_fetcher_test.go

TAOXUY:
The root cause is the `cnt` is read and written in main goroutine and timer callback goroutine. Fix it by extending the sleep time.

Verify it in cl/316995062 by running the TSan Test 4k times.

### DIFF
--- a/src/go/serviceconfig/rollout_id_change_fetcher_test.go
+++ b/src/go/serviceconfig/rollout_id_change_fetcher_test.go
@@ -17,6 +17,7 @@ package serviceconfig
 import (
 	"fmt"
 	"net/http"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


src/go/serviceconfig/rollout_id_change_fetcher_test.go
```diff
@@ -65,22 +66,29 @@ func TestRolloutIdChangeFetcherSetDetectRolloutIdChangeTimer(t *testing.T) {
 	accessToken := func() (string, time.Duration, error) { return "token", time.Duration(60), nil }
 	cif := NewRolloutIdChangeDetector(&http.Client{}, serviceControlServer.GetURL(), "service-name", accessToken)
 
-	cnt := 0
-	wantCnt := 3
+	// `unsafeCnt` will be accessed in muilti-goroutine so please access it using atomic
+	// library.
+	var unsafeCnt, wantCnt uint32
```
qiwzhang, line 71: I am not so sure about this fix.  My understanding is we don't have race

main thread:   cnt = 0,   
timer thread:  t1:  cnt++,   and read,  t2:  cnt++,  and read,   t3:  cnt++,  and read.   and  cnt is not changed since rollout_id is not changed.
main thread:  cnt is read (if it waits long enough for timer thread to pass t3) 

so the sleep at line 88 is important, it should be long enough to pass t3.

I don't see the race.

TAOXUY: You are totally right. Updated.

PR #203, opened 2020-06-18
Handle unknown enums properly

nareddyt:
[Proto documentation](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#enum) explains that unknown enum values will be preserved. Use `protoc-gen-validate` to force only valid proto configs.

Ref: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23098

### DIFF
--- a/api/envoy/http/backend_routing/config.proto
+++ b/api/envoy/http/backend_routing/config.proto
@@ -34,7 +34,7 @@ message BackendRoutingRule {
   }
 
   PathTranslation path_translation = 4
-      [(validate.rules).enum = { not_in: [ 0 ] }];
+      [(validate.rules).enum = { in: [ 1, 2 ] }];
 
   // Prefix used when re-writing the path for the backend.
   string path_prefix = 3 [(validate.rules).string = {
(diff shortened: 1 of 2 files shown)


api/envoy/http/backend_routing/config.proto
```diff
@@ -34,7 +34,7 @@ message BackendRoutingRule {
   }
 
   PathTranslation path_translation = 4
-      [(validate.rules).enum = { not_in: [ 0 ] }];
+      [(validate.rules).enum = { in: [ 1, 2 ] }];
```
qiwzhang, line 37: not_in 0 is the same as in [1,2].   I don't see the different.  Why do we need this change?

nareddyt: The warning here does a good job explaining it: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#enum

> Proto 3 enums are open enum types with possible values outside the range of specified symbols.

`not_in 0` will allow any other number, including numbers outside the range [1, 2].

In the corpus example I added, you can see `path_translation: 1197`. Our filter factory currently accepts this.

qiwzhang: I see.  Thanks

PR #204, opened 2020-06-18
Update envoy to 6/18

TAOXUY:
Include the change of fixing jsonError in grpc-message

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,4 +19,5 @@ require (
 	google.golang.org/api v0.7.0
 	google.golang.org/genproto v0.0.0-20200617032506-f1bdc9086088
 	google.golang.org/grpc v1.27.0
+	google.golang.org/protobuf v1.24.0
 )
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -19,4 +19,5 @@ require (
 	google.golang.org/api v0.7.0
 	google.golang.org/genproto v0.0.0-20200617032506-f1bdc9086088
 	google.golang.org/grpc v1.27.0
+	google.golang.org/protobuf v1.24.0
```
nareddyt, line 22: Where did this come from? Shouldn't `go.sum` also be changed?

TAOXUY: It should be introduced by latest envoy change.

This module has indirectly introduced by other module(genproto) and it is already show in go.sum
https://github.com/GoogleCloudPlatform/esp-v2/blob/2cf1126f62ba9ca2e2d0faf7004c8fd1bb8ce0a9/go.sum#L174

PR #206, opened 2020-06-19
b/155218831: Send error message as JSON format in response

TAOXUY:


tests/e2e/client/apiproxy_bookstore_test.py
```diff
@@ -54,20 +55,25 @@ def _send_request(self, path, api_key=None,
...
                 r = self._call_http(path, api_key, None, data, method)
                 self.assertEqual(r.status_code, 401)
-                self.assertEqual(
-                    r.text,
-                    'Jwt is missing')
+                try:
+                    rpcStatus = json.loads(r.text)
+                    self.assertEqual(rpcStatus, json.loads('{"message":"Jwt is missing","code":401}'))
+                except:
+                    print("fail to decode json payload")
```
qiwzhang, line 62: should we call self.assertXXX to mark the test is failed.  Just printing some error will not mark the test as failured. 
Same for line 76

tests/integration_test/grpc_integration_test.go
```diff
@@ -370,24 +370,24 @@ func TestGRPCJwt(t *testing.T) {
 			clientProtocol:   "grpc",
 			method:           "ListShelves",
 			token:            testdata.FakeCloudToken,
-			wantError:        "code = PermissionDenied desc = Audiences in Jwt are not allowed",
-			wantGRPCWebError: "403 Forbidden, Audiences in Jwt are not allowed",
+			wantError:        `code = PermissionDenied, message = {"code":403,"message":"Audiences in Jwt are not allowed"}`,
+			wantGRPCWebError: `403 Forbidden, {"code":403,"message":"Audiences in Jwt are not allowed"}`,
```
qiwzhang, line 374: Is this error from gprc-status response header.   If no, could you add some negative tests to verify error message from grpc-message and grpc-status

TAOXUY: Yes, `code = ` is from grpc-status and `message=` from grpc-message.  I translate the plain error in golang grpc to rpc.Status in bookstore client,

Makefile
```diff
@@ -102,8 +102,11 @@ build-grpc-interop:
 	@cp -f bazel-bin/external/com_github_grpc_grpc/test/cpp/interop/interop_server bin/
 	@cp -f bazel-bin/external/com_github_grpc_grpc/test/cpp/interop/stress_test bin/
 
-# This target is to upload e2e client binaries so once they get updated,
-# this target should be executed.
+# This target is to upload e2e client binaries so once they get updated.
+# The execution environment of prow image may be imcompatiable with your local
+# builidng environemnt which fails the client binaries so you may have to make
+# this target in the prow image, for which you can find the latest version under
+# https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
```
qiwzhang, line 109: hmm,  is this statement correct?  I thought we are upload the client binaries into GCS bucket.  I assume our test scripts download it from the GCS bucket.

TAOXUY: Our build job currently doesn't build and uploads client binaries, which is fine since it is not modified often(I guess it hasn't for half year).

qiwzhang: So the comment should be
# E2e tests are using e2e-client-binaries from GCS bucket, if they are updated, make this target
# to upload the binaries to GCS bucket

TAOXUY: Actually, I am trying to emphasize the library incompatility of OS.

If you want, I can also add this piece.

TAOXUY: Done.

tests/e2e/scripts/prow-utilities.sh
```diff
@@ -229,10 +229,11 @@ function create_service() {
   case "$#" in
     '1')
       local swagger_json="${1}"
-      retry -n 3 run ${GCLOUD} endpoints services deploy "${swagger_json}"
+      # b/158767160: workaround servicemanagement service deploy issue.
```
nareddyt, line 232: We don't need this now

tests/endpoints/bookstore_grpc/client/client.go
```diff
@@ -28,16 +28,18 @@ import (
 	"path"
 	"strings"
 
+	bsgrpcv1 "github.com/GoogleCloudPlatform/esp-v2/tests/endpoints/bookstore_grpc/proto/v1"
```
nareddyt, line 31: Nit: Jilin prefers putting named imports under unnamed imports

PR #210, opened 2020-06-24
Support api config versioning:  add v6 to api folder name and package name

qiwzhang:
This is to support api_config versioning.  

* update api/VERSION to v6
* add api_version (v6) to api folder name as `api/envoy/v6/http/*`
* add api_version (v6) to proto package name as `api.envoy.v6.http`
* changed go and c++ code to use the new package names and folder names.


api/VERSION
```diff
@@ -1 +1 @@
-v5
+v6
```
TAOXUY, line 1: Add a README under api/ about the instructions of adding api-breaking change?

TAOXUY, line 1: Add a Makefile target to regex replace to replace? I see several patterns which we can do.

qiwzhang: Good idea.  we can start with a script to do the replacement when changing version.  I will do that with a separate PR.

qiwzhang: Will do this as a separate pr with the script to help version change.

nareddyt: +1. Some questions that should be clarified:

- When we move from v6 to v7, do we keep the v6 folder and make a new v7 folder? Or completely move v6 to v7?
- Will there be some script to do the replacements?

PR #214, opened 2020-06-30
[DO NOT SUBMIT] Test client cache destruct mode use-after-free

nareddyt:


nareddyt:
Asan failure fixed by #216

PR #216, opened 2020-07-01
Create FilterStats in client_cache

qiwzhang:

asan test found a memory use-after-free case: the log is [here](https://storage.googleapis.com/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/214/ESPv2-presubmit-asan/1278081084902346755/build-log.txt).  

Issue:
* FilterStat is owned by FilterConfig, but is used by client_cache.  
* FilterConfig is freed when a new Listener comes, but client_cache is stored in TLS(thread local slot) for each worker thread.
* client_cache may have some async operations, such as timer flushed out the cached items,  or cache destructor flushed out the cached items,  these items need to send to remote server, and its callback will update the filter_stats.  But filter_stats may be freed.

Changes:
* create filter_stats on client_cache.  This should be fine, as long as they are using the same stats_prefix and stat_scope,  they will be using the same stats.

### DIFF
--- a/src/envoy/http/service_control/service_control_call_impl.cc
+++ b/src/envoy/http/service_control/service_control_call_impl.cc
@@ -130,20 +130,21 @@ void ServiceControlCallImpl::createIamTokenSub() {
 
 ServiceControlCallImpl::ServiceControlCallImpl(
     FilterConfigProtoSharedPtr proto_config, const Service& config,
-    ServiceControlFilterStats& filter_stats,
+    const std::string& stats_prefix,
     Envoy::Server::Configuration::FactoryContext& context)
     : filter_config_(*proto_config),
       token_subscriber_factory_(context),
       tls_(context.threadLocal().allocateSlot()) {
   // Pass shared_ptr of proto_config to the function capture so that
   // it will not be released when the function is called.
-  tls_->set([proto_config, &config, &filter_stats,
+  tls_->set([proto_config, &config, stats_prefix, &scope = context.scope(),
              &cm = context.clusterManager(),
              &time_source =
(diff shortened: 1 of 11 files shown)


src/envoy/http/service_control/service_control_call_impl.cc
```diff
@@ -130,20 +130,21 @@ void ServiceControlCallImpl::createIamTokenSub() {
...
-    ServiceControlFilterStats& filter_stats,
+    const std::string& stats_prefix,
     Envoy::Server::Configuration::FactoryContext& context)
     : filter_config_(*proto_config),
       token_subscriber_factory_(context),
       tls_(context.threadLocal().allocateSlot()) {
   // Pass shared_ptr of proto_config to the function capture so that
   // it will not be released when the function is called.
-  tls_->set([proto_config, &config, &filter_stats,
+  tls_->set([proto_config, &config, stats_prefix, &scope = context.scope(),
```
nareddyt, line 140: Will `stats_prefix` be copied in the lambda capture? Or captured by reference because the original variable is already a reference?

qiwzhang: it is a copy, you have to specify "&" if you want to pass in as reference.

nareddyt:
```
//src/envoy/http/service_control:filter_stats_test TIMEOUT in 315.1s
```

/retest

src/envoy/http/service_control/service_control_call_impl.cc
nareddyt: Right, got confused because the variable being captured is a reference. But this will copy.

PR #217, opened 2020-07-06
Update envoy to 7/6/2020

TAOXUY:


src/envoy/http/backend_auth/filter.cc
```diff
@@ -83,11 +83,17 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
     return FilterHeadersStatus::StopIteration;
   }
 
-  if (headers.Authorization()) {
+  // Copy the existing `Authorization` header to `x-forwarded-authorization`
+  // header.
+  const Envoy::Http::HeaderEntry* existAuthToken =
```
nareddyt, line 88: Reading `Authorization` like this is not O(1) anymore. With the new O(1) header map, we need to register a handle in each filter to provide O(1) access.

Example: https://github.com/envoyproxy/envoy/blob/7d5e12cb31bce4750838fa3f02a1e4e26261c067/source/extensions/filters/http/cors/cors_filter.cc#L17

Well, I'm not sure if the previous code was O(1) either. So maybe we can clean this up in another PR. What do you think @qiwzhang?

TAOXUY: Good catch for the `origin` header.

Agree, we can clean up in another PR.

qiwzhang: Yes,  good catch @nareddyt.   I can start to work on the fix in upstream envoy if nobody has started it

qiwzhang: For our code,  we can file a separate bug to optimize our header handling.

qiwzhang: b/160645642

nareddyt: Thanks Wayne. Maybe comment on the PR to let matt klein know you are working on it.

TAOXUY:
PTAL

PR #223, opened 2020-07-11
b/152977807: rewrite unit test to use table driven format

TAOXUY:


src/go/util/marshal_test.go
```diff
@@ -43,28 +40,36 @@ func TestUnmarshalBytesToPbMessage(t *testing.T) {
...
-
-	wantResp = &scpb.ReportResponse{
-		ServiceConfigId: "test-id",
+	testCases := []struct {
+		getResp   proto.Message
+		wantResp  proto.Message
+		wantError string
+	}{
+		{
+			getResp: &confpb.Service{},
```
nareddyt, line 49: Can you add `desc` variable with string description? I still don't understand why there are fields populated in the `wantResp`. Like why is `Id` set in this test case, where did `test-id` come from?

src/go/util/marshal_test.go
```diff
@@ -43,28 +40,36 @@ func TestUnmarshalBytesToPbMessage(t *testing.T) {
...
+			wantError: "not support unmarshalling",
+		},
 	}
-	getResp = &scpb.ReportResponse{}
-	_test()
 
-	wantResp = &scpb.ReportRequest{
-		ServiceConfigId: "test-id",
+	for _, tc := range testCases {
+		_test(tc.getResp, tc.wantResp, tc.wantError)
```
nareddyt, line 73: You're passing getResp, wantResp. But func parameters are ordered wantResp, getResp

src/go/util/marshal_test.go
```diff
@@ -26,10 +26,7 @@ import (
 )
 
 func TestUnmarshalBytesToPbMessage(t *testing.T) {
-	var wantResp, getResp proto.Message
-	wantError := ""
-
-	_test := func() {
+	_test := func(wantResp, getResp proto.Message, wantError string) {
```
nareddyt, line 29: No need for func, just move this code into the test loop.

src/go/util/marshal_test.go
TAOXUY: It is just to differentiate the gotMessage is right. Rename it to be gotMessageHolder.

nareddyt: Ah i see

PR #218, opened 2020-07-08
b/158262661: Move snake to json segment mapping to per-operation instead of global

nareddyt:
In OpenAPI specs, a path template jsonName may conflict with the jsonName of well known `google.protobuf.*` types. To prevent this, define the snake_to_json segment mapping per operation instead of globally in the path matcher filter. Then the validation of name conflicts will only be per-operation, greatly making the validation less strict.

Also optimize when this needs to be configured in the path matcher config. We only pass this when we need path parameter extraction under constant address backend routing.

Diff is too large already. I will follow-up with PRs for:
- Adding integration tests.
- Incrementing api version (breaking API change).

Ref: https://groups.google.com/u/1/g/google-cloud-endpoints/c/hxyuDx1yrvQ
Closes #185

Signed-off-by: Teju Nareddy <nareddyt@google.com>

PR #225, opened 2020-07-13
Format examples with script, redeploy examples in `cloudesf-testing`

nareddyt:
- Added a script to sort example JSONs automatically when running `make format`
- Redeployed all examples in `cloudesf-testing`. That way, everyone on the team has permissions to deploy and update them.

FYI the diff looks big because this sorts the JSONs alphabetically by key. Semantically, they are identical. The only difference is the name of each API.

Unblocks #218


examples/auth/openapi_swagger.json
```diff
@@ -1,21 +1,41 @@
...
   "consumes": [
     "application/json"
   ],
-  "produces": [
-    "application/json"
-  ],
-  "schemes": [
-    "https"
-  ],
+  "definitions": {
```
nareddyt, line 6: FYI this sorts the OpenAPI files as well. It's a little weird to see the definitions come before the paths, but I don't have a better solution. I think it should be fine, otherwise it will be hard to maintain. These are small files anyways, so it should not be confusing.

examples/auth/openapi_swagger.json
```diff
@@ -74,55 +94,35 @@
...
       "x-google-issuer": "123456789-compute@developer.gserviceaccount.com",
-      "x-google-jwks_uri": "https://www.googleapis.com/service_accounts/v1/jwk/123456789-compute@developer.gserviceaccount.com",
-      "x-google-audiences": "bookstore.endpoints.cloudesf-testing.cloud.goog"
+      "x-google-jwks_uri": "https://www.googleapis.com/service_accounts/v1/jwk/123456789-compute@developer.gserviceaccount.com"
     },
     "firebase": {
       "authorizationUrl": "",
       "flow": "implicit",
       "type": "oauth2",
+      "x-google-audiences": "apiproxy-231719",
```
TAOXUY, line 116: Also update them?

nareddyt: We can, but there's no point. I just needed to update the `host` field so we can re-deploy them to `cloudesf-testing`. All the other fields can be set to anything, it doesn't matter.

examples/grpc_dynamic_routing/envoy_config.json
```diff
@@ -1040,32 +1035,14 @@
...
+                            "serviceName": "examples-grpc-dynamic-routing-wd6ufmzfya-uc.a.run.app"
                           }
                         ]
                       }
                     },
                     {
                       "name": "envoy.filters.http.grpc_web"
                     },
-                    {
-                      "name": "envoy.filters.http.grpc_json_transcoder",
```
qiwzhang, line 1052: why grpc_trancoder filter is removed

scripts/format-examples.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Fail on any error.
+set -eo pipefail
+
+# Recurse on globs
+shopt -s globstar
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+#shellcheck source=./scripts/all-utilities.sh
+. "${ROOT}/scripts/all-utilities.sh" || { echo 'Cannot load Bash utilities' && exit 1; }
```
qiwzhang, line 25: I don't see we use any code from all-utiilties.sh.

scripts/format-examples.sh
```diff
@@ -0,0 +1,34 @@
...
+
+for filename in $ROOT/examples/**/*.json; do
+    echo "Formatting $filename"
+
+    TEMP_FILE=$(mktemp)
+    jq -S '.' "$filename" > "$TEMP_FILE"
```
qiwzhang, line 31: Could you add comment on how "jq" can re-format the json?  For readers that know familir with "jq"

examples/README.md
```diff
@@ -7,7 +7,13 @@ Each folder contains three files:
...
+To generate this, deploy `openapi_swagger.json` to service management:
+
+```shell script
+gcloud endpoints services deploy ./examples/$DIRECTORY/openapi_swagger.json
+```
+
+Then run:
 
 ```shell script
 gcloud endpoints configs describe "${CONFIG_ID}" \
```
qiwzhang, line 19: We may have to use "curl" to fetch service config, just like `gcloud_build_image`. 
otherwise, the proto descriptor is missing, and grpc_transcoder is not added.

examples/grpc_dynamic_routing/envoy_config.json
nareddyt: Good catch, I skipped the large diff... Will fix.

scripts/format-examples.sh
nareddyt: Done

scripts/format-examples.sh
nareddyt: Done

examples/README.md
nareddyt: Right, updated README with this. I'll make a PR later to cleanup instructions for grpc vs openapi.

PR #227, opened 2020-07-14
b/161168963: Include Access Loggers in JSON Any resolver

toddbeckman-g:
- Add unit test for Resolver which fails for these extensions.
- Add extensions to `marshal.go` so that these pass.

While access loggers are not added by configgenerator, GCS Runner will fail to parse configs which have these extensions. In the future, additional extensions may be needed.


TAOXUY:
/override ESPv2-cloud-run-e2e-app-engine-http-bookstore

Override the unready test

toddbeckman-g:
@nareddyt There probably isn't a general solution, but we should be fine using a YAGNI approach of adding them when they make sense.

PR #229, opened 2020-07-15
b/161293641: Support ESP versions in `gcloud_build_image`

nareddyt:
Changes:
- When building the custom image, include the fully-qualified ESP version number in the tag.
- Allow the user to specify a ESP version via `-v` flag. If not specified, defaults to `2` (latest release).

Examples:
```
./gcloud_build_image -s service -c config -p project
gcr.io/project/endpoints-runtime-serverless:2.13.0-service-config
```
```
./gcloud_build_image -s service -c config -p project -v 2.11
gcr.io/project/endpoints-runtime-serverless:2.11.0-service-config
```
```
./gcloud_build_image -s service -c config -p project -i gcr.io/cloudesf-testing/apiproxy-serverless:dev-fdc79992f0db0af82dc0f5848ec196b753f4668f
gcr.io/project/endpoints-runtime-serverless:custom-service-config
```

Ref: https://groups.google.com/g/google-cloud-endpoints/c/MsQIqfwbJ74

### DIFF
--- a/docker/serverless/gcloud_build_image
+++ b/docker/serverless/gcloud_build_image
@@ -12,9 +12,13 @@
 #   gcloud endpoints services list
 #   gcloud endpoints configs list --service=${SERVICE}
 # Use the latest one for the CONFIG_ID
+#
+# The script will use the latest ESPv2 version by default.
+# Use the -v option to pass in a custom version (example: -v 2.9).
 
-# Default to the release image if not specified
-BASE_IMAGE=gcr.io/endpoints-release/endpoints-runtime-serverless:2
+# Default to the latest released ESPv2 version.
+BASE_IMAGE_NAME="gcr.io/endpoints-release/endpoints-runtime-serverless"
+ESP_TAG="2"
 
 function error_exit() {
   # ${BASH_SOURCE[1]} is the file name of the caller.
@@ -22,12 +26,16 @@ function error_exit() {
   exit ${2:-1}
 }
 
-while getopts :c:s:p:i: arg; do
+while getopts :c:s:p:v:i: arg; do
   case ${arg} in
     c) CONFIG_ID="${OPTARG}";;
     s) SERVICE="${OPTARG}";;
     p) PROJECT="${OPTARG}";;
-    i) BASE_IMAGE="${OPTARG}";;
+    v) ESP_TAG="${OPTARG}";;
+    i)
+      BASE_IMAGE="${OPTARG}"
+      ESP_FULL_VERSION="custom"
+      ;;
     \?) error_exit "Unrecognized argument -${OPTARG}";;
   esac
 done
@@ -35,7 +43,33 @@ done
 [[ -n "${PROJECT}" ]] || error_exit "Missing required PROJECT"
 [[ -n "${SERVICE}" ]] || error_exit "Missing required SERVICE"
 [[ -n "${CONFIG_ID}" ]] || error_exit "Missing required CONFIG_ID"
+
+# If user did not pass in custom image, then form the fully-qualified base image.
+if [ -z "${BASE_IMAGE}" ]; then
+  BASE_IMAGE="${BASE_IMAGE_NAME}:${ESP_TAG}"
+fi;
 echo "Using base image: ${BASE_IMAGE}"
+
+# If user did not pass in a custom image, then determine the ESP version.
+if [ -z "${ESP_FULL_VERSION}" ]; then
+  echo "Determining fully-qualified ESP version for tag: ${ESP_TAG}"
+
+  ALL_TAGS=$(gcloud container images list-tags "${BASE_IMAGE_NAME}" \
+      --filter="tags~^${ESP_TAG}$" \
+      --format="value(tags)")
+  IFS=',' read -ra TAGS_ARRAY <<< "${ALL_TAGS}"
+
+  if [ ${#TAGS_ARRAY[@]} -eq 0 ]; then
+    error_exit "Did not find ESP version: ${ESP_TAG}"
+  fi;
+
+  # Get the longest tag by creating an array indexed by tag length. Then use the last element.
+  for item in "${TAGS_ARRAY[@]}"; do
+    TAGS_BY_LENGTH[${#item}]=$item
+  done
+  ESP_FULL_VERSION="${TAGS_BY_LENGTH[-1]}"
+fi
+echo "Building image for ESP version: ${ESP_FULL_VERSION}"
 
 cd "$(mktemp -d /tmp/docker.XXXX)"
 
@@ -59,7 +93,7 @@ USER envoy
 ENTRYPOINT ["/env_start_proxy.py"]
 EOF
 
-NEW_IMAGE="gcr.io/${PROJECT}/endpoints-runtime-serverless:${SERVICE}-${CONFIG_ID}"
+NEW_IMAGE="gcr.io/${PROJECT}/endpoints-runtime-serverless:${ESP_FULL_VERSION}-${SERVICE}-${CONFIG_ID}"
 gcloud builds submit --tag "${NEW_IMAGE}" . --project="${PROJECT}"
 )
 
--- a/prow/e2e-gcloud-build-image.sh
+++ b/prow/e2e-gcloud-build-image.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# E2E test for gcloud_build_image script.
+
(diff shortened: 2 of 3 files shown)


docker/serverless/gcloud_build_image
```diff
@@ -12,31 +12,65 @@
...
 [[ -n "${CONFIG_ID}" ]] || error_exit "Missing required CONFIG_ID"
+
+# If user did not pass in custom image, then form the fully-qualified base image.
+if [ -z "${BASE_IMAGE}" ]; then
+  BASE_IMAGE="${BASE_IMAGE_NAME}:${ESP_TAG}"
+fi;
 echo "Using base image: ${BASE_IMAGE}"
 
+# If user did not pass in a custom image, then determine the ESP version.
+if [ -z "${ESP_FULL_VERSION}" ]; then
```
nareddyt, line 54: Now the logic in this script is complicated. I wonder if we need some automated tests for this. This logic is never tested in our e2e tests, because we always deploy with the `-i` flag.

@qiwzhang @TAOXUY what do you think? Not sure what the best way to test this is.

docker/serverless/gcloud_build_image
```diff
@@ -12,31 +12,65 @@
...
+  ALL_TAGS=$(gcloud container images list-tags "${BASE_IMAGE_NAME}" \
+      --filter="tags~^${ESP_TAG}$" \
+      --format="value(tags)")
+  IFS=',' read -ra TAGS_ARRAY <<< "${ALL_TAGS}"
+
+  if [ ${#TAGS_ARRAY[@]} -eq 0 ]; then
+    error_exit "Did not find ESP version: ${ESP_TAG}"
+  fi;
+
+  # Get the longest tag by creating an array indexed by tag length. Then use the last element.
```
TAOXUY, line 66: Sort array by dictionary order? Otherwise, 2.10.0 and 2.10.1 are likely be out of order according to the comment though we haven't published minor version so far.

docker/serverless/gcloud_build_image
TAOXUY: Prefer to add e2e for this piece of  code path since they are heavily used for our users.

qiwzhang: It is tricky.  Our e2e tests not use official released tag.  only use dev images with SHA.  Not easy to test it as end users 
can we re-org the script into some functions, and use a separate script to test these functions.  only test generated image name, not e2e test?

docker/serverless/gcloud_build_image
nareddyt: This would never happen. If we publish 2.10.1, then tag 2 and 2.10 will only point to 2.10.1. It would never get a string with both 2.10.0 and 2.10.1 at the same time.

docker/serverless/gcloud_build_image
nareddyt: Discussed offline with Wayne. Difficult to unit test because we still rely on GCR to provide tags.

I'll add a quick e2e test for this instead, should be fairly easy. And it will emulate how a user will use the script.

qiwzhang: SG

docker/serverless/gcloud_build_image
TAOXUY: OK, I see you query the tags of the image pointed by tag "2", which only gives 2, 2.X and 2.X.Y uniquely.

docker/serverless/gcloud_build_image
nareddyt: Updated with test.

prow/e2e-gcloud-build-image.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
+# E2E test for gcloud_build_image script.
```
TAOXUY, line 3: Copyright paragraph?

prow/e2e-gcloud-build-image.sh
```diff
@@ -0,0 +1,94 @@
...
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+PROJECT_NAME="cloudesf-testing"
+
+# gcloud config
+gcloud config set project "${PROJECT_NAME}"
+export CLOUDSDK_CORE_DISABLE_PROMPTS=1
+
+# Use a service that is not used for any other tests.
+SERVICE_NAME="hellogrpc.endpoints.cloudesf-testing.cloud.goog"
```
TAOXUY, line 16: More meaningful service name like 
`gcloud_build_image-e2e-test.xxx`?

nareddyt:
New e2e test passes and behaves as expected: https://oss-prow.knative.dev/view/gcs/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/229/ESPv2-e2e-gcloud-build-image/1284257742298550273

prow/e2e-gcloud-build-image.sh
```diff
@@ -0,0 +1,94 @@
...
+echo "=== Test 5: When no ESP version is specified, the script uses the latest ESPv2 release. ==="
+# Knowing the latest ESP version number is hard, it depends on what is tagged in GCR.
+# This is a chicken and egg problem, because `gcloud_build_image` uses that.
+# That means we don't have a reliable way of checking if the output is correct.
+# So just test the script passes, and allow the developer to manually verify the output.
+${ROOT}/docker/serverless/gcloud_build_image \
+    -s "${SERVICE_NAME}" \
+    -c "${CONFIG_ID}" \
+    -p "${PROJECT_NAME}"
+echo ">>> WARNING: For the test above, manually verify the output version of the image is expected."
```
qiwzhang, line 94: can we get the latest version from file "VERSION" that we updated for each release?

prow/e2e-gcloud-build-image.sh
nareddyt: Done

prow/e2e-gcloud-build-image.sh
nareddyt: Done

prow/e2e-gcloud-build-image.sh
nareddyt: I considered this before making the PR, but there's a few complications. I don't think it's worth trying to work around these:

1) VERSION file has the version for the next release. Current stable version is 1 below the content in the file.
2) During a release, we make a PR to increment the version before we make the release stable. In this time period, the VERSION file would be 2 ahead the current stable.
3) If we need to roll back a release (as we've done in the past), the VERSION file will be 2 (or more) ahead the current stable version.
4) If we ever release with a patch version, then it gets more complicated. Complications 1-3 are only considering minor versions, not patch.

qiwzhang: I see.
How about defining the office_release_version here in this script and  update release doc to add a step to update this version for each release and each rollback.

nareddyt: We can, but I'm still worried it will complicate the release process. 

The release engineer needs to submit the PR **as soon as** they update the docker tag in GCR. Otherwise, the versions will be out of sync, and the test will start failing.

Honestly, I do not think it's worth making the release process more complicated for this test. But it's up to you, what do you think?

qiwzhang: I am concerning about if some change is changed, the script is not getting correct release version,  how do we catch this?
For now, it depends on manually run this script,  of check this e2e test log if it is passed.   Such manual step will not catch bugs.

nareddyt: Discussed offline, decided this would complicate the release process too much.

vorobievalex:
@nareddyt, what is a possible way for a script, that uses gcloud_build_image, to refer to the container later in 'gcloud run deploy'?
(A script, that performs steps described in https://cloud.google.com/endpoints/docs/openapi/get-started-cloud-functions)
I am fixing such a script, that was previously able to calculate the name as "gcr.io/ESP_PROJECT_ID/endpoints-runtime-serverless:CLOUD_RUN_HOSTNAME-CONFIG_ID" from the variables known from the appropriate gcloud commands output.
Adding "ESP_VERSION-" to the name adds a variable, that is calculated inside gcloud_build_image and never exposed outside.
I can yet only think of temporary option of patching gcloud_build_image to expose the variable or repeating ESP_VERSION calculation outside of the script...

nareddyt:
Good question, let me open a new issue for this in case other people are searching.

PR #230, opened 2020-07-16
b/161168963: Add StatsConfig and StatsSink to Resolver

toddbeckman-g:
These are more configs that the Resolver fails to resolve, although it's odd as it's not an extension.

TESTED=Added the test cases first to reproduce the error; adding the Resolver code fixes this.

### DIFF
--- a/src/go/util/marshal_test.go
+++ b/src/go/util/marshal_test.go
@@ -40,6 +40,7 @@ func TestResolver(t *testing.T) {
 		{msg: &accessgrpcpb.TcpGrpcAccessLogConfig{}},
 		{msg: &accessgrpcpb.CommonGrpcAccessLogConfig{}},
 		{msg: &statspb.StatsSink{}},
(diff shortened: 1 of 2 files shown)


src/go/util/marshal_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 
+	statspb "github.com/envoyproxy/go-control-plane/envoy/config/metrics/v3"
```
TAOXUY, line 26: NIT: format import in dictionary order

toddbeckman-g: gofmt automatically sorts this back to the top. I think it's because it's sorting the import string alphabetically (config < extensions).

toddbeckman-g:
Hold on, there's another one to add.

PR #233, opened 2020-07-21
b/161200191: add api_key in filter_state for access logging

TAOXUY:
Implement this by adding one method "fillFilterState" in handler's interface to inject `filterState`,since streamInfo in handler shared by check and report is `const` and it is required by `log` interface.

### DIFF
--- a/src/envoy/http/service_control/filter.cc
+++ b/src/envoy/http/service_control/filter.cc
@@ -49,7 +49,7 @@ Envoy::Http::FilterHeadersStatus ServiceControlFilter::decodeHeaders(
 
   handler_ =
       factory_.createHandler(headers, decoder_callbacks_->streamInfo(), stats_);
-
+  handler_->fillFilterState(*decoder_callbacks_->streamInfo().filterState());
   state_ = Calling;
   stopped_ = false;
 
--- a/src/envoy/utils/filter_state_utils.h
+++ b/src/envoy/utils/filter_state_utils.h
@@ -26,6 +26,8 @@ constexpr char kOperation[] =
     "com.google.espv2.filters.http.path_matcher.operation";
 constexpr char kQueryParams[] =
     "com.google.espv2.filters.http.path_matcher.query_params";
+constexpr char kApiKey[] =
+    "com.google.espv2.filters.http.service_control.api_key";
 
 // Sets a read only string value in the filter state.
 void setStringFilterState(Envoy::StreamInfo::FilterState& filter_state,
--- a/src/envoy/utils/filter_state_utils.h
+++ b/src/envoy/utils/filter_state_utils.h
@@ -22,11 +22,13 @@ namespace envoy {
 namespace utils {
 
 // Data names in `FilterState` set by Path Matcher filter:
-constexpr char kOperation[] =
+constexpr char kFilterStateOperation[] =
     "com.google.espv2.filters.http.path_matcher.operation";
-constexpr char kQueryParams[] =
+constexpr char kFilterStateQueryParams[] =
     "com.google.espv2.filters.http.path_matcher.query_params";
-constexpr char kApiKey[] =
+
+// Data name in `FilterState` set by Service Control filter:
+constexpr char kFilterStateApiKey[] =
     "com.google.espv2.filters.http.service_control.api_key";
 
 // Sets a read only string value in the filter state.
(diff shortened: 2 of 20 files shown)


src/envoy/http/service_control/filter.cc
```diff
@@ -49,7 +49,7 @@ Envoy::Http::FilterHeadersStatus ServiceControlFilter::decodeHeaders(
 
   handler_ =
       factory_.createHandler(headers, decoder_callbacks_->streamInfo(), stats_);
```
qiwzhang, line 51: There is another createHandler in the ::log() function.  we need to add there too.  Or just pass filterState to createHandler()

TAOXUY: It is unlikely you can pass changeable filterState to createHandler since it is from `const` streamInfo.
We only add filterState once in decodeHeader.

src/envoy/utils/filter_state_utils.h
```diff
@@ -26,6 +26,8 @@ constexpr char kOperation[] =
     "com.google.espv2.filters.http.path_matcher.operation";
 constexpr char kQueryParams[] =
     "com.google.espv2.filters.http.path_matcher.query_params";
+constexpr char kApiKey[] =
```
qiwzhang, line 29: can we prefix these names with FilterState  or change the name space to filter_state?

tests/integration_test/access_log_test.go
```diff
@@ -76,7 +78,8 @@ func TestAccessLog(t *testing.T) {
 	makeOneRequest(t, s)
 	s.TearDown(t)
 	expectAccessLog := fmt.Sprintf("\"POST /echo?key=api-key HTTP/1.1\"200"+
-		" - 20 19\"-\" \"Go-http-client/1.1\"\"localhost:%v\" \"127.0.0.1:%v\"\n",
+		" - 20 19\"-\" \"Go-http-client/1.1\"\"localhost:%v\" \"127.0.0.1:%v\" " +
+		"\"1.echo_api_endpoints_cloudesf_testing_cloud_goog.Echo\" \"api-key\"\n",
```
qiwzhang, line 82: can we change "api-key" to "test-api-key"  or "my-api-key" for more readability

src/envoy/http/service_control/filter.cc
nareddyt: Can we just remove the `createHandler` in `log` and assume it is always set? Envoy will always guarantee to call `decodeHeaders` before `log`.

src/envoy/utils/filter_state_utils.h
nareddyt: Agreed, we should change the naming before API Gateway team depends on it. Is there an envoy convention for naming filter state?

tests/integration_test/access_log_test.go
```diff
@@ -62,7 +62,9 @@ func TestAccessLog(t *testing.T) {
 	accessLogFormat := "\"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\"" +
```
nareddyt, line 62: Update link above: https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators

src/envoy/http/service_control/filter.cc
TAOXUY: Yes, I believe the accessLog from HCM is collected earlier than that of listeners.
https://github.com/envoyproxy/envoy/blob/88dcb292817946510bb87f8f379a954962cece79/source/common/http/conn_manager_impl.cc#L618

TAOXUY: It is just protecting from nullness. No much value to remove it.

src/envoy/utils/filter_state_utils.h
```diff
@@ -22,11 +22,15 @@ namespace envoy {
 namespace utils {
 
 // Data names in `FilterState` set by Path Matcher filter:
-constexpr char kOperation[] =
+constexpr char kFilterStateOperation[] =
     "com.google.espv2.filters.http.path_matcher.operation";
```
nareddyt, line 26: Just to confirm: The values in these strings are OK? We don't need to specifically include filter_state in the value?

TAOXUY: The `FilterState` prefix in var names is to make them more specific but as for value, when they are used, it's already known they are filter states.

TAOXUY:
Pass all the test except not ESPv2-e2e-gcloud-build-image.

PR #236, opened 2020-07-22
b/161827325: add original_path in FilterState for access log

TAOXUY:


src/envoy/http/service_control/handler_impl.cc
```diff
@@ -107,9 +111,16 @@ ServiceControlHandlerImpl::ServiceControlHandlerImpl(
...
-void ServiceControlHandlerImpl::fillFilterState(FilterState& filter_state) {
+void ServiceControlHandlerImpl::fillFilterState(
+    const Envoy::Http::RequestHeaderMap& headers, FilterState& filter_state) {
   utils::setStringFilterState(filter_state, utils::kFilterStateApiKey,
                               api_key_);
+
+  auto* entry = headers.get(Headers::get().Path);
+  if (entry) {
+    utils::setStringFilterState(filter_state, utils::kFilterStateOriginalPath,
+                                std::string(entry->value().getStringView()));
```
qiwzhang, line 122: can setStringFilterState take string_view

TAOXUY:
> Why are we having service control filter add the original path to the filter state? Shouldn't this be a feature in the backend routing filter? The backend routing filter is the one that modifies the path.
> 
> Think about it this way: For ESPv2 on GKE, backend routing filter is not enabled. Then why should service control filter log the original path?

Not exactly, both grpc_transcoder filter(GKE, GCE) and backend_routing filter(Cloud Run) will modify the path even though transcoder filter will keep it in `x-envoy-original-path`, so we need to log original path before them, like service control filter.

Besides, logging it in one place is more convenient for Gateway team to query.

qiwzhang:
Teju may have a good point.  How about this;  we don't create a filterState,  we just use x-envoy-origin-path.
We will create it in our backend_auth if it is not set.   access_log can use it directly.

nareddyt:
> Teju may have a good point. How about this; we don't create a filterState, we just use x-envoy-origin-path.
> We will create it in our backend_auth if it is not set. access_log can use it directly.

Yes I like this approach. Then we only need to update `x-envoy-original-path` when the backend routing filter does a path rewrite. If it does not do any path rewrite, no need to update it.

TAOXUY:
Maybe I am wrong, one concern is that:
- `x-envoy-original-path` can be from upstream(can be faked or generated by upstream envoy)
- `x-envoy-original-path` can be overwritten in [grpc json transcoder](https://github.com/envoyproxy/envoy/blob/67649af8d4881fb7593c728f181eff393f85a638/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L389)
- `x-envoy-original-path` can be added from backend routing filter if this header doesn't exist

The `x-envoy-original-path` got by Gateway may not be the real `original path` if it's case 1.

nareddyt:
> Maybe I am wrong, one concern is that:
> 
> * `x-envoy-original-path` can be from upstream(can be maliciously faked from upstream or generated by upstream envoy)
> * `x-envoy-original-path` can be overwritten in [grpc json transcoder](https://github.com/envoyproxy/envoy/blob/67649af8d4881fb7593c728f181eff393f85a638/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L389)
> * `x-envoy-original-path` can be added from backend routing filter if this header doesn't exist
> 
> For the Gateway case, if it is the case 1 happens with case 3, the `x-envoy-original-path` got by Gateway may not be the real `original path` of the request passing through. Unless we make backend routing filter know if the transcoder filter is not there and overwrite the `x-envoy-original-path` header optionally.

Case #1, you mean malicious *downstream* client, right?

I see your point though, we can't just add a check for the header in backend auth. We have to always override it, which will lose the gRPC transcoding information.

I guess the current approach is OK. The downside is we always write the original URL, even if there is no modification.

qiwzhang:
SG then. Thanks.

TAOXUY:
> > Maybe I am wrong, one concern is that:
> > 
> > * `x-envoy-original-path` can be from upstream(can be maliciously faked from upstream or generated by upstream envoy)
> > * `x-envoy-original-path` can be overwritten in [grpc json transcoder](https://github.com/envoyproxy/envoy/blob/67649af8d4881fb7593c728f181eff393f85a638/source/extensions/filters/http/grpc_json_transcoder/json_transcoder_filter.cc#L389)
> > * `x-envoy-original-path` can be added from backend routing filter if this header doesn't exist
> > 
> > For the Gateway case, if it is the case 1 happens with case 3, the `x-envoy-original-path` got by Gateway may not be the real `original path` of the request passing through. Unless we make backend routing filter know if the transcoder filter is not there and overwrite the `x-envoy-original-path` header optionally.
> 
> Case #1, you mean malicious _downstream_ client, right?
> 
> I see your point though, we can't just add a check for the header in backend auth. We have to always override it, which will lose the gRPC transcoding information.
> 
> I guess the current approach is OK. The downside is we always write the original URL, even if there is no modification.

Admit the name of `original_path` is misleading. Actually gateway only cares about getting the `raw` path no matter if it is modified or not later.

Change the filter state name to `com.google.espv2.filters.http.service_control.path`

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -107,9 +121,16 @@ ServiceControlHandlerImpl::ServiceControlHandlerImpl(
...
 
-void ServiceControlHandlerImpl::fillFilterState(FilterState& filter_state) {
+void ServiceControlHandlerImpl::fillFilterState(const RequestHeaderMap& headers,
+                                                FilterState& filter_state) {
   utils::setStringFilterState(filter_state, utils::kFilterStateApiKey,
                               api_key_);
+
+  auto* entry = headers.get(Headers::get().Path);
+  if (entry) {
+    utils::setStringFilterState(filter_state, utils::kFilterStateOriginalPath,
```
tomboc, line 131: Would the filter state be updated if the path_matcher filter doesn't find a corresponding operation?

TAOXUY: You are right. We need put set this filterState before rejection in path matcher filter.

Issue #239, opened 2020-07-24
Bug in gcloud_build_image


ToddMorrill:
If a user doesn't pass the `-i` parameter to `gcloud_build_image` then `ESP_FULL_VERSION` doesn't get defined.
https://github.com/GoogleCloudPlatform/esp-v2/blob/db1ec6b904c9e0a72e3d05708c0eb62a83640f1b/docker/serverless/gcloud_build_image#L37

In turn, this causes errors when tagging the resulting image.
https://github.com/GoogleCloudPlatform/esp-v2/blob/db1ec6b904c9e0a72e3d05708c0eb62a83640f1b/docker/serverless/gcloud_build_image#L96

I removed `${ESP_FULL_VERSION}-` from the image tag since that seems to be in line with the name convention referred to in the documentation [here](https://cloud.google.com/endpoints/docs/openapi/get-started-cloud-run#configure_esp).

nareddyt:
I recently updated the script to automatically tag the image name with the latest ESP version.

That is odd. We do a check here: https://github.com/GoogleCloudPlatform/esp-v2/blob/db1ec6b904c9e0a72e3d05708c0eb62a83640f1b/docker/serverless/gcloud_build_image#L54

And then specifically set the `ESP_FULL_VERSION` at the bottom of the if statement.

Can you provide the exact error you received?

nareddyt:
Also FYI I am updating the documentation to refer to the new format.

ToddMorrill:
Here's the relevant portion of the error. TLDR; starting an image tag with `-` appears to be invalid.
```
BUILD
Already have image (with digest): gcr.io/cloud-builders/docker

                   ***** NOTICE *****

Alternative official `docker` images, including multiple versions across
multiple platforms, are maintained by the Docker Team. For details, please
visit https://hub.docker.com/_/docker.

                ***** END OF NOTICE *****
invalid argument "gcr.io/<my-project>/endpoints-runtime-serverless:-my-gateway-hash-uk.a.run.app-2020-07-24r0" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 125
--------------------------------------------------------------------------------------------------------

ERROR: (gcloud.builds.submit) build <build id> completed with status "FAILURE"
```

nareddyt:
[...]

PR #240, opened 2020-07-24
Fail `gcloud_build_image` early on error

nareddyt:
Will help users debug issues, like #239


nareddyt:
> My suspicious is CLI "gcloud container images" may fail in user host due to some reason. We need to check its error

Right, but this `set -e` will catch the error and exit the script. I think that's what we want, it will print out the exact error detail then.

PR #241, opened 2020-07-24
b/161827325: set `x-envoy-original-path` in backend routing filter

TAOXUY:
Set `x-envoy-original-path` in backend routing filter when it modifies the `path` if `x-envoy-original-path` header doesn't exist.

### DIFF
--- a/src/envoy/http/backend_routing/filter.cc
+++ b/src/envoy/http/backend_routing/filter.cc
@@ -112,6 +112,9 @@ FilterHeadersStatus Filter::decodeHeaders(
       NOT_REACHED_GCOVR_EXCL_LINE;
   }
 
+  if (!headers.EnvoyOriginalPath()) {
+    headers.setEnvoyOriginalPath(headers.getPathValue());
+  }
   headers.setPath(new_path);
   return FilterHeadersStatus::Continue;
 }
--- a/src/envoy/http/path_matcher/filter.cc
+++ b/src/envoy/http/path_matcher/filter.cc
@@ -74,7 +74,6 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
 
   Envoy::StreamInfo::FilterState& filter_state =
       *decoder_callbacks_->streamInfo().filterState();
-  utils::setStringFilterState(filter_state, utils::kFilterStatePath, path);
 
   const PathMatcherRule* rule = config_->findRule(method, path);
   if (rule == nullptr) {
--- a/src/envoy/http/path_matcher/filter.cc
+++ b/src/envoy/http/path_matcher/filter.cc
@@ -71,7 +71,6 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
 
   std::string method(headers.Method()->value().getStringView());
   std::string path(headers.Path()->value().getStringView());
-
   const PathMatcherRule* rule = config_->findRule(method, path);
   if (rule == nullptr) {
     rejectRequest(Envoy::Http::Code(404),
@@ -79,10 +78,10 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
     return Envoy::Http::FilterHeadersStatus::StopIteration;
   }
 
-  Envoy::StreamInfo::FilterState& filter_state =
-      *decoder_callbacks_->streamInfo().filterState();
   const absl::string_view operation = rule->operation();
   ENVOY_LOG(debug, "matched operation: {}", operation);
+  Envoy::StreamInfo::FilterState& filter_state =
+      *decoder_callbacks_->streamInfo().filterState();
   utils::setStringFilterState(filter_state, utils::kFilterStateOperation,
                               operation);
 
--- a/src/envoy/http/path_matcher/filter.cc
+++ b/src/envoy/http/path_matcher/filter.cc
@@ -25,6 +25,7 @@ namespace espv2 {
 namespace envoy {
 namespace http_filters {
 namespace path_matcher {
+using ::Envoy::Http::RequestHeaderMap;
 
 using ::espv2::api::envoy::v7::http::path_matcher::PathMatcherRule;
 using ::espv2::api::envoy::v7::http::path_matcher::PathParameterExtractionRule;
@@ -47,7 +48,7 @@ using RcDetails = Envoy::ConstSingleton<RcDetailsValues>;
 }  // namespace
 
 Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
-    Envoy::Http::RequestHeaderMap& headers, bool) {
+    RequestHeaderMap& headers, bool) {
   if (!headers.Method()) {
     rejectRequest(Envoy::Http::Code(400), "No method in request headers.");
     return Envoy::Http::FilterHeadersStatus::StopIteration;
@@ -70,6 +71,11 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
 
   std::string method(headers.Method()->value().getStringView());
   std::string path(headers.Path()->value().getStringView());
+
+  Envoy::StreamInfo::FilterState& filter_state =
+      *decoder_callbacks_->streamInfo().filterState();
+  utils::setStringFilterState(filter_state, utils::kFilterStatePath, path);
+
   const PathMatcherRule* rule = config_->findRule(method, path);
   if (rule == nullptr) {
     rejectRequest(Envoy::Http::Code(404),
@@ -79,8 +85,6 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
 
   const absl::string_view operation = rule->operation();
   ENVOY_LOG(debug, "matched operation: {}", operation);
-  Envoy::StreamInfo::FilterState& filter_state =
-      *decoder_callbacks_->streamInfo().filterState();
   utils::setStringFilterState(filter_state, utils::kFilterStateOperation,
                               operation);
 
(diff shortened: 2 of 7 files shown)


TAOXUY:
Hi @tomboc, please take a look.

src/envoy/http/path_matcher/filter.cc
```diff
@@ -70,6 +71,11 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
 
   std::string method(headers.Method()->value().getStringView());
   std::string path(headers.Path()->value().getStringView());
+
+  Envoy::StreamInfo::FilterState& filter_state =
+      *decoder_callbacks_->streamInfo().filterState();
+  utils::setStringFilterState(filter_state, utils::kFilterStatePath, path);
```
qiwzhang, line 77: path_matcher filter may go away.  I don't want to put too much logic here.

tests/integration_test/access_log_test.go
```diff
@@ -61,39 +62,64 @@ func TestAccessLog(t *testing.T) {
...
 	accessLogFormat := "\"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\"" +
 		"%RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT%" +
-		"\"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\"" +
-		"\"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\" " +
-		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.operation):60% " +
+		"\"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" " +
+		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.operation):70% " +
 		"%FILTER_STATE(com.google.espv2.filters.http.service_control.api_key):30% " +
-		"%FILTER_STATE(com.google.espv2.filters.http.service_control.path):60%" +
+		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.path):60%" +
```
qiwzhang, line 68: Such filter_state name is too much implemenation detail.   If we remove path_matcher,   it will break customers.

src/envoy/http/path_matcher/filter.cc
TAOXUY: ACK.

tests/integration_test/access_log_test.go
TAOXUY: ACK.

TAOXUY:
> There are many envoy rejected requests that path_matcher filter will not be called. But their access log will have :path, will not have origin-path. "rejected" by path_matched can be just like them.

IIUC, the access log is collected by HCM and if the request is already processed in http filter chains, it should hit our path matcher filter as it is the first one in ours. It seems to me that no such case where the accessLog logs `:path` while path matcher doesn't log it.

Admit it would be hard if path matcher will be refactored. Again, the `x-envoy-origin-path` can be maliciously faked by downstream client or not suppressed by downstream server. 

How about this we try remove `x-envoy-original-path`, and we can set "x-envoy-original-path" if it doesn't exist in backend routing filter.

qiwzhang:
We don't need to worry about malicious faked x-envoy-origin-path. We don't use it for access control, just for access log

qiwzhang:
the issue is: access log has "path", but it may or may not have "origin-path" depends on how the request are handled. Whoever using access_log have to handle the access log correctly, e.g. if "origin-path" is absent,  use "path" for origin path. 
With this:  we don't need to log "origin-path" in all cases.

qiwzhang:
With above, I still prefer to use "x-envoy-origin-path"

TAOXUY:
> the issue is: access log has "path", but it may or may not have "origin-path" depends on how the request are handled. Whoever using access_log have to handle the access log correctly, e.g. if "origin-path" is absent, use "path" for origin path.
> With this: we don't need to log "origin-path" in all cases.

FYI, `%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%` can check `X-ENVOY-ORIGINAL-PATH` first and then `PATH`.

TAOXUY:
> We don't need to worry about malicious faked x-envoy-origin-path. We don't use it for access control, just for access log

Not only the malicious attack, but also downstream envoy, I see some issues in envoy repo that they forgot suppress `x-envoy-` headers in `router` so I suspect it may happen in the future.

TAOXUY:
OK, since it is access log, let's put aside the issue of request with `x-envoy-original-path`.

TAOXUY:
PTAL

TAOXUY:
> Can we do it in two PR: one to rollback the last pr, and add new changes on the rollback.

https://github.com/GoogleCloudPlatform/esp-v2/pull/242

src/envoy/http/backend_routing/filter.cc
```diff
@@ -112,6 +112,9 @@ FilterHeadersStatus Filter::decodeHeaders(
       NOT_REACHED_GCOVR_EXCL_LINE;
   }
 
+  if (!headers.EnvoyOriginalPath()) {
+    headers.setEnvoyOriginalPath(headers.getPathValue());
```
nareddyt, line 116: Not clear to me: why do we need this check? This is the first filter that runs before gRPC transcoding. Can't we just always override it?

src/envoy/http/path_matcher/filter.cc
```diff
@@ -25,6 +25,7 @@ namespace espv2 {
 namespace envoy {
 namespace http_filters {
 namespace path_matcher {
+using ::Envoy::Http::RequestHeaderMap;
```
nareddyt, line 28: make clang-format

src/envoy/http/path_matcher/filter.cc
```diff
@@ -70,6 +71,10 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
 
   std::string method(headers.Method()->value().getStringView());
   std::string path(headers.Path()->value().getStringView());
+
+  Envoy::StreamInfo::FilterState& filter_state =
```
nareddyt, line 75: I assume the changes here are not needed?

src/envoy/http/backend_routing/filter.cc
nareddyt: Nvm, this is wrong. Makes sense.

tests/integration_test/access_log_test.go
```diff
@@ -61,37 +62,62 @@ func TestAccessLog(t *testing.T) {
 	// https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators
 	accessLogFormat := "\"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\"" +
 		"%RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT%" +
-		"\"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\"" +
-		"\"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\" " +
-		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.operation):60% " +
-		"%FILTER_STATE(com.google.espv2.filters.http.service_control.api_key):30%\n"
+		"\"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" " +
+		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.operation):70% " +
```
qiwzhang, line 66: in prepare removing path_matcher,  can we rename this filter_state key as 
"com.google.espv2.filters.http.service_control.api_method"
before it is used by api gateway. 

We can do it in another pr.

tests/integration_test/access_log_test.go
```diff
@@ -61,37 +62,62 @@ func TestAccessLog(t *testing.T) {
...
+			if err := s.Setup(args); err != nil {
+				t.Fatalf("fail to setup test env, %v", err)
+			}
+			defer s.TearDown(t)
+			makeOneRequest(t, s, tc.requestPath, tc.wantError)
 
-	if err := tryRemoveFile(accessLog); err != nil {
-		t.Fatalf("fail to remove accessLog file, %v", err)
+			bytes, err := ioutil.ReadFile(accessLog)
+			if err != nil {
```
qiwzhang, line 108: can we call it accessLogFile?

tests/integration_test/access_log_test.go
TAOXUY: ACK.

src/envoy/http/path_matcher/filter.cc
```diff
@@ -47,7 +48,7 @@ using RcDetails = Envoy::ConstSingleton<RcDetailsValues>;
 }  // namespace
 
 Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
-    Envoy::Http::RequestHeaderMap& headers, bool) {
+    RequestHeaderMap& headers, bool) {
```
tomboc, line 51: PR description says that path filterState is moved to path mather, but I cannot see any implementation changes to path_matcher/filter.cc.

tests/integration_test/access_log_test.go
TAOXUY: b/162231154

src/envoy/http/path_matcher/filter.cc
TAOXUY: done.

PR #243, opened 2020-07-25
[Don't submit]add service account file e2e test

TAOXUY:


nareddyt:
Yikes, prow did not like the test name change :)

Issue #248, opened 2020-07-29
Tracing sample rate not working


nareddyt:
_Originally posted by @EatonZ in https://github.com/GoogleCloudPlatform/esp-v2/issues/237#issuecomment-663915556_

By the way, I didn't need to set `--tracing_sample_rate` at all. Turns out all requests are being traced by default. Not sure if that's intended, but that is what I am seeing.🤔

nareddyt:
Thanks for reporting this issue. Turns I never tested the sample rate, I assumed the feature (which is part of upstream Envoy) would work as expected... oops

I am also able to reproduce with some integration tests in #249.

Need to investigate envoy's opencensus tracer extension. Relevant envoy code: https://github.com/envoyproxy/envoy/blob/6eb7e642d33f5a55b63c367188f09819925fca34/source/extensions/tracers/opencensus/opencensus_tracer_impl.h#L31

nareddyt:
Some implementation details for ESPv2 engineers.

Relevant portion of envoy bootstrap config when setting flag `--tracing_sample_rate=0.5`:

```
{
   "node":{
      ... trimmed ...
   },
   "staticResources":{
      ... trimmed ...
   },
   "dynamicResources":{
      ... trimmed ...
   },
   "tracing":{
      "http":{
         "name":"envoy.tracers.opencensus",
         "typedConfig":{
            "@type":"type.googleapis.com/envoy.config.trace.v3.OpenCensusConfig",
            "traceConfig":{
               "probabilitySampler":{
                  "samplingProbability":0.5
               },
               "maxNumberOfAttributes":"32",
               "maxNumberOfAnnotations":"32",
               "maxNumberOfMessageEvents":"128",
               "maxNumberOfLinks":"128"
            },
            "stackdriverExporterEnabled":true,
            "stackdriverProjectId":"test-project-id",
            "stackdriverAddress":"ipv4:127.0.0.1:21805"
         }
      }
   },
   "admin":{
      ... trimmed ...
   }
}
```

nareddyt:
Ah, I found the issue. Configuring tracing via bootstrap config is deprecated now. Instead it must be configured in HCM:

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-tracing-provider

[...]

PR #249, opened 2020-07-29
b/162452414: Fix tracing sample rate

nareddyt:
All requests are being traced by default. Root caused in #248, tldr:
- Trace config should be set in HCM, not bootstrap config (deprecated since we've implemented this feature)
- Envoy's OpenCensus tracer extension does not use OpenCensus sampling config to determine sample rate. It uses Envoy's HCM Tracer config instead.

Added integration test for sampling rate to confirm the fix works.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -930,8 +930,6 @@ def gen_proxy_config(args):
     if args.enable_debug:
         proxy_conf.append("--suppress_envoy_headers=false")
 
-
-
     return proxy_conf
 
 def gen_envoy_args(args):
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -58,26 +58,6 @@ def gen_bootstrap_conf(args):
     cmd = [BOOTSTRAP_CMD, "--logtostderr"]
 
     cmd.extend(["--admin_port", str(args.status_port)])
-    if args.disable_tracing:
-        cmd.append("--disable_tracing")
-    else:
-        if args.tracing_project_id:
-            cmd.extend(["--tracing_project_id", args.tracing_project_id])
-        if args.tracing_incoming_context:
-            cmd.extend(
-                ["--tracing_incoming_context", args.tracing_incoming_context])
-        if args.tracing_outgoing_context:
-            cmd.extend(
-                ["--tracing_outgoing_context", args.tracing_outgoing_context])
-        if args.cloud_trace_url_override:
-            cmd.extend(["--tracing_stackdriver_address",
-                        args.cloud_trace_url_override])
-
-        if args.disable_cloud_trace_auto_sampling:
-            cmd.extend(["--tracing_sample_rate", "0"])
-        elif args.tracing_sample_rate:
-            cmd.extend(["--tracing_sample_rate", str(args.tracing_sample_rate)])
-
     if args.http_request_timeout_s:
         cmd.extend(
             ["--http_request_timeout_s",
@@ -442,7 +422,6 @@ environment variable or by passing "-k" flag to this script.
         help="The Google project id for Stack driver tracing")
     parser.add_argument(
         '--tracing_sample_rate',
-        default=0.001,
         help='''
         Tracing sampling rate from 0.0 to 1.0.
         By default, 1 out of 1000 requests are sampled.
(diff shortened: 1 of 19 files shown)


nareddyt:
Hmm, the sampling rate is working, but it seems low for some reason... Need to look into it some more.

docker/generic/start_proxy.py
```diff
@@ -442,7 +422,6 @@ def make_argparser():
         help="The Google project id for Stack driver tracing")
     parser.add_argument(
         '--tracing_sample_rate',
-        default=0.001,
         help='''
         Tracing sampling rate from 0.0 to 1.0.
         By default, 1 out of 1000 requests are sampled.
```
TAOXUY, line 427: Instead of 0.001, so what is the default sample rate now?

nareddyt: It's still 0.001. It's set in config generator flags instead, and python will just not pass any value if the user doesn't override it.

This is a pattern @JLXIA and I have discussed in the past. Keep the python flags empty, and place the defaults in the config generator flags. Otherwise the defaults will be duplicated across binaries and can lead to problems in the future.

Added a test case for the default flag value in config generator.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -1808,10 +1812,12 @@ func TestMakeHttpConMgr(t *testing.T) {
...
 		},
 		{
-			desc: "Generate HttpConMgr when tracing is disabled",
+			desc: "Generate HttpConMgr when tracing is enabled",
 			opts: options.ConfigGeneratorOptions{
 				CommonOptions: options.CommonOptions{
-					DisableTracing: true,
+					DisableTracing:      false,
+					TracingProjectId:    "test-project",
+					TracingSamplingRate: 1,
```
TAOXUY, line 1820: So only the sample rate doesn't work in bootstrap?  
So the current sample rate in our code is 1?

nareddyt: Yup, apparently setting tracing in bootstrap was deprecated sometime within the last year. Envoy defaults to sample rate of 1 then.

TAOXUY: Does that mean all the ESPv2 users are using 100% tracing and no one find/complain till now?

nareddyt: Yup, for at least 5 months from when the deprecation happened.

EatonZ: Since tracing all the requests has been the (unintended) default behavior for many months now (possibly since project start?), it may make sense for the fix to set 1.0 as the new default, in case your other users have been relying on it, and to maintain backwards compatibility. Or you could just highlight it as a breaking change by making 0.001 work as intended, up to you.

nareddyt: Good question, but I think it should be fine:
1) The current behavior does not match the [documented behavior](https://cloud.google.com/endpoints/docs/openapi/specify-esp-v2-startup-options#tracing). In this case, making the bug fix with a breaking change is OK.
2) The breaking change does not affect client-facing behavior. If anything, requests will be slightly faster. So the impact of the breaking change is minimal.
3) The majority of Cloud Endpoints users are ESPv1 on GKE, which ESPv2 will "very soon" support. Our aim is to make the migration as easy as possible. So we don't want to change any flag defaults values that ESPv1 uses.

src/go/bootstrap/static/bootstrap.go
```diff
@@ -49,12 +49,6 @@ func ServiceToBootstrapConfig(serviceConfig *confpb.Service, id string, opts opt
 		return nil, err
 	}
 
-	if !opts.DisableTracing {
```
TAOXUY, line 52: https://github.com/GoogleCloudPlatform/esp-v2/blob/6e9e093240a79f9de0f97745c6db37e957b4071c/src/go/commonflags/flags.go#L38

Please categorize by comment which flags are used by bootstrap and which are used by configmanager

nareddyt: https://github.com/GoogleCloudPlatform/esp-v2/blob/6e9e093240a79f9de0f97745c6db37e957b4071c/src/go/commonflags/flags.go#L27

All of these *should* be used by bootstrap. But with the changes in this PR, the tracing ones are not.

As mentioned above, I will make a followup PR to clean this up. Don't want to do it in this PR, the diff will be too big.

src/go/tracing/tracing_test.go
```diff
@@ -340,16 +223,114 @@ func TestGCPTracingConfig(t *testing.T) {
...
+
+	testData := []struct {
+		desc              string
+		tracingSampleRate float64
+		wantResult        *hcmpb.HttpConnectionManager_Tracing
+		wantError         string
+	}{
+		{
+			desc:              "Default sampling rate works",
+			tracingSampleRate: options.DefaultCommonOptions().TracingSamplingRate,
```
qiwzhang, line 248: do we have tests for disabled,  rate=0 and ratte=1?  If no, please add them

nareddyt: Sure, added. There's nothing special about 0 and 1 anymore, it gets translated directly to 0% or 100%.

nareddyt: And there are integration tests already for rate 0 and 1.

PR #250, opened 2020-07-29
b/161200191,b/162231154: Add api_method in service control filter

TAOXUY:
Consider path matcher filter may be moved out of ESPv2 and using `xxx.path_matcher.operation` is not a long-standing solution for api-gateway to query the api-method log, so move this dependency on service control filter.

### DIFF
--- a/tests/integration_test/access_log_test.go
+++ b/tests/integration_test/access_log_test.go
@@ -63,7 +63,7 @@ func TestAccessLog(t *testing.T) {
 	accessLogFormat := "\"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\"" +
 		"%RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT%" +
 		"\"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" " +
-		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.operation):70% " +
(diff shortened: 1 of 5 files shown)


tests/integration_test/access_log_test.go
```diff
@@ -63,7 +63,7 @@ func TestAccessLog(t *testing.T) {
 	accessLogFormat := "\"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\"" +
 		"%RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT%" +
 		"\"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" " +
-		"%FILTER_STATE(com.google.espv2.filters.http.path_matcher.operation):70% " +
+		"%FILTER_STATE(com.google.espv2.filters.http.service_control.api_method):70% " +
```
nareddyt, line 66: Are you sure `api_method` is the correct terminology?

In the first test case below:
- Method name = EchoHeader
- Operation / selector = 1.echo_api_endpoints_cloudesf_testing_cloud_goog.EchoHeader

nareddyt: Well not a big deal to me. Just whatever will prevent confusion in the future. If we use `method` to refer to selector already, then this is fine.

TAOXUY: My initial thought is:
- `api_method` is more straightforward to user/api-gateway compared to operation/selector
- EchoHead can be short `api_method` and we use its full name 1.echo_api_endpoints_cloudesf_testing_cloud_goog.EchoHeader

PR #251, opened 2020-07-29
b/148306474,b/156410254: move generating access token by service account credential to configmanager

TAOXUY:
Add the `local access token server` to provide envoy process with access token

Details:
- setup http server along with the ads server
- add integration test
- add e2e test

Followups: 
- remove access token generated code in envoy


nareddyt:
I think you need to rebase on master...

TAOXUY:
PTAL

api/envoy/v7/http/service_control/config.proto
```diff
@@ -110,26 +110,22 @@ message FilterConfig {
...
+    // Uri used to fetch access token from Instance
     // Metadata Server.
     espv2.api.envoy.v7.http.common.HttpUri imds_token = 4;
 
-    // The local path or inline content of the service account json file used to
-    // generate access token.
-    espv2.api.envoy.v7.http.common.DataSource service_account_secret = 5;
-
     // Information used to fetch access token from Google Cloud IAM.
-    espv2.api.envoy.v7.http.common.IamTokenInfo iam_token = 6;
```
nareddyt, line 122: Keep these field numbers as is. Just add `reserved 5` to the top of this message. Then we don't need a new config version, this is completely backwards compatible for API Gateway.

prow/gcpproxy-e2e.sh
```diff
@@ -28,19 +28,20 @@ exit 1; }
 
 function runE2E() {
   local OPTIND OPTARG arg
-  while getopts :f:p:c:g:m:R:t: arg; do
+  while getopts :f:p:c:g:m:R:St: arg; do
```
nareddyt, line 31: Missing `:` ?

scripts/all-utilities.sh
```diff
@@ -326,7 +325,7 @@ function get_gcsrunner_image_release_name() {
 function get_tag_name() {
   local tag_format="%H"
   tag_name="$(git show -q HEAD --pretty=format:"${tag_format}")"
-  echo -n ${tag_name}
+  echo -n "92ed2e1b8a92a1bd3d73c2d605775a0a65828026"
```
nareddyt: Revert

src/envoy/http/service_control/service_control_call_impl.cc
```diff
@@ -61,31 +50,6 @@ void ServiceControlCallImpl::createImdsTokenSub() {
       });
 }
 
-void ServiceControlCallImpl::createTokenGen() {
-  const std::string service_control_audience =
-      filter_config_.service_control_uri().uri() + kServiceControlService;
-  sc_token_gen_ = token_subscriber_factory_.createServiceAccountTokenGenerator(
```
nareddyt, line 67: Remember to delete all the code for `createServiceAccountTokenGenerator(...)`.

Maybe in a followup PR. Diff is too big here.

api/envoy/v7/http/common/base.proto
```diff
@@ -67,20 +67,6 @@ message HttpUri {
...
-    // Local filesystem data source.
-    string filename = 1 [(validate.rules).string.min_bytes = 1];
-
-    // String inlined in the configuration.
-    string inline_string = 2 [(validate.rules).string.min_bytes = 1];
-  }
-}
-
 message AccessToken {
   oneof token_type {
```
nareddyt, line 71: `reserved 2`

src/go/configgenerator/listener_generator.go
```diff
@@ -623,21 +623,10 @@ func makeServiceControlFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
...
-			break
-		case *commonpb.AccessToken_ServiceAccountSecret:
-			filterConfig.AccessToken = &scpb.FilterConfig_ServiceAccountSecret{
-				ServiceAccountSecret: serviceInfo.AccessToken.GetServiceAccountSecret(),
-			}
-			break
-		default:
-			break
+		filterConfig.AccessToken = &scpb.FilterConfig_ImdsToken{
+			ImdsToken: serviceInfo.AccessToken.GetRemoteToken(),
```
nareddyt, line 627: `GetRemoteTokenUri` or `GetAccessTokenUri`

Otherwise it sounds like you're passing the actual token value.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -1472,6 +1472,101 @@ func TestPathMatcherFilter(t *testing.T) {
...
+    },`,
+		},
+		{
+			desc:              "get access token from the local access token server",
+			serviceAccountKey: "this-is-sa-cred",
+			wantPartialServiceControlFilter: `
+    "imdsToken": {
+      "cluster": "local-access-token-server-cluster",
+      "timeout": "30s",
+      "uri": "http://127.0.0.1:8791/v1/instance/service-accounts/default/token"
```
nareddyt, line 1541: Nit: The URL here does not match the IMDS url, it's missing `/computeMetadata/`

```
http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token
```

Probably doesn't matter. But should we keep them the same for consistency?

src/go/configgenerator/listener_generator.go
nareddyt: Nevermind, this is part of the proto file. Then it's fine, no need to change the name unless you think it's worth it.

src/go/configinfo/service_info_test.go
```diff
@@ -2208,3 +2209,57 @@ func TestProcessTypes(t *testing.T) {
...
+			},
+		},
+	}
+	testCases := []struct {
+		desc              string
+		serviceAccountKey string
+		wantAccessToken   *commonpb.AccessToken
+	}{
+		{
+			wantAccessToken: &commonpb.AccessToken{
```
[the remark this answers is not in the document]
nareddyt: This is not resolved

src/go/configmanager/main/server.go
```diff
@@ -74,6 +77,15 @@ func main() {
 		grpcServer.Stop()
 	}()
 
+	if opts.ServiceAccountKey != "" {
+		// Setup local service-account-generated token server
+		r := tokengenerator.MakeLatsHandler(opts.ServiceAccountKey)
+		go func() {
+			_ = http.ListenAndServe(fmt.Sprintf(":%v", opts.LocalAccessTokenServerPort), r)
```
nareddyt, line 84: You're ignoring the error here. We have to handle it.

Are you sure this needs to be in a goroutine? I thought `ListenAndServe` would handle the blocking call internally.

src/go/util/util.go
```diff
@@ -94,6 +94,9 @@ const (
 	// The metadata server cluster name.
 	MetadataServerClusterName = "metadata-cluster"
 
+	// The local access token  server cluster name.
+	LatsClusterName = "local-access-token-server-cluster"
```
nareddyt, line 98: remove `server`, similar to the IMDS one above

tests/e2e/scripts/prow-utilities.sh
```diff
@@ -294,7 +295,11 @@ function download_client_binaries() {
 
 function get_apiproxy_service() {
   if [[ "${1}" == "bookstore" ]]; then
-    echo "bookstore.endpoints.cloudesf-testing.cloud.goog"
+    if [ -n "${2}" ]; then
+      echo "bookstore-using-sa-cred.endpoints.cloudesf-testing.cloud.goog"
```
nareddyt, line 299: Why do we need a different Endpoints Service for this? We should be able to re-use the bookstore one, right?

Conceptually, deploying with or without the SA does not require any difference in the service. So our tests should follow that.

prow/gcpproxy-e2e.sh
TAOXUY: No. It is using -S without value.

src/envoy/http/service_control/service_control_call_impl.cc
TAOXUY: Mentioned in description.

src/go/configmanager/main/server.go
TAOXUY: ListenAndServe is blocking 
https://blog.zikes.me/post/serving-multiple-sites-from-one-go-process/

tests/e2e/scripts/prow-utilities.sh
TAOXUY: It is unrelated by SA.

The reason is that two tests will deploy the same http bookstore service in the same test causing one deployment fail.

src/go/tokengenerator/token_generator_test.go
```diff
@@ -0,0 +1,110 @@
...
+	token, duration, err := generateAccessTokenFromData(fakeKeyData)
+	if token != "ya29.new" || duration.Seconds() < 3598 || err != nil {
+		t.Errorf("Test : Fail to make access token, got token: %s, duration: %v, err: %v", token, duration, err)
+	}
+
+	latestFakeToken := `{"access_token": "ya29.latest", "expires_in":3599, "token_type":"Bearer"}`
+	mockTokenServer.SetResp(latestFakeToken)
+
+	// The token is cached so the old token gets returned.
+	token, duration, err = generateAccessTokenFromData([]byte("Invalid data, not a service account"))
```
qiwzhang, line 48: This is very confusing.   Should we cache the token by using service_account_key data as the key?  Now we are assuming we only support one service account key, which is the case,  I am wondering if we need to support multiple.  or by the file path?

src/go/tokengenerator/token_generator.go
```diff
@@ -12,26 +12,33 @@
...
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 )
 
 var (
 	_GOOGLE_API_SCOPE = []string{
+		// Call servicemanagement to fetch service config.
 		"https://www.googleapis.com/auth/service.management.readonly",
+		// Call servicecontrol to get latest rollout id.
+		"https://www.googleapis.com/auth/servicecontrol",
```
qiwzhang, line 35: e need to add QuotaController

src/go/configgenerator/cluster_generator.go
```diff
@@ -40,12 +40,27 @@ func MakeClusters(serviceInfo *sc.ServiceInfo) ([]*clusterpb.Cluster, error) {
...
+		metadataCluster, err := makeMetadataCluster(serviceInfo)
+		if err != nil {
+			return nil, err
+		}
+		if metadataCluster != nil {
+			clusters = append(clusters, metadataCluster)
+		}
+
+	} else {
+		latsCluster := makeLatsCluster(serviceInfo)
```
qiwzhang, line 60: Lats is confusing.   Can we call it TokenServer?

src/go/configinfo/service_info.go
```diff
@@ -243,27 +242,30 @@ func (s *ServiceInfo) addGrpcHttpRules() {
...
 		s.AccessToken = &commonpb.AccessToken{
-			TokenType: &commonpb.AccessToken_ServiceAccountSecret{
-				ServiceAccountSecret: &commonpb.DataSource{
-					Specifier: &commonpb.DataSource_InlineString{
-						InlineString: string(data),
-					},
+			TokenType: &commonpb.AccessToken_RemoteToken{
+				RemoteToken: &commonpb.HttpUri{
+					// Use the same path as IMDS to get access token.
+					Uri:     fmt.Sprintf("http://%s:%v%s", util.LoopbackIPv4Addr, s.Options.LocalAccessTokenServerPort, util.AccessTokenPath),
```
qiwzhang, line 249: Why are we using the same access token path as metadata server?   I could not come up any reason why not.  But still prefer not to use the same path.    how about just /local/access_token?

src/go/configmanager/flags/flags.go
```diff
@@ -66,6 +66,7 @@ var (
 	ServiceAccountKey = flag.String("service_account_key", "", `Use the service account key JSON file to access the service control and the
 	service management.  You can also set {creds_key} environment variable to the location of the service account credentials JSON file. If the option is
   omitted, the proxy contacts the metadata service to fetch an access token`)
+	LocalAccessTokenServerPort = flag.Uint("local_access_token_server_port", 8791, "Port that configmanager use to setup server to provide envoy with access token using service account credential, for accessing servicecontrol.")
```
qiwzhang, line 69: just call it token_agent_port?

src/go/configgenerator/cluster_generator.go
TAOXUY: Admit it is a little bit confusing at the first time but the reason I choose it:
- TokenServer is too general compared to local access token server(the full name of lates). 
- Also, we already use imds for instance metadata server and we clarify in the comments, which is the similar case.

src/go/tokengenerator/token_generator.go
TAOXUY: "https://www.googleapis.com/auth/servicecontrol" look enough. 
https://developers.google.com/identity/protocols/oauth2/scopes#servicecontrol

Adding e2e test for quota.

src/go/tokengenerator/token_generator_test.go
TAOXUY: Currently, we only have two types of access token, access servicemanagmenet/servicecontrol. 

Unnecessary to provide two service_account_key for each and separate the token?

qiwzhang: I mean to support multiple service account key paths.  If so, the token cache should be keyed by the file paths.

But for now, we only support one service account key file, so a global token cache is fine.

src/go/tokengenerator/token_generator.go
qiwzhang: Thanks, we need to make sure quotaController works for service account key file

src/go/configgenerator/cluster_generator.go
qiwzhang: OK, then.  I just need to get used to this name.

src/go/configmanager/flags/flags.go
qiwzhang: Not sure if this port is a security risk.  If a hacker can access this port, he can get access token.  But normally, we don't expose this port outside of container, it should be fine.

TAOXUY: This port should be treated same as discovery_port(the port of xds, 8790 by default), both of which should not be exposed.

src/go/configinfo/service_info.go
nareddyt: Curious about this too. I don't see any need for using the same path, we just need to maintain the same response format. Might make debugging easier too if we clearly distinguish between the paths.

src/go/configmanager/flags/flags.go
nareddyt: Should be fine. Our threat model indicates the riskiest threat is malicious clients attacking Envoy. Getting ssh access in the container / remote code execution is very low probability.

src/go/bootstrap/tracing_test.go
```diff
@@ -429,7 +429,7 @@ func runTest(_ *testing.T, shouldRunServer bool, f func()) {
 	if shouldRunServer {
 		// Run a mock server and point injected client to mock server
 		mockMetadataServer := util.InitMockServerFromPathResp(map[string]string{
-			util.ProjectIDSuffix: fakeMetadataProjectId,
+			util.ProjectIDPath: fakeMetadataProjectId,
```
nareddyt, line 432: FYI, you'll have to fix some merge conflicts :)

src/go/commonflags/flags.go
```diff
@@ -52,7 +52,7 @@ var (
 	//listener is marked as ready but the whole Envoy server is not marked as ready
 	//(worker did not start) somehow. To work around this problem, use IP for
 	//metadata server to fetch access token.
-	MetadataURL = flag.String("metadata_url", "http://169.254.169.254/computeMetadata", "url of metadata server")
+	MetadataURL = flag.String("metadata_url", "http://169.254.169.254", "url of metadata server")
```
nareddyt, line 55: Is this flag exposed so that API Producers can override it? If so, we can't make this change. This is a breaking change if a user overrides the value. Now we'll append `/computeMetadata` to it, even if they specify it.

Please check if this was exposed. Also consider the flag default in ESPv1.

src/go/configmanager/main/server.go
nareddyt: Then I recommend checking the error and logging it, it will help debugging when a shutdown occurs.

Recommend using `glog.Errorf(...)` here.

src/go/tokengenerator/token_generator_test.go
TAOXUY: ACK

src/go/bootstrap/tracing_test.go
TAOXUY: ACK

src/go/commonflags/flags.go
TAOXUY: Nop, it is not exposed in py start script.

src/go/tokengenerator/token_generator.go
```diff
@@ -106,7 +106,7 @@ func generateAccessToken(keyData []byte) (string, time.Duration, error) {
 func MakeLatsHandler(serviceAccountKey string) http.Handler {
 	r := mux.NewRouter()
 
-	r.PathPrefix(util.AccessTokenPath).Methods("GET").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	r.PathPrefix(util.LatsAccessTokenPath).Methods("GET").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
qiwzhang, line 109: Can we have consistent naming? for port, we call it TokenAgent,  or path we call it LatsAccessToken.   Can we all call it TokenAgent

src/go/tokengenerator/token_generator.go
TAOXUY: Quota test has been checked in.

TAOXUY:
> Suggest you don't rebase and force push, it makes it difficult to review on Github. Especially for such a large PR.

ACK the `force` push make confusing.
The force push is because of added merged PR on this branch so this force push is still keeping the history as it is.

Issue #237, opened 2020-07-22
Added Latency


EatonZ:
I have ESPv2 deployed as a Cloud Run service, and the backend is another Cloud Run service hosting my simple API.
Both are in us-central1.
Doing a quick TTFB test on my API through the ESPv2 Cloud Run URL yields an average of **68ms**.
Doing the same test on the API directly (essentially bypassing ESPv2) yields an average of **56ms**.

Is this amount of latency normal?
I recognize this project is still in beta - is there ongoing work being done to reduce latency before the GA release?

If you need more details from me, let me know. I'd be interested in seeing your results as well.

nareddyt:
Within ESPv2, the latency will drastically change based on what [filters](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/doc/architecture.md) are enabled. We automatically enable filters based on the API configuration you provide (OpenAPI spec or gRPC service config). Have you added any features such as API key checks, JWT token verification, gRPC transcoding, etc? Or do you expect your request to essentially pass through without any extra modifications / checks?

ESPv2 also heavily uses caching (eviction time ~5 minutes). Make sure run the same request multiple times and then measure the latency.

There are some items outside of ESPv2 control. When deploying ESPv2 on Cloud Run, you are adding in an extra network hop before the request hits your backend server. This is inside Google Cloud's data centers, so it shouldn't be too bad, but it may add a few ms. This is an inherent problem in a serverless deployment.

If you have a strong requirement for low latency, then you may want to try using [Cloud Endpoints for GKE](https://cloud.google.com/endpoints/docs/openapi/get-started-kubernetes-engine) and compare the results.

Unfortunately we do not have performance benchmarks that test with complex configuration. For some of our simple E2E tests on serverless, I only see a latency overhead of 6ms for 95th percentile of traffic.

qiwzhang:
You can also try Cloud Run for [Anthos](https://cloud.google.com/endpoints/docs/openapi/get-started-cloud-run-anthos).  ESPv2 is running at the same cluster as your backend.  It network latency between ESPv2 and your backend should be much smaller.

EatonZ:
Thank you for the responses!

@nareddyt:

> Have you added any features such as API key checks, JWT token verification, gRPC transcoding, etc? Or do you expect your request to essentially pass through without any extra modifications / checks?

None of that - it's just a simple "hello world" JSON API. Do I need to try and explicitly disable anything?

[...]

PR #252, opened 2020-07-30
Switch e2e tests to test tracing with W3C trace context propagation

nareddyt:
GPC's serverless offerings will all start supporting `traceparent` headers soon for integration with OpenTelemetry. go/w3c-trace-context-guidelines

Switch our e2e tests to also use this header for trace context propagation. This will help give a better view of networking latency mentioned in #237. For each compute platform, we can determine exactly how much time is spent between the request leaving ESPv2 and entering the backend.

Here is an example of a cold start that we can now detect because of this PR. My plan is to let this run in our e2e tests for a while, and the evaluate if we should make this the default option for `--tracing_incoming_context` and `--tracing_outgoing_context`. It will make it easier for API Producers to debug cold starts.

image


tests/e2e/scripts/cloud-run/deploy.sh
```diff
@@ -324,7 +324,7 @@ function setup() {
     -i "${APIPROXY_IMAGE}"
 
   # Redeploy ESPv2 to update the service config
-  proxy_args="^++^--tracing_sample_rate=0.01++--tracing_outgoing_context=x-cloud-trace-context"
+  proxy_args="^++^--tracing_sample_rate=0.05++--tracing_outgoing_context=traceparent++--tracing_incoming_context=traceparent"
```
TAOXUY, line 327: Why don't we set tracing_incoming_context before?

TAOXUY (approved):

tests/e2e/scripts/cloud-run/deploy.sh
nareddyt: Technically it's not needed yet. But soon Cloud Run will start publishing traces as well, and we want to keep propagating the trace context.

nareddyt: This will matter as soon as b/151160669 is fixed. Seems like there is progress maybe made on it.

EatonZ:
@nareddyt Is `x-cloud-trace-context` essentially being deprecated across GCP soon in favor of `traceparent`?

nareddyt:
@EatonZ Recently, there is a unified effort across all GCP serverless products (Cloud Run, Cloud Functions, App Engine) to improve application logging and tracing. They are working on adding new features, such as more traces that display cold start times, networking hops, etc. This will all be backed by OpenTelemetry, which uses the standard `traceparent` header.

Currently our serverless infra is in a transition phrase - it supports both `x-cloud-trace-context` and `traceparent` headers. So you can use either one, they information will be mirrored. But the long-term goal is to switch over to the standard `traceparent` to support distributed tracing across all applications, even your own applications you deploy on GCP. I don't know how long this transition phrase will occur or if they will fully deprecate `x-cloud-trace-context`, but they definitely prefer `traceparent` if possible.

Issue #254, opened 2020-07-31
Proto constraint validation failed (Duplicated pattern or invalid pattern)


glassonion1:
The following error occurs even though the URL pattern is different.
```
esp    | I0731 10:22:58.984492       6 config_manager.go:297] respond type.googleapis.com/envoy.config.listener.v3.Listener[] version "" with version "2020-07-XXr1"
esp    | W0731 10:22:59.014 19 envoy] [19][config][external/envoy/source/common/config/grpc_subscription_impl.cc:100] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) ingress_listener: Proto constraint validation failed (Duplicated pattern or invalid pattern): uri_template: \"/v1/companies/{id}/users/{user_id}\"\nhttp_method: \"OPTIONS\"
```

My gRPC service is below.
```
service UserService {
  rpc Get(GetRequest) returns (User) {
    option (google.api.http) = {
      get: "/v1/companies/{id}/users/{user_id}"
    };
  }
  rpc Update(UpdateRequest) returns (User) {
    option (google.api.http) = {
      put: "/v1/companies/{id}/users/{user.id}"
      body: "user"
    };
  }
}
```

My protobuf messages below.
```
message GetRequest {
  string id = 1;
  string user_id = 2;
}

message User {
  string id = 1;
  string name = 2;
}

message UpdateRequest {
  string id = 1;
  User user = 2;
}
```

Is there a good way to avoid this error?

nareddyt:
Huh, that's interesting. That is not the intended behavior, it should be supported.

I'm having trouble reproducing it. But I think this only happens if you enable CORS in ESPv2 (your backend does not support it, so you let ESPv2 handle it):
- Do you have CORS configuration enabled in ESPv2? For example, are you using any of the CORS flags [documented here](https://cloud.google.com/endpoints/docs/openapi/specify-esp-v2-startup-options#cors)?
- If so, can you try disabling CORS and see if it works?

nareddyt:
Or does your backend support CORS, and you enabled `allowCors: True` under `x-google-endpoints` under your service config?

Can you temporarily delete that and see if the error still occurs? I'm fairly certain you won't see the error anymore if you do this, but I just want to confirm.

[...]

PR #255, opened 2020-07-31
Deduplicate semantically equivalent paths generated by `allowCors`

nareddyt:
**Background:** When `AllowCors` is enabled, ESPv2 will generate `HttpRules` that allow `OPTIONS` requests for each path to pass through to the backend. We currently have some logic to deduplicate equivalent paths across multiple `Methods`.

**Problem**: The paths can have path parameters. If two paths have different parameter names, then our current deduplication logic will treat them as distinct paths. When we push this config to Path Matcher, it fails because Path Matcher understands the path parameter names do not matter in route matching.

See #254 for a customer running into this problem. While it's unlikely to happen, their use case makes sense to me.

**Fix**: We already have some _hacky_ regex that will construct a URL where any path parameters match. Re-use this for our path deduplication. Now our deduplication logic can handle semantically equivalent paths, just like Path Matcher.

**There is one drawback**: When we generate `Methods` for `OPTIONS` requests, we generate the operation name in a specific format:
```
Configured Path:   /shelves/{shelf_id}/books/{book.id}
Operation Name:   CORS_shelves_shelf_id_books_book.id
```

But now that we support looser path deduplication, the operation name is not accurate anymore. It could be a request for a similar path, but with different variable bindings (example: `/shelves/{shelf_id}/books/{b_id}`. This results extra REPORTs for the first path, and no REPORTs for any semantically equivalent paths.

It should be fine though, this is a minor bug and most users probably do not care about which `OPTIONS` requests are going to their backends.

**Testing**: 
- Unit tests to verify paths are correctly deduplicated now.
- Integration test. Without the fix, it results in a continuous loop with the following error message:
```
[2020-07-31 11:59:22.910][3185866][warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:100] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) ingress_listener: Proto constraint validation failed (Duplicated pattern or invalid pattern): uri_template: "/bookstore/shelves/{shelf_with_different_path_parameter}" http_method: "OPTIONS"
```

### DIFF
--- a/src/go/configgenerator/route_generator.go
+++ b/src/go/configgenerator/route_generator.go
@@ -16,7 +16,6 @@ package configgenerator
 
 import (
 	"fmt"
-	"regexp"
 	"time"
 
 	"github.com/GoogleCloudPlatform/esp-v2/src/go/configinfo"
@@ -239,11 +238,16 @@ func makeHttpRouteMatcher(httpRule *commonpb.Pattern) *routepb.RouteMatch {
 		return nil
 	}
 	var routeMatcher routepb.RouteMatch
-	re := regexp.MustCompile(`{[^{}]+}`)
 
-	// Replace path templates inside "{}" by regex "[^\/]+", which means
-	// any character except `/`, also adds `$` to match to the end of the string.
-	if re.MatchString(httpRule.UriTemplate) {
+	regex, err := util.GetRegexForUriWithPathParams(httpRule.UriTemplate)
+	if err != nil {
+		// Match with HttpHeader method. Some methods may have same path.
+		routeMatcher = routepb.RouteMatch{
+			PathSpecifier: &routepb.RouteMatch_Path{
+				Path: httpRule.UriTemplate,
+			},
+		}
+	} else {
 		routeMatcher = routepb.RouteMatch{
 			PathSpecifier: &routepb.RouteMatch_SafeRegex{
 				SafeRegex: &matcher.RegexMatcher{
@@ -254,18 +258,12 @@ func makeHttpRouteMatcher(httpRule *commonpb.Pattern) *routepb.RouteMatch {
 							},
 						},
 					},
-					Regex: re.ReplaceAllString(httpRule.UriTemplate, `[^\/]+`) + `$`,
+					Regex: regex,
 				},
 			},
 		}
-	} else {
-		// Match with HttpHeader method. Some methods may have same path.
-		routeMatcher = routepb.RouteMatch{
-			PathSpecifier: &routepb.RouteMatch_Path{
-				Path: httpRule.UriTemplate,
-			},
-		}
 	}
+
 	routeMatcher.Headers = []*routepb.HeaderMatcher{
 		{
 			Name: ":method",
--- a/src/go/configgenerator/route_generator.go
+++ b/src/go/configgenerator/route_generator.go
@@ -239,8 +239,8 @@ func makeHttpRouteMatcher(httpRule *commonpb.Pattern) *routepb.RouteMatch {
 	}
 	var routeMatcher routepb.RouteMatch
 
-	regex, err := util.GetRegexForUriWithPathParams(httpRule.UriTemplate)
-	if err != nil {
+	regex := util.WildcardMatcherForPath(httpRule.UriTemplate)
+	if regex == "" {
 		// Match with HttpHeader method. Some methods may have same path.
 		routeMatcher = routepb.RouteMatch{
 			PathSpecifier: &routepb.RouteMatch_Path{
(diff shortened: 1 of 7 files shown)


nareddyt:
@qiwzhang @TAOXUY PTAL

src/go/configinfo/service_info.go
```diff
@@ -321,7 +321,11 @@ func addHttpRule(method *methodInfo, r *annotationspb.HttpRule, httpPathWithOpti
 			UriTemplate: r.GetCustom().GetPath(),
 			HttpMethod:  r.GetCustom().GetKind(),
 		}
-		httpPathWithOptionsSet[r.GetCustom().GetPath()] = true
+		matcher, err := util.GetRegexForUriWithPathParams(r.GetCustom().GetPath())
```
qiwzhang, line 324: I don't understand here.  Why we only call GetRegex for "custom" type,  not for other methods?

qiwzhang: I see.  the code assume Custom is OPTIONS.  actually,  custom can be any method.   We should check
if r.GetCustom().GetKind() == OPTIONS

src/go/configgenerator/route_generator.go
```diff
@@ -239,11 +238,16 @@ func makeHttpRouteMatcher(httpRule *commonpb.Pattern) *routepb.RouteMatch {
 		return nil
 	}
 	var routeMatcher routepb.RouteMatch
-	re := regexp.MustCompile(`{[^{}]+}`)
 
-	// Replace path templates inside "{}" by regex "[^\/]+", which means
-	// any character except `/`, also adds `$` to match to the end of the string.
-	if re.MatchString(httpRule.UriTemplate) {
+	regex, err := util.GetRegexForUriWithPathParams(httpRule.UriTemplate)
```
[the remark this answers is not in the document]
TAOXUY: So wildcard is the only case of using regex here?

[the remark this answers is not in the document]
TAOXUY: Can we comment all the possible route cases here?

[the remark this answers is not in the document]
nareddyt: - Changed function to return single string, no error or boolean needed.
- Renamed it, in case it was not clear.
- No, this method supports any URI that has path parameters. See the test. It has nothing to do with wildcards inside the path params, instead the path parameters should be treated as wildcards.

src/go/configinfo/service_info.go
nareddyt: Good point, done. I assume custom options will only be used for OPTIONS, but we can double check to be sure.

src/go/configinfo/service_info.go
```diff
@@ -317,15 +317,20 @@ func addHttpRule(method *methodInfo, r *annotationspb.HttpRule, httpMatcherWithO
...
 		httpRule = &commonpb.Pattern{
 			UriTemplate: r.GetCustom().GetPath(),
-			HttpMethod:  r.GetCustom().GetKind(),
+			HttpMethod:  httpMethod,
 		}
-		matcher, err := util.GetRegexForUriWithPathParams(r.GetCustom().GetPath())
-		if err != nil {
-			matcher = r.GetCustom().GetPath()
+
+		if httpMethod == util.OPTIONS {
```
qiwzhang, line 326: Could you add a test for a non OPTIONS custom http rule and verify allowCors will create a rule for OPTIONS

nareddyt: Done

src/go/util/url_util_test.go
```diff
@@ -450,6 +450,11 @@ func TestWildcardMatcherForPath(t *testing.T) {
 			uri:         "/shelves/{shelf_id}/books/{book.id}",
 			wantMatcher: `/shelves/[^\/]+/books/[^\/]+$`,
 		},
+		{
+			desc:        "Path param with wildcard",
+			uri:         "/test/{x=*}/test/{y=**}",
+			wantMatcher: `/test/[^\/]+/test/[^\/]+$`,
```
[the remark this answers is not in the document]
nareddyt: I'm not sure what the best way to fix would be. Unless we have another regex that looks for {x=**} patterns. But that feels hacky to me.

Can we clean fix this issue in a separate PR? This is a bug ESPv2 currently has. We don't build the route match paths correctly for `**`, so I think ESPv2 doesn't actually work for `**` configurations.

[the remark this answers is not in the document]
qiwzhang: Sure,  let us file an issue and fix it in a separate PR.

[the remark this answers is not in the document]
nareddyt: b/162888708

src/go/configinfo/service_info.go
```diff
@@ -361,9 +370,13 @@ func (s *ServiceInfo) processHttpRule() error {
 			method := s.Methods[r.GetSelector()]
 			for _, httpRule := range method.HttpRule {
 				if httpRule.HttpMethod != "OPTIONS" {
```
TAOXUY, line 372: NIT: replace with util.OPTIONS

nareddyt: Will do in the next PR, then I won't be blocked.

PR #256, opened 2020-08-01
b/161479450: verify transcoding Any type

TAOXUY:
Add an integration test for transcoding any type.


tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto
```diff
@@ -172,6 +165,17 @@ message GetShelfRequest {
...
 }
 
+// A shelf resource.
+message Shelf {
+  // A unique shelf id.
+  int64 id = 1;
+  // A theme of the shelf (fiction, poetry, etc).
+  string theme = 2;
+
+  google.protobuf.Any any = 3;
```
nareddyt, line 175: What does the bookstore gRPC server do with this Any? I don't see any code in this PR that changes the bookstore gRPC server.

I would prefer if you could add some validation on the backend to ensure it receives the correct type. For example, in your integration test you send a `type.googleapis.com/endpoints.examples.bookstore.Book`. So have the bookstore server verify it can unmarshal the `Any` into a `Book`.

TAOXUY: Done.

qiwzhang: It is different problems:  from JSON to proto.Any on the request path, and from proto.Any to JSON on the response path. 

We need to test both.

tests/endpoints/bookstore_grpc/server/server.go
```diff
@@ -266,6 +267,22 @@ func (s *BookstoreServerV1Impl) CreateShelf(ctx context.Context, req *bspbv1.Cre
...
+	if req.Shelf.Any != nil {
+		var book bspbv1.Book
+		err := ptypes.UnmarshalAny(req.Shelf.Any, &book)
+		if err != nil {
+			return nil, status.New(codes.Internal, "cannot unmarshal shelf.any").Err()
+		}
+
+		newBook := book
+
+		req.Shelf.Any, err = ptypes.MarshalAny(&newBook)
```
qiwzhang, line 280: What is the purpose of marshaling into Any again?  I think we don't need it

tests/endpoints/bookstore_grpc/server/server.go
```diff
@@ -266,6 +267,22 @@ func (s *BookstoreServerV1Impl) CreateShelf(ctx context.Context, req *bspbv1.Cre
 		return nil, err
 	}
 
+	// Unmarshal, deepcopy and marshal, to verify the received binary `any`
+	if req.Shelf.Any != nil {
+		var book bspbv1.Book
+		err := ptypes.UnmarshalAny(req.Shelf.Any, &book)
```
qiwzhang, line 273: can we use a message type that is not used in any rpc call?  In OP config, this message has to be added into "types" field.  But in Envoy transcoding,  I hope the descriptor should include it. so it will work. But we need to test it out.

tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto
TAOXUY: The test added covers json -> any on the request path and any -> json on the response path.

tests/endpoints/bookstore_grpc/server/server.go
TAOXUY: The normal test procedure I think would be 
- sending a json request, getting/unmarshalling received any request and verify it
- sending an any response, getting received any response and verify it.

I choose to verify only on client side with the echoed request/response, since here we are using a real backend and I don't want add too much test checking logic there. 

We can believe json-any transcoding is correct by:
- the echoed json is correct
- the backend can successfully unmarshal the `Any`, which is more convincing than sending back binary `Any` directly.

tests/endpoints/bookstore_grpc/server/server.go
TAOXUY: ACK

TAOXUY: Done.

TAOXUY:
PTAL

tests/endpoints/bookstore_grpc/server/server.go
```diff
@@ -266,6 +267,21 @@ func (s *BookstoreServerV1Impl) CreateShelf(ctx context.Context, req *bspbv1.Cre
...
 
+	// Unmarshal, deepcopy and marshal, to verify the received binary `any`
+	if req.Shelf.Any != nil {
+		var obj bspbv1.ObjectOnlyForAny
+		err := ptypes.UnmarshalAny(req.Shelf.Any, &obj)
+		if err != nil {
+			return nil, status.New(codes.Internal, "cannot unmarshal shelf.any").Err()
+		}
+
+		newObj := obj
```
nareddyt, line 278: I don't think this will make a deep copy. It will copy the struct, but it wont copy members that are pointers inside the struct.

I don't think a deep copy is really needed, but would be nice to have. Otherwise, update the comment on line 270.

tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto
```diff
@@ -172,6 +165,22 @@ message GetShelfRequest {
   int64 shelf = 1;
 }
 
+message ObjectOnlyForAny {
+  int32 id = 1;
```
JLXIA, line 169: So that means the open source protocol tool actually compiled this into proto descriptor. 

No need to add it into Types, which is different from internal Blaze rule.

Also, Seems ESPv1 need to do this.

TAOXUY:
PTAL

tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto
JLXIA: Can you explain the reason?

TAOXUY: I suspected it is by some config option in api_service buildRule.
Raised the question on yaqs/5494949928370176.

PR #257, opened 2020-08-01
Add logic to delete temporary directory in gcloud_build_image

EatonZ:
When gcloud_build_image creates a temporary directory to store files for the build, that temporary directory is not deleted after the build has finished.
This PR adds directory deletion logic after the build.

### DIFF
--- a/docker/serverless/gcloud_build_image
+++ b/docker/serverless/gcloud_build_image
@@ -100,5 +100,8 @@ EOF
 
 NEW_IMAGE="gcr.io/${PROJECT}/endpoints-runtime-serverless:${ESP_FULL_VERSION}-${SERVICE}-${CONFIG_ID}"
 gcloud builds submit --tag "${NEW_IMAGE}" . --project="${PROJECT}"
+
+# Delete the temporary directory we created earlier.
+rm -r "${PWD}"
 )
 


docker/serverless/gcloud_build_image
```diff
@@ -100,5 +100,8 @@ EOF
 
 NEW_IMAGE="gcr.io/${PROJECT}/endpoints-runtime-serverless:${ESP_FULL_VERSION}-${SERVICE}-${CONFIG_ID}"
 gcloud builds submit --tag "${NEW_IMAGE}" . --project="${PROJECT}"
+
+# Delete the temporary directory we created earlier.
+rm -r "${PWD}"
```
qiwzhang, line 105: where do you go after deleting current folder?  Will this has any consequences on some OS?

EatonZ: I originally was going to cd to the home dir before deleting the tmp folder, but found it was unnecessary. When running this on WSL, it doesn't matter. Where does it put you after running on your machine?

qiwzhang: I just tried my Linux,  it is not deleted since you are in the current folder.  It is deleted after you move.

EatonZ: Ok, that's what I thought I had to do, but it wasn't the case in WSL. I will change it...

nareddyt: On ubuntu it keeps you in the tmp directory even though it's deleted. But I think it's fine, users can cd to wherever they want afterwords.

nareddyt: You can also add a `cd` if you'd like.

PR #259, opened 2020-08-04
Verify trace invariants in integration tests

nareddyt:
Modify our integration test framework to verify extra traces are not created.

- For most tests where tracing is disabled with `--disable_tracing`, the fake stackdriver server should receive no spans. We actually don't have any integration test to verify `--disable_tracing` works, so this will verify it across all tests.
- For tests where tracing is enabled, all spans should be accounted for at the end of the test. If there are extra traces, the test is verifying traces incorrectly, and this will fail the test.


Makefile
```diff
@@ -153,7 +153,7 @@ integration-test-run-sequential:
 	@go test -timeout 20m ./tests/endpoints/...
 	@go test -timeout 20m ./tests/env/... --logtostderr
 	@go test -timeout 20m ./tests/utils/... --logtostderr
-	@go test -timeout 50m -parallel 1 ./tests/integration_test --logtostderr
+	@go test -v -timeout 80m -parallel 1 ./tests/integration_test --logtostderr
```
nareddyt, line 156: These are intentional. 
- We were close to hitting the timeouts, increase them some more.
- `-v` to display all logs even when there is no test failure. I find this helpful, and having more logs does not hurt.

nareddyt:
PTAL @TAOXUY or @qiwzhang . Just an integration test change.

PR #260, opened 2020-08-05
Remove port replacement from GCS Runner

toddbeckman-g:
This is the job of the management and control planes, so by the time the runtime plane starts up, the port replacement is not only redundant but a vector for bugs.

This was originally used because Cloud Run did not guarantee the correct port was in use and recommended that containers read from the PORT variable. By switching off of Cloud Run, we can gaurantee the ports we want will be available.

TESTED=Ran E2E on our end using the image built in this PR.


nareddyt:
You can ignore that test failure. #251 will fix it.

toddbeckman-g:
Ah okay. Thanks.

PR #262, opened 2020-08-05
b/162888708: Support wildcards in envoy route matching with dynamic routing

nareddyt:
In dynamic routing mode, config manager constructs envoy route config per HttpRule. Envoy route config doesn't understand Google's Http Template syntax, so some regex re-writes must be done to support the most common OpenAPI cases.

This aims to support paths in form:
- /foo/*/bar
- /foo/**

Without this change, ESPv2 crash loops on invalid envoy config:
```
[2020-08-06 11:27:17.851][2809084][warning][config] [external/envoy/source/common/config/grpc_subscription_impl.cc:100] gRPC config for type.googleapis.com/envoy.config.listener.v3.Listener rejected: Error adding/updating listener(s) ingress_listener: bad repetition operator: **
```


src/go/util/url_util_test.go
```diff
@@ -446,14 +446,34 @@ func TestWildcardMatcherForPath(t *testing.T) {
...
+		{
+			desc:        "Path param with wildcard in segment binding",
 			uri:         "/test/{x=*}/test/{y=**}",
-			wantMatcher: `/test/[^\/]+/test/[^\/]+$`,
+			wantMatcher: `^/test/[^\/]+/test/.*$`,
+		},
+		{
+			desc:        "Path param with mixed wildcards",
+			uri:         "/test/{name=*}/test/**",
+			wantMatcher: `^/test/[^\/]+/test/.*$`,
```
qiwzhang, line 471: can we support following two cases?
/foo/*/bar
/foo/**/bar

nareddyt: - /foo/*/bar is supported, it's tested on line 460.
- /foo/** is supported, it's tested on line 460 as well.
- /foo/**/bar is also technically supported, but it is actually not valid syntax. From the HttpRule documentation:

> The syntax * matches a single URL path segment. The syntax ** matches zero or more URL path segments, which must be the last part of the URL path except the Verb.

But it turns out that our path matcher does support it:

https://github.com/GoogleCloudPlatform/esp-v2/blob/af02e4ecce9401f3287360ea12d52dcfab579ee3/src/api_proxy/path_matcher/http_template_test.cc#L218

So what would you recommend here? Explicitly support `**` where it's not the last path segment? Or just treat is as not supported?

qiwzhang: Well,  if htttp rule document say not need to support /foo/**/bar, then we don't need to support it.

qiwzhang:
Since you are here, I am wondering if it is worth the effort to  add some c++ unit-test to use Envoy routeMatch to test these regex

Just like path_match_test to test out the actual path match. 

Or we can trust Envoy RouteMatch with regex will do the right thing?

nareddyt:
> Since you are here, I am wondering if it is worth the effort to add some c++ unit-test to use Envoy routeMatch to test these regex
> 
> Just like path_match_test to test out the actual path match.
> 
> Or we can trust Envoy RouteMatch with regex will do the right thing?

Well, envoy regex route match uses [Google's RE2](https://github.com/google/re2) under the hood. I think it's safe to assume all of these will be supported.

The PR is not complete yet, i'll add a few integration tests for the regexes we use. We don't need to test every possible regex.

tests/integration_test/dynamic_routing_integration_test.go
```diff
@@ -191,6 +191,42 @@ func TestDynamicRouting(t *testing.T) {
...
+		},
+		{
+			desc:     "Succeed, CONSTANT_ADDRESS with multiple segments in double wildcards",
+			path:     "/wildcard/a/1/b/2/c/3/4/5",
+			method:   "GET",
+			wantResp: `{"RequestURI":"/dynamicrouting/const_wildcard?name=2"}`,
+		},
+		{
+			desc:     "Succeed, CONSTANT_ADDRESS with no segments in double wildcards",
+			path:     "/wildcard/a/1/b/2/c/",
```
qiwzhang, line 208: What will happen for "/wildcard/a/1/b/2/c"  without last "/"?

nareddyt: Good catch, it didn't work. I fixed it.

src/go/util/url_util.go
```diff
@@ -38,7 +38,22 @@ const (
...
+	pathParamMatcher = regexp.MustCompile(`/{[^{}]+}`)
+	// Replace path templates with double wildcards: /v1/{name=**}
+	pathParamDoubleWildcardMatcher = regexp.MustCompile(`/{[^{}]+=\*\*}`)
+
+	// Common regex forms that emulate http template syntax.
+	// Matches 1 or more segments of any character except '/'.
+	singleWildcardReplacementRegex = `/[^\/]+`
+	// Matches any character or no characters at all.
+	// Note this also allows paths without a '/' suffix.
+	doubleWildcardReplacementRegex = `.*`
```
qiwzhang, line 56: I looked at the definition of url_template rules:  Its syntax:

// ### Path template syntax
//
//     Template = "/" Segments [ Verb ] ;
//     Segments = Segment { "/" Segment } ;
//     Segment  = "*" | "**" | LITERAL | Variable ;
//     Variable = "{" FieldPath [ "=" Segments ] "}" ;
//     FieldPath = IDENT { "." IDENT } ;
//     Verb     = ":" LITERAL ;

It has following forms
*
**
{x} which is same as {x=*}
{x=**}
{x=/foo/bar*}
{x=/foo/*/bar/**}

We could not support last two forms now.   

I am not sure how easy to support them.

tests/integration_test/dynamic_routing_integration_test.go
nareddyt: I see, I reverted that change.

src/go/util/url_util.go
nareddyt: As stated in the PR description, let's focus on only supporting patterns that the OpenAPI compiler will generate for OpenAPI spec. For gRPC transcoding errors, we can handle them on a case-by-case basis later.

OpenAPI compiler will primarily generate:
- /foo/{x}
- /foo/**

This PR handles both of them. It also handles some other common cases like:
- /foo/*
- /foo/{x=**}

It will definitely not work for:
- {x=/foo/bar}
- {x=/foo/*/bar/**}

I think that's fine.

tests/integration_test/dynamic_routing_integration_test.go
nareddyt: Still kept the test case to show 404.

src/go/util/url_util_test.go
```diff
@@ -446,14 +446,37 @@ func TestWildcardMatcherForPath(t *testing.T) {
...
-			desc:        "Multiple path params",
+			desc:        "Path params with fieldpath-only bindings",
 			uri:         "/shelves/{shelf_id}/books/{book.id}",
-			wantMatcher: `/shelves/[^\/]+/books/[^\/]+$`,
+			wantMatcher: `^/shelves/[^\/]+/books/[^\/]+$`,
 		},
 		{
-			desc:        "Path param with wildcard",
+			desc:        "Path param with wildcard segments",
+			uri:         "/test/*/test/**",
```
TAOXUY, line 455: Do we have url pattern conflict check for duplicate regex?
/test/*
/test/{x=2}

I remember api compiler has this check but it is good to have on control plane if we don't have.

nareddyt: Path matcher filter will reject the config if there are duplicated http templated paths. We don't need another validation for the hacky regex conversion in config generator, since the end result to the user will be the same.

nareddyt: To clarify, duplicated path templates will account for the two cases you provided above.

PR #264, opened 2020-08-07
b/163118657: Remove transformation logic from GCS Runner.

toddbeckman-g:
Doing this in the management plane yields greater dev velocity and reduces a vector for errors.


nareddyt:
Do you want me to merge this? Or do you need any more reviews?

toddbeckman-g:
This should be fine as is.

PR #266, opened 2020-08-07
b/163128449: add e2e quota test on http bookstore

TAOXUY:
Details:
- Add an api in openapi config with quota limit
- Call the api beyond its quota limit
- Verify the count of code 200 and code 429(too many request)


tests/e2e/client/apiproxy_bookstore_quota_test.py
```diff
@@ -0,0 +1,132 @@
...
+    FLAGS.verbose = False
+    self.set_verbose(FLAGS.verbose)
+
+    # exhaust the quota in the current window.
+    print("Exhaust current quota...");
+    response = self._call_http(path='/quota_read',
+                               api_key=FLAGS.api_key)
+    if response.status_code != 429:
+      while True:
+        response = self._call_http(path='/quota_read',
```
nareddyt, line 54: Nit: Maybe add a max limit to the number of requests so we don't infinite loop if ESPv2 doesn't track quota correctly? Same for other while True loops here.

nareddyt:
I took a look at the e2e logs. Pretty interesting to see our quota control functionality in an e2e tests, thanks for adding this.

tests/e2e/client/apiproxy_bookstore_quota_test.py
TAOXUY: Done.

tests/e2e/client/apiproxy_bookstore_quota_test.py
```diff
@@ -50,20 +50,24 @@ def verify_quota_control(self):
...
                                    api_key=FLAGS.api_key)
         if response.status_code == 429:
           break;
+        elif i == 99:
+          sys.exit(utils.red("Fail to exhaust quota"))
 
     # waiting for the next quota refill.
     print("Wait for the next quota refill...");
-    while True:
+    for _ in range(100):
```
nareddyt, line 63: need `i` here, thought i guess your current code will cause the job to fail anyways with a runtime error :)

TAOXUY: Good catch, though it hasn't been found by test due to the first quota refill protection.

Issue #258, opened 2020-08-03
Errors when building on macOS


sgammon:
Hello esteemed ESPv2 authors,

I am having trouble building ESPv2 via Bazel on macOS. Here is the error I get:
```
2020/08/03 01:36:35 Downloading https://releases.bazel.build/3.0.0/release/bazel-3.0.0-darwin-x86_64...
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
DEBUG: Rule 'io_bazel_rules_python' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1523916428 -0700"
DEBUG: Call stack for the definition of repository 'io_bazel_rules_python' which is a git_repository (rule definition at /private/var/tmp/_bazel_sam.g/c5262b9dd666f2a9cad5b9bc4d2c9fe6/external/bazel_tools/tools/build_defs/repo/git.bzl:195:18):
 - <builtin>
 - /Volumes/KICKSTART/esp-v2/bazel/bazel_rules_python.bzl:22:9
 - /Volumes/KICKSTART/esp-v2/bazel/repositories.bzl:22:5
 - /Volumes/KICKSTART/esp-v2/WORKSPACE:91:1
ERROR: /Volumes/KICKSTART/esp-v2/tests/endpoints/grpc_interop/proto/BUILD:1:1: at least one argument must be given to the 'package' function
ERROR: /Volumes/KICKSTART/esp-v2/tests/env/testdata/BUILD:1:1: at least one argument must be given to the 'package' function
ERROR: /Volumes/KICKSTART/esp-v2/examples/BUILD:1:1: at least one argument must be given to the 'package' function
ERROR: package contains errors: tests/env/testdata
ERROR: package contains errors: tests/endpoints/grpc_interop/proto
ERROR: package contains errors: examples
ERROR: error loading package 'examples': Package 'examples' contains errors
Loading: 28 packages loaded

```

This, after running `bazelisk query //...`. Any ideas?

sgammon:
Additionally, when building (rather than querying), I get the following error:
```
(01:40:55) INFO: From Generating Descriptor Set proto_library @envoy_api//envoy/config/trace/v3:pkg:
envoy/config/trace/v3/opencensus.proto:12:1: warning: Import validate/validate.proto but not used.
envoy/config/trace/v3/trace.proto:5:1: warning: Import udpa/annotations/status.proto but not used.
[...]

PR #270, opened 2020-08-11
gcloud_build_image: cd back to original directory

nareddyt:
- `cd` back after deleting.
- Don't run inside `set -x`, too verbose.
- Have e2e test check working directory.

### DIFF
--- a/docker/serverless/gcloud_build_image
+++ b/docker/serverless/gcloud_build_image
@@ -100,8 +100,9 @@ EOF
 
 NEW_IMAGE="gcr.io/${PROJECT}/endpoints-runtime-serverless:${ESP_FULL_VERSION}-${SERVICE}-${CONFIG_ID}"
 gcloud builds submit --tag "${NEW_IMAGE}" . --project="${PROJECT}"
-
-# Delete the temporary directory we created earlier.
-rm -r "${PWD}"
 )
 
+# Delete the temporary directory we created earlier.
+# Move back to the previous directory with an echo.
+rm -r "${PWD}"
+cd ~-
--- a/prow/e2e-gcloud-build-image.sh
+++ b/prow/e2e-gcloud-build-image.sh
@@ -29,6 +29,7 @@ export CLOUDSDK_CORE_DISABLE_PROMPTS=1
 # Use a service that is not used for any other tests.
 SERVICE_NAME="gcloud-build-image-e2e-test.endpoints.cloudesf-testing.cloud.goog"
 CONFIG_ID="2020-07-20r1"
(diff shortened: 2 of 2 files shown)


nareddyt:
cc @EatonZ, just a few minor improvements over #257

docker/serverless/gcloud_build_image
```diff
@@ -100,8 +100,9 @@ EOF
...
 NEW_IMAGE="gcr.io/${PROJECT}/endpoints-runtime-serverless:${ESP_FULL_VERSION}-${SERVICE}-${CONFIG_ID}"
 gcloud builds submit --tag "${NEW_IMAGE}" . --project="${PROJECT}"
+)
 
 # Delete the temporary directory we created earlier.
+# Move back to the previous directory with an echo.
 rm -r "${PWD}"
-)
-
+cd ~-
```
qiwzhang, line 108: Hmm,  if we enlarge () to: before line 77 to after line 108,   we don't need such "cd" back.   I believe () scope will restore to original folder, is that right

nareddyt: Yes, but then you end up with extra output at the end when you `set +x` before `rm -f`

```
c2f9790b-3d85-4418-b086-2b324a9e53f0  2020-08-11T17:03:24+00:00  1M9S      gs://nareddyt-cloud-esf_cloudbuild/source/1597165403.568198-c15932054a884012813eb22745228e82.tgz  gcr.io/nareddyt-cloud-esf/endpoints-runtime-serverless:2.15.0-espv2-path-matcher-lrnm3ejpeq-uc.a.run.app-2020-05-13r1  SUCCESS
+ set +x
```

Preferably I'd like the last output to be the image name, nothing else.

qiwzhang: then, can we use pushd and popd?

nareddyt: Worried about portability. Most articles I've read online mention that `cd -` is preferred over pushd/popd.

https://mamchenkov.net/wordpress/2017/03/22/pushdpopd-vs-cd/

> pushd and popd, it turns out, are built into Bash, but they are not standard POSIX commands, so not all the shells have them. pushd and popd, it turns out, are built into Bash, but they are not standard POSIX commands, so not all the shells have them. 

Given we'd had issues with this script on macos before (#258), i'd prefer to use `cd -`.

qiwzhang: OK,  thanks.

nareddyt:
Ignored flakey tests that will be fixed by #271

PR #271, opened 2020-08-11
Improve quota e2e test

TAOXUY:
- For [ssl read timeout exception](https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/270/ESPv2-cloud-run-e2e-cloud-run-http-bookstore/1293227326766583809), add try catch. Since we are not testing ssl for this test case, ok to ignore.
- For [code count out of bound](https://prow.k8s.io/view/gcs/oss-prow/logs/ESPv2-continuous-cloud-run-e2e-app-engine-http-bookstore/1292793162674212865), increase the margin(admit 90% is wide but it is OK if we only aim to test if the rate limit works ).

### DIFF
--- a/tests/e2e/client/apiproxy_bookstore_quota_test.py
+++ b/tests/e2e/client/apiproxy_bookstore_quota_test.py
@@ -45,29 +45,30 @@ class ApiProxyBookstoreTest(ApiProxyClientTest):
     FLAGS.verbose = False
     self.set_verbose(FLAGS.verbose)
 
-    # exhaust the quota in the current window.
-    print("Exhaust current quota...");
-    response = self._call_http(path='/quota_read',
-                               api_key=FLAGS.api_key)
-    if response.status_code != 429:
-      for i in range(10000):
-        response = self._call_http(path='/quota_read',
-                                   api_key=FLAGS.api_key)
+
+    def _exhaust_quota():
+      for i in range(100):
+        time.sleep(1)
+        try:
+          response = self._call_http(path='/quota_read',
+                                     api_key=FLAGS.api_key)
+        except Exception, e:
+          print "Exception {0} occurred".format(e)
+          continue
         if response.status_code == 429:
           break;
-        elif i == 9999:
+        elif i == 99:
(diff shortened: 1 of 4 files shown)


tests/e2e/client/apiproxy_bookstore_quota_test.py
```diff
@@ -45,29 +45,30 @@ def verify_quota_control(self):
...
+          print "Exception {0} occurred".format(e)
+          continue
         if response.status_code == 429:
           break;
-        elif i == 9999:
+        elif i == 99:
           sys.exit(utils.red("Fail to exhaust quota"))
 
+    # exhaust the quota in the current window.
+    print("Exhaust current quota...")
```
qiwzhang, line 64: Could you add the comment on the algorithm you are using?  Like to know what are the purposes of these
 the exhaust_quota() and sleep(5) calls

TAOXUY: Done.

TAOXUY:
PTAL

TAOXUY:
PTAL

TAOXUY:
PTAL

tests/e2e/client/apiproxy_bookstore_quota_test.py
```diff
@@ -45,29 +46,34 @@ def verify_quota_control(self):
     FLAGS.verbose = False
     self.set_verbose(FLAGS.verbose)
 
+
+    def _try_call_quota_read():
+      try:
+        return self._call_http(path='/quota_read',
+                                   api_key=FLAGS.api_key)
+      except ssl.SSLError as e:
```
nareddyt, line 54: Why do only the quota requests get this exception? Other requests don't?

TAOXUY: I remember I have seen this before and other tests using same util lib doesn't call same api so frequently.

TAOXUY:
PTAL. The quota test are pretty flaky now.

PR #273, opened 2020-08-12
Managed service rollout during e2e load test

nareddyt:
Let's test rebuilding envoy's listener while there traffic going through. This may help us catch any problems with the teardown of our custom filters.

For now, only runs during our bookstore e2e test. Redeploys the exact same service config to force a new rollout / config id.


nareddyt:
PTAL @TAOXUY

tests/e2e/scripts/gke/deploy.sh
```diff
@@ -121,6 +122,40 @@ fi
...
+    else
+      allow_cors=false
+    fi
+
+    echo "doServiceRollout: Setting allowCors = ${allow_cors}"
+    local tmp_file=$(mktemp)
+    cat "${SERVICE_IDL}" \
+        | jq ".\"x-google-endpoints\"[0].allowCors = $allow_cors" \
+        > "${tmp_file}"
+    mv -f "${tmp_file}" "${SERVICE_IDL}"
```
qiwzhang, line 150: not need to change service config.  We don't have anything to compare them, everytime if there is a "deploy",  a new config_id and rollout_id is generated.

tests/e2e/scripts/gke/deploy.sh
```diff
@@ -121,6 +122,40 @@ fi
...
+# This will cause ESPv2 to rebuild the Envoy listener while lots of traffic is running through.
+function doServiceRollout() {
+  if [ "${BACKEND}" != 'bookstore' ]; then
+    echo "doServiceRollout: TODO(nareddyt): Support managed service rollout for ${BACKEND}"
+    return 0
+  fi
+
+  while true; do
+    echo 'doServiceRollout: Sleeping until next service rollout'
+    sleep 15m
```
TAOXUY, line 136: If this test only cares about crash, is the period too long? 
I checked the running part of all tests are less than 30mins so the config will be updated once.

tests/e2e/client/apiproxy_bookstore_test.py
```diff
@@ -188,6 +188,20 @@ def verify_jwt_locations(self):
...
         self.assertEqual(response.status_code, 200)
 
+    def run_options_queries(self):
+        # turn off verbose log
+        verbose = FLAGS.verbose
+        FLAGS.verbose = False
+        self.set_verbose(False)
+
+        print('Running CORS requests, no verifications')
+        for _ in range(100):
```
TAOXUY, line 198: Can we have accuracy test? A random idea is like repeatedly calling this `OPTIONS /shelves` until the new service config is applied.

tests/e2e/scripts/gke/deploy.sh
```diff
@@ -121,6 +122,40 @@ fi
...
+        | jq ".\"x-google-endpoints\"[0].allowCors = $allow_cors" \
+        > "${tmp_file}"
+    mv -f "${tmp_file}" "${SERVICE_IDL}"
+
+    echo "doServiceRollout: Deploying and rolling out new config for service ${APIPROXY_SERVICE}"
+    create_service ${CREATE_SERVICE_ARGS}
+  done
+}
+
+doServiceRollout &
```
TAOXUY, line 157: Is it supposed by wrapped by checking "bookstore" backend?

tests/e2e/scripts/gke/deploy.sh
nareddyt: For our presubmits, yet. But we mostly care about this during our long run tests. Then it will be updated many times.

If we update it more often than 15 minutes, we may have conflicts between different test runs. They all use the same service.

tests/e2e/client/apiproxy_bookstore_test.py
nareddyt: Prefer not to test accuracy, as we can have 2 tests running at the same time. They use the same service.

tests/e2e/scripts/gke/deploy.sh
nareddyt: Hmm, is there a good reason why?

Keep in mind config manager will take up to 1 minute to detect the new rollout, so checking right after a deploy will do nothing.

TAOXUY: I saw the function only modified the bookstore openapi yaml?
Also, why you only kill it if it is bookstore in line 177?

tests/e2e/scripts/gke/deploy.sh
TAOXUY: I see.

tests/e2e/scripts/gke/deploy.sh
nareddyt: Agreed as mentioned in PR description. I guess I should clarify, I'm trying to kill 2 birds with 1 stone in this PR:
1. Do a service rollout while traffic is running
2. Test SC Reports for `allowCors`. We generate custom operation names for the CORS methods we generate. I want to verify they show up in the Endpoints Console, and if they'll be formatted correctly.

I can remove the modifying `allowCors` every time before deploy, but I find it helpful for debugging. Then I can look at the prow job logs and verify that Config Manager picked up the new rollout. Otherwise, we have no logging in our prow job to verify this, we have to look at the container logs.

tests/e2e/client/apiproxy_bookstore_test.py
TAOXUY: You can still pass flag to differentiate long run jobs and presubmit jobs so that no overwritting in the same time.

These queries don't make sense to me if no any verification.

tests/e2e/scripts/gke/deploy.sh
qiwzhang: I see.   Or if we can find a string field in openapi spec that we can modify for each change. 

We should be able to get the config_id It is using from envoy admin port or from config_mgr, verify the new config id is picked up correctly.

tests/e2e/scripts/gke/deploy.sh
nareddyt: Done, moved the if statement to clarify.

tests/e2e/client/apiproxy_bookstore_test.py
nareddyt: I thought about this for some time. Even if we limit it to only periodic tests, it's not easy. 

The background process makes a rollout, but ESPv2 may take up to 2 minutes to pick up this new rollout and serve traffic on it. How can the background process communicate the expected value to the e2e client script? And how can we account for the potential 2 minute difference between a deploy and ESPv2 picking it up?

See [this comment](https://github.com/GoogleCloudPlatform/esp-v2/pull/273#discussion_r470897756) for why I don't think verifications are needed. Our integration tests verify the new rollout is picked up. The requests in e2e here are just for debugging, not verification.

tests/e2e/scripts/gke/deploy.sh
nareddyt: It's hard to do a verification on the config id. We don't expose this anywhere easily accessible. Even in the envoy admin port, the only place we could view this is the `/config_dump`. This requires parsing the entire listener config to find the config id in the SC filter.

I think just having some client-facing behavior change is good enough. In this case, the response code changes when we make an OPTIONS request based on the current config:
- `allowCors: true` will let the backend response with 200s
- `allowCors: false` will cause ESPv2 to respond with 404

tests/e2e/client/apiproxy_bookstore_test.py
TAOXUY: ACK. The test is not easy to setup because the config update cannot be updated right before verification.

We still need to make sure if the new config is picked up for ESPv2 finally or as Wayne said, by something else(like configId) . 
We don't need to reason the loading time. 

One extra question: why call 100 times without sleep? I can image most likely the response will be same. On the other hand, there is possibility these calls hit the same allowCors which give all 200 or 404, like  one complete test using 30min.

nareddyt: Simplified the e2e test. Removed all cors related items, we don't need to verify anything.

tests/e2e/scripts/gke/deploy.sh
nareddyt: SImplified the test, remove all the cors related items. No need for verification, our integration test do this. We just want to see if there are any crashes.

PR #275, opened 2020-08-13
Add allowCors to a static bootstrap test

nareddyt:
Was curious how our Envoy config will look when `x-google-allow: all` and `allowCors: true` is turned on at the same time.

Other than the current bug with supporting `x-google-allow: all` with backend routing, config looks good. Let's keep it as verification.

### DIFF
--- a/examples/testdata/path_matcher/openapi_swagger.json
+++ b/examples/testdata/path_matcher/openapi_swagger.json
@@ -75,8 +75,8 @@
   "x-google-allow": "all",
   "x-google-endpoints": [
     {
-      "name": "examples-path-matcher.endpoints.cloudesf-testing.cloud.goog",
-      "allowCors": true
+      "allowCors": true,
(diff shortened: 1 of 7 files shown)


examples/testdata/path_matcher/openapi_swagger.json
```diff
@@ -72,5 +72,11 @@
     "https"
   ],
   "swagger": "2.0",
-  "x-google-allow": "all"
+  "x-google-allow": "all",
+  "x-google-endpoints": [
+    {
+      "allowCors": true,
```
qiwzhang, line 78: I can see: "allowCors" is supported, but "allow: all" is not supported. 

In the Envoy route match,  we have a route match for "option" for /libraries/{library_id}/shelves/{shelf_id}
we don't have any route match for /** for any methods.

nareddyt: Yup, we OpenAPI compiler doesn't generate backend rules for `x-google-allow: all`. So it's not supported for dynamic routing.

Issue #274, opened 2020-08-13
Not to bailed container to cause it to re-start for 429 error when calling service management 


qiwzhang:
This is from an [issue](https://github.com/cloudendpoints/esp/issues/811) at ESPv1.  But I think it applies to ESPv2 too.

PR #278, opened 2020-08-14
b/164477080: Add retry mechanism on callgoogleapis when it comes into 429

TAOXUY:
This is for https://github.com/GoogleCloudPlatform/esp-v2/issues/274. Considering quota is refilled by every min, add retry of 3 sec x 30 times for 429 error.

This method will be used both in
- first time start configmanager
- poll latest config in managed mode

The case 1 is what we are fixing. For the case 2, its interval is 60s so there would be at most 2 goroutine retrying calling servicemanagement, which should be fine. 

FYI, current servicemanagement API quota is 120/min by default.

### DIFF
--- a/src/go/configmanager/flags/flags.go
+++ b/src/go/configmanager/flags/flags.go
@@ -109,7 +109,6 @@ var (
 	// Flags for testing purpose.
 	SkipJwtAuthnFilter       = flag.Bool("skip_jwt_authn_filter", false, "skip jwt authn filter, for test purpose")
 	SkipServiceControlFilter = flag.Bool("skip_service_control_filter", false, "skip service control filter, for test purpose")
-	SmCallRetryConfigs       = flag.String("service_management_call_retry_configs", "", `Set retry times and retry intervals for different status calling service management by json string. For example, setting retry 30 times with 3sec interval for status 429 is {"429":{"RetryNum":30,"RetryInterval":3000000000}}. This flag is used for test only`)
 
 	TranscodingAlwaysPrintPrimitiveFields   = flag.Bool("transcoding_always_print_primitive_fields", false, "Whether to always print primitive fields for grpc-json transcoding")
 	TranscodingAlwaysPrintEnumsAsInts       = flag.Bool("transcoding_always_print_enums_as_ints", false, "Whether to always print enums as ints for grpc-json transcoding")
@@ -150,7 +149,6 @@ func EnvoyConfigOptionsFromFlags() options.ConfigGeneratorOptions {
 		TokenAgentPort:                          *TokenAgentPort,
 		SkipJwtAuthnFilter:                      *SkipJwtAuthnFilter,
 		SkipServiceControlFilter:                *SkipServiceControlFilter,
-		SmCallRetryConfigs:                      *SmCallRetryConfigs,
 		EnvoyUseRemoteAddress:                   *EnvoyUseRemoteAddress,
 		EnvoyXffNumTrustedHops:                  *EnvoyXffNumTrustedHops,
 		LogJwtPayloads:                          *LogJwtPayloads,
--- a/src/go/configmanager/flags/flags.go
+++ b/src/go/configmanager/flags/flags.go
@@ -109,6 +109,7 @@ var (
 	// Flags for testing purpose.
 	SkipJwtAuthnFilter       = flag.Bool("skip_jwt_authn_filter", false, "skip jwt authn filter, for test purpose")
 	SkipServiceControlFilter = flag.Bool("skip_service_control_filter", false, "skip service control filter, for test purpose")
+	SmCallRetryConfigs       = flag.String("service_management_call_retry_configs", "", `Set retry times and retry intervals for different status calling service management. For example, setting retry 30 times with 3sec interval for status 429 is {"429":{"RetryNum":30,"RetryInterval":3000000000}}. This flag is used for test only`)
 
 	TranscodingAlwaysPrintPrimitiveFields   = flag.Bool("transcoding_always_print_primitive_fields", false, "Whether to always print primitive fields for grpc-json transcoding")
 	TranscodingAlwaysPrintEnumsAsInts       = flag.Bool("transcoding_always_print_enums_as_ints", false, "Whether to always print enums as ints for grpc-json transcoding")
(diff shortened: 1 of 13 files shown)


tests/integration_test/managed_service_config_test.go
```diff
@@ -102,3 +105,53 @@ func TestManagedServiceConfig(t *testing.T) {
...
+		t.Fatalf("fail to setup test env, %v", err)
+	}
+
+	url := fmt.Sprintf("http://localhost:%v/echo", s.Ports().ListenerPort)
+
+	resp, err := echoClient.DoPost(fmt.Sprintf("%s?key=api-key", url), echo)
+	if err != nil {
+		t.Errorf("got unexpected error: %v", err)
+	}
+	wantResp := `{"message":"hello"}`
```
qiwzhang, line 153: can we have two test cases, one fail and one success?

src/go/util/request.go
```diff
@@ -18,39 +18,58 @@ import (
...
 	"io/ioutil"
 	"net/http"
+	"time"
 
+	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 )
 
-func callWithAccessToken(client *http.Client, path, method, token string) ([]byte, error) {
+var Retries = 30
```
qiwzhang, line 27: Can we implement as generic OnStatusRetry.  At first, status list can be just 429,  later, we can add other status. 

Config may look like this map:
OnStatusRetryConfig:
{
   [429, 433] => {RetryNum: 30,  RetryInterval: 3}
   [500] =>   {RetryNum: 5,  RetryInterval: 1}
}

src/go/util/request.go
```diff
@@ -18,39 +18,58 @@ import (
...
 	if err != nil {
 		return fmt.Errorf("fail to get access token: %v", err)
 	}
 
-	respBytes, err := callWithAccessToken(client, path, method, token)
+	var respBytes []byte
+	var statusCode int
+
+	// Retry calls for code 429 Too Many Requests.
+	for i := 0; i < Retries; i++ {
```
qiwzhang, line 63: I am not sure if this code only used for calling ServieManagement,  or also for ServiceControl. 
ServiceControl should not have quota limit, so retry policy should be different.

src/go/util/request.go
qiwzhang: Hmm.  may need to have different retry policy with different host,  so the config map can be

host ->  []status -> {retry_num, retry_interval}

src/go/util/request.go
TAOXUY: Yes. It is also used for servicecontrol and its quota limit is super high, so adding this won't effect calling servicecontrol.

src/go/util/request.go
TAOXUY: Done.

src/go/util/request.go
TAOXUY: For new version, calling service control is not effected at all.

src/go/configmanager/flags/flags.go
```diff
@@ -109,6 +109,7 @@ var (
 	// Flags for testing purpose.
 	SkipJwtAuthnFilter       = flag.Bool("skip_jwt_authn_filter", false, "skip jwt authn filter, for test purpose")
 	SkipServiceControlFilter = flag.Bool("skip_service_control_filter", false, "skip service control filter, for test purpose")
+	SmCallRetryConfigs       = flag.String("service_management_call_retry_configs", "", `Set retry times and retry intervals for different status calling service management by json string. For example, setting retry 30 times with 3sec interval for status 429 is {"429":{"RetryNum":30,"RetryInterval":3000000000}}. This flag is used for test only`)
```
nareddyt, line 112: Do we need to expose a flag for this? Can't we unit test with just the option? Not a fan of passing a JSON object via a flag.

src/go/util/request.go
```diff
@@ -18,39 +18,67 @@ import (
...
+		respBytes, statusCode, err = callWithAccessToken(client, path, method, token)
+		if retryConfigs == nil {
+			break
+		} else if retryConfig, ok := retryConfigs[statusCode]; !ok {
+			break
+		} else if retryConfig.RetryNum <= callStatusCnts[statusCode] {
+			break
+		} else {
+			callStatusCnts[statusCode] += 1
+			glog.Warningf("after %v times failures on status %v, retrying http call %s with %v remaining chances", callStatusCnts[statusCode], statusCode, path, retryConfig.RetryNum-callStatusCnts[statusCode])
```
nareddyt, line 76: Remove "times"

src/go/util/request_test.go
```diff
@@ -39,30 +42,45 @@ func initServerForTestCallWithAccessToken(t *testing.T, desc, wantMethod, wantTo
 			t.Errorf("test(%v) fail, want Content-Type: application/x-protobuf, get Content-Type: %s", desc, got)
 		}
 
+		if respStatusCode != 0 && rejectCnt < rejectTimes {
```
nareddyt, line 45: Suggest you re-use our generic retryHandler instead of writing your own code: https://github.com/GoogleCloudPlatform/esp-v2/blob/22f9d24a343b23176e350cf08e65f06dc4f5c4fd/tests/env/components/retry_handler.go#L41

Reduces code duplication.

src/go/configmanager/flags/flags.go
qiwzhang: Yeap, I don't like to expose such flag either.   We can just define a global map, that test code can modify it.

src/go/util/request_test.go
TAOXUY: Prefer to keep the current version as it is a very simple test server, which is shared by all the testscase here and using this generic server also introduce other codes to config it.

src/go/configmanager/flags/flags.go
TAOXUY: Done. The integration test won't have negative test.

src/go/serviceconfig/service_config_fetcher.go
```diff
@@ -29,23 +30,33 @@ type ServiceConfigFetcher struct {
 	serviceName          string
 	client               *http.Client
 	accessToken          util.GetAccessTokenFunc
+	retryConfigs         map[int]util.RetryConfig
+}
+
+var SmRetryConfigs = map[int]util.RetryConfig{
+	http.StatusTooManyRequests: util.RetryConfig{
+		RetryNum:      30,
+		RetryInterval: time.Second * 3,
```
qiwzhang, line 39: How about 30 retry and retry interval 10s for 5 minutes

tests/integration_test/managed_service_config_test.go
qiwzhang: why could n't we have a negative test case?

TAOXUY: Because integration test use the binary and we cannot modify global map, we have to wait 1min30sec to see the failure.

src/go/serviceconfig/service_config_fetcher.go
TAOXUY: Done

TAOXUY:
Add unit test in configmanager_test, which is relatively close to integration test.
PTAL

src/go/serviceconfig/service_config_fetcher.go
TAOXUY: Done

qiwzhang: I did not see the change

PR #283, opened 2020-08-17
b/156286299: GCS Runner: Enable statsd with --service-cluster and --service-node flags

toddbeckman-g:
I'm not entirely sure why these flags are required for StatsdSink, but they are. Without these flags, StatsdSink silently never starts up.


nareddyt:
Let's merge in tomorrow after the 2.16 release.

PR #287, opened 2020-08-18
b/165086608: Use dedicated SA Key for JWT e2e tests

nareddyt:
- Reduce the chance of leaking the SA Key: e2e tests download the key into the `/tmp` folder instead of the repository. This will prevent accidentally committing the file.
- Reduce the risk of malicious use if a leak occurs: Current tests use the default compute developer account. This has broad roles. Instead, create a dedicated service account with no roles at all.

### DIFF
--- a/examples/grpc_dynamic_routing/envoy_config.json
+++ b/examples/grpc_dynamic_routing/envoy_config.json
@@ -449,14 +449,14 @@
                             "fromParams": [
                               "access_token"
                             ],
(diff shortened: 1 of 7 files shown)


nareddyt:
Job was successful, just failed due to unrelated periodic tests overwriting the config

```
=== In Total
Complete requests: 157874,
Failed requests: 0,
Non-2xx responses: 0
```

PR #288, opened 2020-08-19
Update README.md and use cases

nareddyt:
We now support sidecar deployment mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,7 @@ Google Cloud Endpoints and ESPv2 provide:
 
 * **Easy Adoption**: The API service can be implemented in any coding language using any IDLs.
 
-* **Multiple Deployment Modes**: Deploy ESPv2 as an API Gateway.
-Support for ESPv2 as a sidecar is planned.
+* **Multiple Deployment Modes**: Deploy ESPv2 as an API Gateway or a sidecar.
 
 * **Platform Flexibility**: Support the deployment on any cloud or hybrid environment.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,8 +28,7 @@ Google Cloud Endpoints and ESPv2 provide:
 
 * **Easy Adoption**: The API service can be implemented in any coding language using any IDLs.
 
-* **Multiple Deployment Modes**: Deploy ESPv2 as an API Gateway.
-Support for ESPv2 as a sidecar is planned.
+* **Multiple Deployment Modes**: Deploy ESPv2 as an API Gateway or a sidecar.
```
TAOXUY, line 31: gateway should be lowercase?

qiwzhang: I feel it is better to call it "Remote proxy" deployment mode. "API Gateway" could be confused with "API gateway" product. 
We also need to add some description for these two modes:
"Remote proxy" mode is deploying ESPv2 as Cloud Run service as remote proxy to support applications deployed serverless platforms, such as ....
"Side-car" mode is deploying ESPv2 container in the same host as application,  designed for GKE, GCE and k8s

TAOXUY: +1 on confusion. 

How about only keep gateway here no API? So it is just `a gateway or a sidecar`(they are well-known terminology).

qiwzhang: How about "remote gateway"?

nareddyt: Here is the new terminology: Deploy ESPv2 as a **sidecar** for local backends or as a **reverse proxy** for remote backends.

Decided not to use the word gateway anywhere.

I added description of these 2 modes later on in the readme and in our use-cases doc. Don't want to make the beginning of the readme have too much detail.

PR #292, opened 2020-08-21
Add API regression presubmit test

TAOXUY:
The main purpose of this test is to detect API backward incompatible change.  For instance, it can detect the tailing / was removed in #251.

Keep in mind, for any change that modify the existing behavior checked in integration test(mostly like bug fix), it will also break like #295, which changes the report behavior(not always set APIKey in customer_id in report). We can expect there are only few of them. For these changes, this presubmit test will fail with part of broken integration tests.


nareddyt:
@TAOXUY: #295 is an example where the proposed API Compatibility test strategy will fail. 

Even though the APIs are backwards compatible, I made some bug fixes in the report behavior. So now when you check out an older commit and run integration tests, the tests will definitely fail.

Is there any plan to filter out such errors? (I don't think it's easy). Do we just need to be aware of this and manually decide instead?

TAOXUY:
> @TAOXUY: #295 is an example where the proposed API Compatibility test strategy will fail.
> 
> Even though the APIs are backwards compatible, I made some bug fixes in the report behavior. So now when you check out an older commit and run integration tests, the tests will definitely fail.
> 
> Is there any plan to filter out such errors? (I don't think it's easy). Do we just need to be aware of this and manually decide instead?

For #295, it won't. Basically, there are three parts: envoy, configmanger, integration test. Now we are using envoy(latest), configmanager(last release) and integration test(last release). For #295, envoy adds more information of apiKey and its validation and they are transparent to old integration test. 

The only consideration can be #299 and #296, where SC report missed information and it is checked in integration tests. They are backward incompatible in some point and we can expect there are few. For them, we can skip part of go integration test before we do next release.

prow/gcpproxy-presubmit.sh
```diff
@@ -25,83 +25,31 @@ pwd)
...
-    make test-envoy-msan
-    ;;
-  "tsan")
-    make test-envoy-tsan
-    ;;
-  *)
-    make test-envoy
-    ;;
-esac
+SHA=$(git rev-list -n 1 $(git describe --abbrev=0))
```
qiwzhang, line 46: Are you directly change our presubmit test to be api-compatibilty test?   I hope not.  
It should be a separate test that we can ignore for some PR

TAOXUY: Of course not. I just use gcpproxy-presubmit as a test trigger before we config the prow.

nareddyt:
> For #295, envoy adds more information of apiKey and its validation and they are transparent to old integration test.

Not really, we add/remove information based on API Key validity. They will fail the old integration tests.

Maybe this is too general, but any bug fix that _modifies_ an integration test will most likely fail when run against older ones.  I guess that's OK, we won't have too many bug fixes like that.

> we can skip part of go integration test before we do next release.

Do you plan to make this part of the release process? I was under the assumption it would just run for every PR, each PR author needs to consider it. Not the release oncall.

TAOXUY:
> > For #295, envoy adds more information of apiKey and its validation and they are transparent to old integration test.
> 
> Not really, we add/remove information based on API Key validity. They will fail the old integration tests.
> 
> Maybe this is too general, but any bug fix that _modifies_ an integration test will most likely fail when run against older ones. I guess that's OK, we won't have too many bug fixes like that.
> 
> > we can skip part of go integration test before we do next release.
> 
> Do you plan to make this part of the release process? I was under the assumption it would just run for every PR, each PR author needs to consider it. Not the release oncall.

For #1, I see. #295 also fix the bug of apiKey in report. Agree, modifying the existing integration test check is rare.

For #2, not release process and it is just a presubmit. I mean the integration test change won't be in until next release.

TAOXUY:
PTAL

prow/gcpproxy-api-regression.sh
```diff
@@ -0,0 +1,65 @@
...
+echo '================ Download latest envoy ================'
+echo '======================================================='
+download_envoy_binary
+chmod +x ${ROOT}/bin/envoy
+
+# keep the current version
+VERSION=$(cat ${ROOT}/VERSION)
+
+# checkout to last releasd commit
+SHA=$(git rev-list -n 1 $(git describe --abbrev=0))
```
nareddyt, line 53: Hmm, why checkout to a released commit? Why not just checkout to the latest commit on master (the diff would be all the commits in the individual PR).

The problem with checking out to a released commit is: As soon as we submit one PR that breaks the test on purpose, then all the following PRs will keep failing until a new release is made. That doesn't sound very helpful then.

Also if we increment the API version, every single PR afterwords will fail until we make a new release.

Better to just test against master. Each PR will be responsible for making sure they don't break the API version. If the previous PR didn't break the version, then no need to checkout to a older commit than it.

TAOXUY: Ack, good point.

TAOXUY: Done.

PR #293, opened 2020-08-25
b/166168739: Reduce noise of Envoy logs with `--enable_debug`

nareddyt:
DNS resolutions and stats logging do not add much information, and they keep logging automatically every few seconds. Disable them.

DNS are debug logs in the `upstream` component, stats are debug logs in the `main` component. [Screenshot](https://screenshot.googleplex.com/85850ba9-6083-4247-be95-3947fd715cde)

Testing done:
- Manually deployed this image with `--enable_debug` and verified noise is reduced a lot.
- No need for automated tests to verify this. We will rarely change these logging options.


docker/generic/start_proxy.py
```diff
@@ -955,7 +955,7 @@ def gen_envoy_args(args):
            "--log-format-escaped"]
 
     if args.enable_debug:
-        cmd.append("-l debug")
+        cmd.append("-l debug --component-log-level upstream:info,main:info")
```
qiwzhang: Can we just exclude DNS,  I think upstream and main have other useful debug info?  e.g. main has envoy config info.
upstream connection info is important too.

nareddyt: Unfortunately DNS are debug logs in the `upstream` component, stats are debug logs in the `main` component: https://screenshot.googleplex.com/85850ba9-6083-4247-be95-3947fd715cde

I did some analysis, this shows the other logs that we'd lose with this PR: https://screenshot.googleplex.com/AqbNqAavU5gmkQd

Do you think this are useful to have? I don't see the value these ones add.

If you think we need those logs as well, then I'll revert this PR. Seems there is no other way to filter out the noise other than by component.

qiwzhang: OK, then

PR #295, opened 2020-08-25
b/166158386 #1: Fix error with reporting invalid API Keys

nareddyt:
Implements the following:
- If the API Key State if verified by CHECK: Report in operation and log entry
- If the API Key is invalid, Consumer project is not enabled, or CHECK is not called: Report in log entry only, do NOT populate `credential_id`, `consumer_id`, etc.
- If the API Key is missing, preserve the current behavior: API Key not anywhere in report.
- If network error occurs (including network fail open): API Key is treated as untrusted, is only in log entry.

Testing done:
- Unit tests
- Integration tests - mostly modifications to validate behavior, added a couple new cases.
- Manual e2e test - confirmed that requests with invalid api keys now show up in the endpoints logs.

Follow-up PRs:
- Add the API Key State into the log entries.


src/api_proxy/service_control/request_info.h
```diff
@@ -45,6 +45,28 @@ inline const char* ToString(Protocol p) {
 
 }  // namespace protocol
 
+namespace identity {
+
+// Indicates whether the API key is verified by SC Check.
+enum ApiConsumerIdentity {
```
qiwzhang, line 51: I still prefer ApiKeyStatus.   Could you elaborate on the name conflicting issue?

src/api_proxy/service_control/request_builder.cc
```diff
@@ -1276,6 +1292,12 @@ Status RequestBuilder::AppendByConsumerOperations(
     Timestamp current_time) const {
   Operation* op = request->add_operations();
   SetOperationCommonFields(info, current_time, op);
+  if (info.api_consumer_identity == identity::ApiConsumerIdentity::VERIFIED) {
```
qiwzhang, line 1295: We need to add "set_consumer_id" for FillAllocateQuotaRequest

nareddyt: We do that here: https://github.com/GoogleCloudPlatform/esp-v2/pull/295/files/111cf2d08002a6d881b628c951c60bf592de2591#diff-cbee4bff4243945b26c186374c3889fdR1141

We allow any API key, doesn't have to be verified valid.

src/api_proxy/service_control/request_info.h
nareddyt: Because the API Key can be valid, but the consumer service may not have activated the service. In that case, we should not fill in the credential id. It would be conflicting to say `api_key_status: invalid` when the API Key is actually valid.

qiwzhang: How about adding a new enum state: NOT_CHECKED,  INVALID,  VALID, NOT_ENABLED.  
Can we still call it ApiKeyStatus or State?

src/api_proxy/service_control/request_builder.cc
qiwzhang: I see. I missed that. Thanks

src/api_proxy/service_control/request_info.h
nareddyt: Sounds good, I'll use `ApiKeyState`

nareddyt:
> I feel this logic should be part of request_builder ConvertCheckReponse code.
> How about add ApiKeyState into CheckResponseInfo?

@qiwzhang I thought about this, but then we'd have duplicate information in the CheckResponseInfo. We'd have both the ApiKeyState alone with the CheckError.

And anyways, we need to handle the `NOT_CHECKED` status in handler_impl, because check is not called.

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -143,11 +145,20 @@ void ServiceControlHandlerImpl::prepareReportRequest(
...
   fillOperationInfo(info);
 
-  // Report: not to send api-key if invalid or service is not enabled.
-  if (check_response_info_.error_type == ScResponseErrorType::API_KEY_INVALID ||
-      check_response_info_.error_type ==
-          ScResponseErrorType::SERVICE_NOT_ACTIVATED) {
-    info.api_key.clear();
+  // This logic determines if consumer metric is sent in report based on the
+  // validity of the API Key.
+  if (on_check_done_called_) {
```
qiwzhang, line 150: Did not see where on_check_done_called_ is set to true.

nareddyt: It is pre-existing code: https://github.com/GoogleCloudPlatform/esp-v2/blob/ec6c83737213c4ee8401c402a3e0f2984df577ba/src/envoy/http/service_control/handler_impl.cc#L210

I explicitly initialize it in the constructor for safety. Then use it to determine if check was called.

src/api_proxy/service_control/request_builder_test.cc
```diff
@@ -367,17 +422,56 @@ TEST_F(RequestBuilderTest, FillReportRequestEmptyOptionalTest) {
...
+TEST_F(RequestBuilderTest, CredentailIdApiKeyInvalidTest) {
+  ReportRequestInfo info;
+  FillOperationInfo(&info);
+
+  info.api_key_state = api_key::ApiKeyState::INVALID;
+
+  gasv1::ReportRequest request;
+  ASSERT_TRUE(scp_.FillReportRequest(info, &request).ok());
+
+  ASSERT_FALSE(request.operations(0).labels().contains("/credential_id"));
```
qiwzhang, line 448: Can we verify all 3 fields:  consumer_id, credentail_id, and log_entry.  

Verify them for all api_key_states?

src/envoy/http/service_control/handler_impl.cc
qiwzhang: I see.  Thanks

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -143,11 +145,20 @@ void ServiceControlHandlerImpl::prepareReportRequest(
...
+    if (check_response_info_.error_type ==
+        ScResponseErrorType::API_KEY_INVALID) {
+      info.api_key_state = ApiKeyState::INVALID;
+    } else if (check_response_info_.error_type ==
+               ScResponseErrorType::SERVICE_NOT_ACTIVATED) {
+      info.api_key_state = ApiKeyState::NOT_ENABLED;
+    } else {
+      info.api_key_state = ApiKeyState::VERIFIED;
```
qiwzhang, line 158: How about Check returned with no-ok status?  network failure status?  I think the default check_respond_info_check_type is NOT_SPECIFIED.   We may have to handle that.

src/envoy/http/service_control/handler_impl_test.cc
```diff
@@ -421,11 +428,43 @@ TEST_F(HandlerTest, HandlerCheckNotNeeded) {
...
+  TestRequestHeaderMapImpl headers{
+      {":method", "GET"}, {":path", "/echo"}, {"x-api-key", "invalid-key"}};
+  TestResponseHeaderMapImpl response_headers{
+      {"content-type", "application/grpc"}};
+  ServiceControlHandlerImpl handler(headers, mock_stream_info_, "test-uuid",
+                                    *cfg_parser_, test_time_, stats_);
+
+  EXPECT_CALL(*mock_call_, callCheck(_, _, _)).Times(0);
+  EXPECT_CALL(*mock_call_, callQuota(_, _)).Times(0);
+  EXPECT_CALL(mock_check_done_callback_, onCheckDone(Status::OK));
```
qiwzhang, line 457: Can we add a case where Check get network_failure

tests/integration_test/service_control_basic_test.go
```diff
@@ -201,33 +201,34 @@ func TestServiceControlBasic(t *testing.T) {
...
-					Platform:                     util.GCE,
-					Location:                     "test-zone",
+					Version:         utils.ESPv2Version(),
+					ServiceName:     "echo-api.endpoints.cloudesf-testing.cloud.goog",
+					ServiceConfigID: "test-config-id",
+					URL:             "/echo/nokey?key=api-key",
+					ApiName:         "1.echo_api_endpoints_cloudesf_testing_cloud_goog",
+					ApiMethod:       "1.echo_api_endpoints_cloudesf_testing_cloud_goog.Echo_nokey",
+					// API Key is not verified by check, but still log it.
+					ApiKeyInLogEntryOnly: "api-key",
```
qiwzhang, line 218: I am wondering; will it be more readable if we rename them to 
ApiKeyVerified.   
ApiKeyNotVerified.

tests/utils/service_control_utils.go
```diff
@@ -663,7 +664,10 @@ func VerifyReport(body []byte, er *ExpectedReport) error {
 	}
 
 	if er.ApiKeyInOperationAndLogEntry != "" && er.ApiKeyInLogEntryOnly != "" {
-		return fmt.Errorf("cannot set both `ApiKeyInOperationAndLogEntry` and `ApiKeyInLogEntryOnly`")
+		return fmt.Errorf("cannot set both `ApiKeyInOperationAndLogEntry` and `ApiKeyInLogEntryOnly`, they imply different api key validity")
+	}
+	if er.ApiKeyInOperationAndLogEntry != "" && er.JwtAuth != "" {
```
qiwzhang, line 669: We can support both ApiKey and JwtAuth.  Why do we want to reject such case?

src/api_proxy/service_control/request_builder_test.cc
TAOXUY: +1

qiwzhang:
Hi @nareddyt After my pr https://github.com/GoogleCloudPlatform/esp-v2/pull/300 is merged.  you need to rebase to the HEAD

tests/utils/service_control_utils.go
nareddyt: Renamed to make it clearer. This is for setting `CredentialId` in operation. If API Key is valid, then it will be the credential id, so can't specify both.

tests/integration_test/service_control_basic_test.go
nareddyt: I disagree. In our test framework, we should be validating how/where the API Key appears in the report request. It changes based on if it's verified or not, but that is implementation detail.

If we rename it to Verified/NotVerified, it hides the fact that verified keys appear in the operation or not. It will be very easy to forget when adding new tests.

src/envoy/http/service_control/handler_impl_test.cc
nareddyt: Done in ClientCache tests

src/envoy/http/service_control/handler_impl.cc
nareddyt: Good catch. Ok, now the logic is getting too complicated for `HandlerImpl`. 

I agree with your point that we should do the conversion in `ClientCache`, I changed the code.

src/api_proxy/service_control/request_builder_test.cc
nareddyt: Done

src/api_proxy/service_control/request_builder_test.cc
```diff
@@ -305,10 +306,63 @@ TEST_F(RequestBuilderTest, FillReportRequestFailedTest) {
...
+  gasv1::ReportRequest request;
+  ASSERT_TRUE(scp_.FillReportRequest(info, &request).ok());
+
+  std::string text = ReportRequestToString(&request);
+  std::string expected_text =
+      ReadTestBaseline("report_request_invalid_api_consumer.golden");
+  ASSERT_EQ(expected_text, text);
+}
+
+TEST_F(RequestBuilderTest, FillReportWithNotEnabledApiKeyTest) {
```
qiwzhang, line 335: Maybe combine these two tests into one, with a loop to test all 3 states:  INVALID, NOT_ENABLED, NOT_CHECKED.

tests/integration_test/service_control_basic_test.go
qiwzhang: SG then.

tests/utils/service_control_utils.go
```diff
@@ -88,7 +88,7 @@ type ExpectedReport struct {
 	FrontendProtocol             string
 	BackendProtocol              string
 	Platform                     string
-	JwtAuth                      string
+	CredentialId                 string
```
qiwzhang, line 91: How about JwtAuthForCredentialId  or just JwtAuthCredentialId

src/envoy/http/service_control/client_cache_test.cc
```diff
@@ -141,7 +144,7 @@ TEST_F(ClientCacheCheckResponseTest, Sc5xxAllowed) {
   CheckError* check_error = response->mutable_check_errors()->Add();
   check_error->set_code(CheckError::NAMESPACE_LOOKUP_UNAVAILABLE);
 
-  runTest(Code::OK, response, Code::OK);
+  runTest(Code::OK, response, Code::OK, ApiKeyState::NOT_CHECKED);
```
qiwzhang, line 147: Why is is NOT_CHECKED?  

If there are not check_errors,  can we assume api-key is checked?  Can we assume Chemist check api-key first?

src/api_proxy/service_control/request_builder_test.cc
nareddyt: Done

tests/utils/service_control_utils.go
nareddyt: Sure, done

src/envoy/http/service_control/client_cache_test.cc
nareddyt: If there are no check errors and no http network errors, then yet. The current code treats it as VERIFIED.

nareddyt:
Cool, the api regression tests added in #292 caught the breaking change in this PR (as expected). Very useful @TAOXUY, thanks for adding it.

PR #299, opened 2020-08-27
Remove grpc_stats filter

qiwzhang:

As part of removing unused metrics,   streaming related metrics have been removed in Google.
grpc_message_count  are streaming metrics that to be removed.
grpc_stats fitler is only used to calculate message_count, will not be needed any more. 

This PR just removes the grpc_stats filter.

### DIFF
--- a/examples/grpc_dynamic_routing/envoy_config.json
+++ b/examples/grpc_dynamic_routing/envoy_config.json
@@ -1062,14 +1062,6 @@
                       }
                     },
                     {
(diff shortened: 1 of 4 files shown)


qiwzhang:
Envoy service_control code [here](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/src/envoy/http/service_control/handler_impl.cc#L335)
can handle grpc_stats not been used.  If the filter is not used, or it is not a grpc request,  it will not generate that particular filter_state, then we send msg_count as 0.

PR #302, opened 2020-08-28
b/166496629: Better naming of healthz and cors operations

nareddyt:
Just a general usability improvement, have all auto-generated methods follow the same format that OpenAPI compiler follows. OpenAPI compiler will [create methods](https://github.com/GoogleCloudPlatform/esp-v2/blob/0d21aff96635596a99f9fab31bf2b04edfc96052/examples/testdata/path_matcher/service_config_generated.json#L11) in the format `Google_Autogenerated_`. We should use `ESPv2_Autogenerated_`.

For `allowCors=true`, this matters because the CORS operations will show up in the Endpoints Console (as discovered via #290). Here is an example log entry: https://screenshot.googleplex.com/BwzDUzG7SSJMvEL 

For healthz, it doesn't really matter since they are not reported. But keep the format consistent.


src/go/configinfo/service_info.go
```diff
@@ -388,7 +388,10 @@ func (s *ServiceInfo) processHttpRule() error {
 
 	// Add HttpRule for HealthCheck method
 	if s.Options.Healthz != "" {
-		hcMethod, err := s.getOrCreateMethod("ESPv2.HealthCheck")
+		// FIXME: Health check should not associated with any API. But for consistency, just use the first API Name.
+		methodName := fmt.Sprintf("%s.%s_HealthCheck", s.ApiNames[0], util.AutogeneratedOperationPrefix)
```
TAOXUY, line 392: If we have multiple APIs, which is doable, look bad if we use the first one as prefix.

nareddyt: I see your point, but what should we use then? Anything else we use as `api_name` is technically invalid. If we ever support sending reports for healthz in the future, it will fail.

TAOXUY: We don't have plan of sending reports for health, do we?  If yes, maybe use this as workaround.

Also, the API name prefix makes it look like responding from the backend?

nareddyt: I see, yes that does imply it's coming from the backend. Ok, I'll change to some other api-name, even though it's invalid.

nareddyt: Is `espv2_deployment.ESPv2_Autogenerated_HealthCheck` clear?

PR #303, opened 2020-08-29
Improve API Regression test

TAOXUY:
- Move getting SHA from beginning otherwise the head of master can be updated when it is waiting envoy binary, [ex](https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/302/ESPv2-API-regression-test/1299488063969300480)
- Use Makefile target

### DIFF
--- a/prow/gcpproxy-api-regression.sh
+++ b/prow/gcpproxy-api-regression.sh
@@ -37,6 +37,8 @@ exit 1; }
 . ${ROOT}/tests/e2e/scripts/prow-utilities.sh || { echo 'Cannot load Bash utilities';
 exit 1; }
 
+# Get the SHA of the head of master
+SHA=$(git ls-remote https://github.com/GoogleCloudPlatform/esp-v2.git HEAD | cut -f 1)
 
 wait_apiproxy_image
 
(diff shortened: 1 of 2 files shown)


prow/gcpproxy-api-regression.sh
```diff
@@ -37,6 +37,8 @@ exit 1; }
 . ${ROOT}/tests/e2e/scripts/prow-utilities.sh || { echo 'Cannot load Bash utilities';
 exit 1; }
 
+# Get the SHA of the head of master
```
nareddyt, line 40: I think the problem with my PR is that I didn't merge in master. So the latest commit on master isn't on my branch, which is why the test fails.

How about this: Instead of checking out master in our upstream repo, why don't we checkout master in the local repository?

TAOXUY: Yes, that's it is trying to fix.

Actually, for a PR in CI on github, it is already on the master without no other branch. It doesn't know which commit is the original master head.

nareddyt: The problem is line 55, when we do `git checkout ${SHA}`. If the latest commit to master is not in my repo, then it won't be able to check it out. I don't think the ordering has anything to do with it. If I don't merge in the latest master from upstream into my PR, then it won't be able to checkout that SHA.

I didn't fully understand, is our CI not cloning the entire repository? I think there is a way to configure this in prow, see `clonerefs`: https://github.com/kubernetes/test-infra/blob/master/prow/pod-utilities.md. But I didn't fully research into this.

TAOXUY: I remember when we run the CI, it will auto-{rebase|merge} on master the moment it runs. The problem of latest commit won't happen. Instead, if someone checks in the code between test start and build finish, the API Regression test will get a new SHA which it doesn't exist in the git tree.

nareddyt: Good point, i forgot about the auto-merge when running jobs. Sounds good then

PR #305, opened 2020-08-31
b/166158386 #3: Add ApiKeyState into report log entry

nareddyt:
This will make analysis of requests with API Keys easier in the future. Log entries will contain 1 of the 4 states now:
- NOT CHECKED
- VERIFIED
- NOT ENABLED
- INVALID

Testing done: Unit and integration tests.

Follow-up PR: Consider adding a new state `MISSING`. If a method requires an API Key but one is not provided, we consider that `NOT_CHECKED` now. While that's technically correct (we don't make a CHECK call), to the API Producer that's a little misleading. This is low priority though.


tests/utils/service_control_utils_test.go
```diff
@@ -445,6 +445,12 @@ operations: <
         value: <
           string_value: "api-key"
         >
+      >
+			fields: <
```
qiwzhang, line 449: Please fix the indent.

nareddyt: Oops, tabs instead of spaces... Done

Issue #304, opened 2020-08-29
ESPv2 not reporting metrics and log entries correctly when deploying with a service account (non-gcp)


gojohnkevin:
I was originally using esp v1 then migrated to v2 using GKE. After migrating to esp-v2, I cannot see the latency times anymore in Cloud Endpoints->Services->Overview page.

Is this a problem with my configuration? I just changed the version from 1 -> 2 as stated in the migration guide.

TAOXUY:
I am able to see it for a service deployed on GKE[1][2]. Could you check the access log by:
- go to sidebar -> Logging -> Logs Viewer -> the target service[3]
- send a request and see if it appears on the log

[1]
![image](https://user-images.githubusercontent.com/16471764/91640791-f0c4f680-e9d4-11ea-9813-aaf2ddb5f070.png)
[2]
![image](https://user-images.githubusercontent.com/16471764/91640804-0afed480-e9d5-11ea-8ff4-d710085747d4.png)
[3]
![image](https://user-images.githubusercontent.com/16471764/91640846-5d3ff580-e9d5-11ea-84d1-590a0ecb17ef.png)

gojohnkevin:
@TAOXUY It actually does show in the log viewer but the cloud endpoints graph is empty since migration.

![Screenshot 2020-08-30 at 12 33 43 AM](https://user-images.githubusercontent.com/9165701/91641708-eb10f480-ea58-11ea-996a-8246ec953de3.png)

![Screenshot 2020-08-30 at 12 37 32 AM](https://user-images.githubusercontent.com/9165701/91641726-0419a580-ea59-11ea-9022-df6e93cc87d1.png)

qiwzhang:
Do you see any lines in "Request" and "Error" graphs.  Cloud you play with
time tags at the right top, e.g. "1 hour", "6 hours", '1 day", ... to see
if any graph shows your ESPv2 traffic?
Generally,  if logging shows your traffic,  graphs should show too.

Did you use same deployment environment for v1 and v2,  e.g. project, GKE
cluster, service account, ?

gojohnkevin:
@qiwzhang I have already done that. The only graph that shows is when I was still using v1. 

I have not changed any configuration in my deployment except changing the version of the docker image to 1->2. All other configuration is the same.

gojohnkevin:
What I noticed too is that the location label in the log payload is empty for v2.
[...]

PR #307, opened 2020-09-01
Add default location in Report call for non_gcp

qiwzhang:


qiwzhang:
To fix 2) issue in https://github.com/GoogleCloudPlatform/esp-v2/issues/304

PR #311, opened 2020-09-01
b/167400401: Set API version in autogenerated CORS methods

nareddyt:
Generated methods should have the `Version` copied from the `api` they are associated with. Otherwise it shows up weirdly in the endpoints log.


src/go/configinfo/service_info.go
```diff
@@ -423,19 +423,20 @@ func (s *ServiceInfo) addOptionMethod(apiName string, path string, backendInfo *
...
 	corsOperation := fmt.Sprintf("%s_%s", corsOperationBase, formattedPath)
-	genOperation := fmt.Sprintf("%s.%s", apiName, corsOperation)
+	genOperation := fmt.Sprintf("%s.%s", originalMethod.ApiName, corsOperation)
 
 	s.Methods[genOperation] = &methodInfo{
-		ShortName: corsOperation,
-		ApiName:   apiName,
+		ShortName:  corsOperation,
+		ApiName:    originalMethod.ApiName,
+		ApiVersion: originalMethod.ApiVersion,
```
qiwzhang, line 431: How do we use this "ApiVersion" field?   Did we pass it to Envoy?

nareddyt: Yes, it gets passed to the SC Filter

https://github.com/GoogleCloudPlatform/esp-v2/blob/5de5d988a2650a3978a0a3cf9ebfbd47aa3da0c5/api/envoy/v8/http/service_control/requirement.proto#L106

This is used in the reports. It shows up in the log entries.

qiwzhang: I see.  Thanks

PR #316, opened 2020-09-03
Rich access logging of http request information

nareddyt:
**Background:** A few weeks ago, Google Service Managment made some rich log formatting available in the `LogEntries` for a report. This will make the Endpoints Logs much more readable, Stackdriver Logging will extract this information and allow for smart filtering/grouping/sorting/displaying. Example below.

**Enhancement:** We **move** (not copy) some HTTP request-specific fields from the `jsonPayload` struct into the `httpRequest` struct. They are still all under the same log entry and maintain the same values, just the field path changes.

Spoke with Wayne offline, we are OK with making this minor breaking change. Shouldn't be a big deal for our users, it only changes queries for access logs slightly.

image


qiwzhang (approved):

TAOXUY:
Please update the graph to the latest access log(the moved entries are still in `jsonPayload`).

TAOXUY (approved):

nareddyt:
> Please update the graph to the latest access log(the moved entries are still in `jsonPayload`).

Done

nareddyt:
Api regression test is expected to break with this change.

PR #318, opened 2020-09-08
Fix reading the remote client IP when deploying ESPv2 on Cloud Run

### LINKED ISSUE #197: API Key with IP address restriction blocks all IP including whitelisted
I've created ESPv2 Cloud Endpoint for my API, everything works fine except when I enabling IP address restriction in API Key, I get a message "PERMISSION_DENIED: IP address blocked." for all requests including whitelisted IP addresses.

What you expected to happen:
Block all requests except whitelisted.

Steps to reproduce:
Create IP restriction for API Key for Endpoint base on ESPv2.

I've checked the endpoint logs and see the internal google IP address instead of my IP address from which I send a request, a corresponding screenshot is attached.
Screenshot 2020-06-13 at 14 41 40

nareddyt:
**Background:** On Cloud Run, we currently don't read the client IP address correctly. We always get some google internal IP address because we configure `num_trusted_hops` too high. This results in Envoy not using the `X-Forwarded-For` header, and instead uses the IP of the downstream connection (which is a google-internal proxy).

**Fix:** Modify flags to read client IP from the xff header. Even though we are an edge proxy, Cloud Run and GFE sanitize headers. We can trust the xff header:
- When a request has no xff header, Cloud Run will set the header to contain the single remote client IP. We can trust this.
- When a request contains the xff header, it's not trusted. But Cloud Run will sanitize it by appending it's own google internal ip, not the remote client IP. We can trust this.

So NO need to set `use_remote_address=true`, and just use `num_trusted_hops=0` to extract the client IP. For more details on these flags, here is [some documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-forwarded-for).

**Implementation Details:**
- When the serverless image is used, have `env_start_proxy.py` set a specific flag `on_serverless=true`.
- Have `start_proxy.py` set serverless-specific defaults for `compute_platform_override` and `envoy_xff_num_trusted_hops`. Note: This doesn't modify behavior for sidecar deployments, users still may have to set that. But for Cloud Run, the new default will work well.
- Remove `compute_platform_override` flag from the public API.

**Testing Done:**
- Manually verified my IP address shows up in the access log when deployed with the default of 0 trusted hops.
- Enabled e2e api key restriction test (with ip restrictions) for our serverless deployment. This fakes the xff header, so override to 3 trusted hops.

Fixes #197


tests/e2e/scripts/linux-test-kb-long-run.sh
```diff
@@ -101,15 +101,13 @@ while true; do
...
-      python ${ROOT}/tests/e2e/client/apiproxy_bookstore_key_restriction_test.py  \
-        --host="${SCHEME}://${HOST}:${PORT}"   \
-        --allow_unverified_cert=true  \
-        --key_restriction_tests=${ROOT}/tests/e2e/testdata/bookstore/key_restriction_test.json.template  \
+  echo "Starting bookstore API Key restriction test at $(date)."
+  (set -x;
+    python ${ROOT}/tests/e2e/client/apiproxy_bookstore_key_restriction_test.py  \
+      --host="${SCHEME}://${HOST}:${PORT}"   \
+      --allow_unverified_cert=true  \
+      --key_restriction_tests=${ROOT}/tests/e2e/testdata/bookstore/key_restriction_test.tmpl.json  \
```
qiwzhang, line 109: Are you sure the file name?  should it be json.template?

tests/e2e/scripts/cloud-run/deploy.sh
```diff
@@ -326,7 +326,10 @@ function setup() {
     -i "${APIPROXY_IMAGE}"
 
   # Redeploy ESPv2 to update the service config
-  proxy_args="^++^--tracing_sample_rate=0.05++--tracing_outgoing_context=traceparent++--tracing_incoming_context=traceparent"
+  proxy_args="^++^--tracing_sample_rate=0.05"\
+"++--tracing_outgoing_context=traceparent"\
+"++--tracing_incoming_context=traceparent"\
+"++--envoy_xff_num_trusted_hops=3"
```
qiwzhang, line 332: Why hop=3?

tests/e2e/scripts/linux-test-kb-long-run.sh
nareddyt: Reverted

tests/e2e/scripts/cloud-run/deploy.sh
nareddyt: Our e2e tests are not testing real client ip restrictions. If we were validating that the e2e client's IP was the one being restricted, then the default hop=0 would work.

Instead, our e2e tests set up fake IPs in the xff header, and they validate against the fake IPs. Here is one example testcase:

https://github.com/GoogleCloudPlatform/esp-v2/blob/585e66acc4d22b49a3116ad44d1457559cfd4f0c/tests/e2e/testdata/bookstore/key_restriction_test.json.template#L3-L12

On GKE, this works fine with the default hop=2. On Serverless, Cloud Run guarantees to only append 1 IP to the end (which doesn't happen on GKE). So  use hop=3.

docker/generic/start_proxy.py
```diff
@@ -731,7 +741,11 @@ def gen_proxy_config(args):
         proxy_conf.extend(["--v", "0"])
 
     if args.envoy_xff_num_trusted_hops:
-         proxy_conf.extend(["--envoy_xff_num_trusted_hops", args.envoy_xff_num_trusted_hops])
+        proxy_conf.extend(["--envoy_xff_num_trusted_hops",
+                           args.envoy_xff_num_trusted_hops])
+    elif args.on_serverless:
+        proxy_conf.extend(["--envoy_xff_num_trusted_hops",
```
qiwzhang, line 747: Logic wise, I feel this logic should be part of serverless/env_start_proxy.py.   We don't want to move that logic here.
Imaging,  we may have another deploying mode: xyz  calling start_proxy.py,   we don't want to move its logic into here too.

tests/e2e/scripts/cloud-run/deploy.sh
qiwzhang: I see.  Could you add some comment for this code for better readability?

qiwzhang (changes requested):

docker/generic/start_proxy.py
nareddyt: I considered this, but it would result in a lot of code duplication. 

We need to account for the case where the user wants to override `envoy_xff_num_trusted_hops`. If we set it to 0 directly in `env_start_proxy.py`, then we need to add another flag parser to allow the user to override it.

So instead we just signal what platform we're on and pass all the user's flags directly from `env_start_proxy` to `start_proxy`. Since `start_proxy` has the `args_parser`, it's much easier to handle here.

qiwzhang: Ok,  I see your point

qiwzhang (approved):

TAOXUY (approved):

TAOXUY:
`ESPv2-anthos-cloud-run-e2e-anthos-cloud-run-http-bookstore` is broken by this PR.

nareddyt:
> `ESPv2-anthos-cloud-run-e2e-anthos-cloud-run-http-bookstore` is broken by this PR.

Ah sorry, I assumed it was related to the flakes I've been seeing with this job recently. I should have taken a closer look at all the log lines. I'll fix once I'm done working on a customer bug.

Issue #317, opened 2020-09-04
GoogleRE2.max_program_size field is deprecated in Envoy


lmakarov:
I get a bunch of these warnings from Envoy in the ESPV2 logs:

```
2020-09-04 10:16:14.235 PDTW0904 17:16:14.035 16 envoy] [16][misc][external/envoy/source/common/protobuf/message_validator_impl.cc:21] Deprecated field: type envoy.type.matcher.v3.RegexMatcher.GoogleRE2 Using deprecated option \'envoy.type.matcher.v3.RegexMatcher.GoogleRE2.max_program_size\' from file regex.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

From Envoy [v1.15.0 release notes](https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.15.0#deprecated):

> The [GoogleRE2.max_program_size](https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/regex.proto#envoy-v3-api-field-type-matcher-v3-regexmatcher-googlere2-max-program-size) field is now deprecated. Management servers are expected to validate regexp program sizes instead of expecting the client to do it. Alternatively, the max program size can be enforced by Envoy via runtime.

nareddyt:
Thanks, I can clean it up next week.

Some background: I actually [asked Envoy](https://github.com/envoyproxy/envoy/issues/9174#issuecomment-587108068) to raise the limit or deprecate this field because it the default program size is too low. We had to [roll back](https://groups.google.com/u/1/g/google-cloud-endpoints/c/x5zTU0Wkem0/m/0oxdp0x-AQAJ) release 2.3.0 due to the default program size. 

Glad to see it's deprecated!

TAOXUY:
I will quickly update this field along with other deprecated fields.

PR #320, opened 2020-09-09
Update deprecated API fields

TAOXUY:
- For `envoy.extensions.access_loggers.file.v3.FileAccessLog.format`, update to the latest.
    
- For `envoy.type.matcher.v3.RegexMatcher.GoogleRE2.max_program_size`:
   - by  this [PR](https://github.com/envoyproxy/envoy/pull/10971/files), remove this field in envoy and add the validation on the control plane
   - by this [PR](https://github.com/envoyproxy/envoy/pull/11187), the data plane actually still keep the validation so set `re2.max_program_size.error_level=1k` in runtime key to make it same as before


src/go/configgenerator/route_generator.go
```diff
@@ -248,15 +249,15 @@ func makeHttpRouteMatcher(httpRule *commonpb.Pattern) *routepb.RouteMatch {
 			},
 		}
 	} else {
+		if err := validateRegexProgramSize(regex, util.GoogleRE2MaxProgramSize); err != nil {
+			return nil, fmt.Errorf("invalid route path regex: %v", err)
```
qiwzhang, line 253: error is not clear.  The error message should
"invalid regex {} from UriTemplate {}"

src/go/configgenerator/route_generator_test.go
```diff
@@ -381,3 +415,13 @@ func TestMakeRouteConfigForCors(t *testing.T) {
 		}
 	}
 }
+
+// Used to generate a oversize cors origin regex or a oversize wildcard path regex.
+func getOversizeRegex() string {
```
qiwzhang, line 420: getOversizeUriTemplateForTest

nareddyt:
Fixes #317

src/go/configgenerator/route_generator.go
```diff
@@ -272,5 +273,23 @@ func makeHttpRouteMatcher(httpRule *commonpb.Pattern) *routepb.RouteMatch {
 			},
 		},
 	}
-	return &routeMatcher
+	return &routeMatcher, nil
+}
+
+func validateRegexProgramSize(regex string, maxProgramSize int) error {
```
nareddyt, line 279: should we move this to utils? Nothing routing specific here.

src/go/configgenerator/route_generator_test.go
TAOXUY: Rename it to be `getOverSizeRegexForTest` since it will also be used to generate `cors origin regex`.

qiwzhang (approved):

nareddyt (approved):

src/go/configgenerator/route_generator.go
TAOXUY: Before, I regarded that as a helper function used for route_generator only since most likely it wouldn't be shared by other places. 
I think either way is fine. Move it to `regex.go` now.

toddbeckman-g (approved):

TAOXUY:
`ESPv2-anthos-cloud-run-e2e-anthos-cloud-run-http-bookstor` is broken in submitted https://github.com/GoogleCloudPlatform/esp-v2/pull/318.

PR #322, opened 2020-09-10
b/145520483: Update example with support for dynamic routing of unconfigured requests

nareddyt:
Verifies that a top-level `x-google-backend` works with `x-google-allow: all`.

Need to create a follow-up PR to fix the deterministic ordering of Envoy route config. See comments in this PR for more details.


examples/testdata/path_matcher/envoy_config.json
```diff
@@ -332,6 +413,101 @@
...
+                            "match": {
+                              "headers": [
+                                {
+                                  "exactMatch": "DELETE",
+                                  "name": ":method"
+                                }
+                              ],
+                              "safeRegex": {
+                                "googleRe2": {},
+                                "regex": "^/.*$"
```
qiwzhang, line 427: In envoy route match, the order is important.  These wildcard match should be at last.  How do we make sure that?

nareddyt: Good catch, we don't have any logic for this. We just append them in the order we process backend rules. I'll come up with a way of fixing this...

nareddyt: I did some more research. Is the order actually important here for our current implementation?

We don't use Envoy's router to match operation name, path matcher will take care of that. Path matcher will match to the most specific rule when wildcards are involved (see example unit test below). So we'll always get the correct operation, for any ordering of rules path matcher config or router config.

https://github.com/GoogleCloudPlatform/esp-v2/blob/e6afa078b92b4dabf60edc0c330218dd453bcd60/src/api_proxy/path_matcher/path_matcher_test.cc#L142-L145

qiwzhang: I know for path_match, it always match the specific one first, ahead of wildcard one. regardless order
e.g /foo/*,    /**,   for a /foo/bar request, it will match /foo/*,   not /** , more specific one has higher priority than non specific ones.

But Envoy RouteMatch really depends on ordering.  We use it for routing to different backends (cluster)
e.g. we have a toplevel x-google-backend and a path level one
x-google-backend
   address: backendA

paths:
      /foo/{x}
      x-google-backend:
          address: B

If we put RountMatch  "/foo/* after /**,   then we never route to address B even path_matcher match to the correct operator,  but we route to a wrong cluster.  

We may need a integration test for this case.

nareddyt: Good point, this will fail.

Here is my current plan: OpenAPI compiler guarantees to generate the wildcard methods last. We just need to ensure when we generate the envoy route config, we use the same order as the `httpRules`.

Our current logic does not follow this. It actually sorts the operation names alphabetically, which is not the best approach (and resulted in our tests not catching it, we got lucky in our sorting).

https://github.com/GoogleCloudPlatform/esp-v2/blob/41da9dd0ac337ef2c5635124e9ae3134de47bc33/src/go/configinfo/service_info.go#L138-L142

How about this: Let's submit this PR to show the current (incorrect) config generation. I'll make a PR tomorrow to fix this. The diff will be large, prefer a separate PR.

qiwzhang: I was thinking about: we need a way to detect such `/**` wildcard backend rules,  and put them at the last in the RouteMatch,  

SG: to have this in a separate PR.

qiwzhang (approved):

examples/testdata/path_matcher/envoy_config.json
nareddyt: Detecting `/**` wildcard rules would work, but it feels hacky. I'd prefer to just use the order given in the service config, which currently guarantees `/**` rules will be last.

PR #323, opened 2020-09-10
Support IP in backend address

qiwzhang:


nareddyt (approved):
Do you think we should add a few test cases for this function as well?

https://github.com/GoogleCloudPlatform/esp-v2/blob/3f7e63eb6fa39308e8e2a070f3aacfff8c82c312/src/go/util/url_util_test.go#L27

Your current tests cover it indirectly, but maybe we can add it for completeness.

nareddyt (approved):

PR #325, opened 2020-09-11
b/168120858: Split specifying root certs for backend vs sidestream SSL clients

nareddyt:
**Background**: Customers expect to use `--ssl_client_root_certs_file` to configure the root cert for requests made to the backend. But we currently use it for all requests, including Google Service Management, Google Service Control, etc.

**Implementation details**:
- On the config manager side: Split into two flags, one for backend (`--ssl_backend_client_root_certs_path`), one for sidestream (`--ssl_sidestream_client_root_certs_path`).
- In `start_proxy.py`: Only expose the backend root cert flag. Don't expose the sidestream one, as users won't need it. However, we still keep it in config manager for testing.

**Flag rename and deprecation**: For clarity the flags are operating on the backend connection, explicitly rename them to have `backend`. This is consistent with ESPv1:
- Rename `ssl_client_cert_path` -> `ssl_backend_client_cert_path`
- Rename `ssl_client_root_certs_file` -> `ssl_backend_client_root_certs_file`

This is **backwards compatible**. The original flag names still exist and can be used, but are treated as deprecated.

**Testing done**:
- Modified integration tests to use the new split flags. Specifically, only some tests in `transport_security_test.go` use `--ssl_sidestream_client_root_certs_path`. All dynamic routings tests use `--ssl_backend_client_root_certs_path` by default for https backends.
- Added a regression test to ensure the flags do not cross over


docker/generic/start_proxy.py
```diff
@@ -150,12 +150,12 @@ def make_argparser():
         HTTP/2 secure connections on listener_port. Requires the certificate and
         key files "server.crt" and "server.key" within this path.''')
 
-    parser.add_argument('--ssl_client_cert_path', default=None, help='''
+    parser.add_argument('--ssl_backend_client_cert_path', default=None, help='''
```
qiwzhang, line 153: What about multiple backends?   Each backend has its own ssl cert files or root files?  Do we want to support that?

It is OK we don't support it now, but we may need to have a plan that it is easy to support it in the future, especially not introduce a breaking changes.

src/go/configmanager/flags/flags.go
```diff
@@ -54,13 +54,14 @@ var (
...
-	SslMinimumProtocol   = flag.String("ssl_minimum_protocol", "", "Minimum TLS protocol version for Downstream connections.")
-	SslMaximumProtocol   = flag.String("ssl_maximum_protocol", "", "Maximum TLS protocol version for Downstream connections.")
-	RootCertsPath        = flag.String("root_certs_path", util.DefaultRootCAPaths, "Path to the root certificates to make TLS connection.")
-	EnableHSTS           = flag.Bool("enable_strict_transport_security", false, "Enable HSTS (HTTP Strict Transport Security).")
-	DnsResolverAddresses = flag.String("dns_resolver_addresses", "", `The addresses of dns resolvers. Each address should be in format of either IP_ADDR or IP_ADDR:PORT and they are separated by ';'.`)
+	SslServerCertPath                = flag.String("ssl_server_cert_path", "", "Path to the certificate and key that ESPv2 uses to act as a HTTPS server")
+	SslSidestreamClientRootCertsPath = flag.String("ssl_sidestream_client_root_certs_path", util.DefaultRootCAPaths, "Path to the root certificates to make TLS connection to all external services other than the backend.")
```
qiwzhang, line 58: is this "sidestream" just for ssl integration test? How easy to change that test to test backend ssl?  If it is easy, I prefer not to introduce this flag.

docker/generic/start_proxy.py
nareddyt: I see. We can follow the same approach as we decided for the `--backend` flag:
1. For now, keep the `--ssl_backend_client_cert_path` flag that applies to all backends.
2. If we need to support multiple backends in the future, then add it into the service config. Have this flag override it if the field is empty.

nareddyt:
Hi @qiwzhang and @JLXIA , are you OK with the flag rename to have `backend` in the name? It's technically a breaking change now, but I doubt it will affect many people. PR description has more details.

qiwzhang (approved):

TAOXUY (approved):

nareddyt:
Hii @qiwzhang, can you take one more quick look at my latest commits? I modified `start_proxy` to treat the previous flags as deprecated to maintain backwards compatibility.

qiwzhang (approved):

PR #329, opened 2020-09-15
Removed un-used segmentNames in samples envoy config

qiwzhang:

Since bootstrap/status test is using "allowUnknownFields", these removed fields in the expected envoy json are not removed.


nareddyt (approved):

nareddyt:
For reference, it seems I missed removing these fields in PR #218.

TAOXUY (approved):

PR #330, opened 2020-09-15
b/167586109: Add name prefix for backend/jwtProvider cluster

TAOXUY:
Currently, we use address to name these 2 types of clusters. The change will
- for backend clusters: add `backend-cluster-` prefix
- for jwtProvider clusters: add `jwt-provider-cluster-` prefix

It helps to distinguish with other clusters.

### DIFF
--- a/src/go/bootstrap/static/bootstrap_test.go
+++ b/src/go/bootstrap/static/bootstrap_test.go
@@ -34,7 +34,7 @@ var (
 	FakeConfigID = "2019-12-16r0"
 )
 
-func zTestServiceToBootstrapConfig(t *testing.T) {
+func TestServiceToBootstrapConfig(t *testing.T) {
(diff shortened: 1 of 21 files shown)


src/go/bootstrap/static/bootstrap_test.go
```diff
@@ -34,7 +34,7 @@ var (
 	FakeConfigID = "2019-12-16r0"
 )
 
-func TestServiceToBootstrapConfig(t *testing.T) {
+func zTestServiceToBootstrapConfig(t *testing.T) {
```
qiwzhang, line 37: why the test is diabled?

qiwzhang (approved):

nareddyt (approved):

toddbeckman-g (approved):

PR #332, opened 2020-09-16
Cleanup autogeneted CORS method names

nareddyt:
**Background**
- In commit c84fc101db498634b58a0c74d10e0a6f98289bf7, we renamed CORS operations to use paths instead of indexes to avoid user confusion and increase debuggability. This was hacky, but we did not have a better solution at the time.
- A lot later, we made PR #255 to deduplicate semantically equivalent paths. This PR introduces logic that can be re-used now to cleanup operation naming.

**Example**: For a method with name `GetBook` when `allowCors = true`:
- Before we generated: `ESPv2_Autogenerated_CORS_{path}`
- Now we will now generate: `ESPv2_Autogenerated_CORS_GetBook`

Keep in mind that not every single method and/or HttpRule will be duplicated, as we account for semantically equivalent paths.

**Why is this needed now?** Based on the discussion in #322, we need to change our operation ordering logic (not in this PR, in a follow-up PR). Implementation wise, it's easier to do this if all methods are created with the same function `getOrCreateMethod`. CORS method generation is the only code that does not use this function. But to switch it over, we need to standardize the method names.

**Testing Done**
- Verified changes to example envoy configs.
- Added some more unit tests for CORS method generation.


src/go/configinfo/service_info.go
```diff
@@ -408,35 +407,23 @@ func (s *ServiceInfo) processHttpRule() error {
...
-		HttpRule: []*commonpb.Pattern{
-			{
-				UriTemplate: path,
-				HttpMethod:  util.OPTIONS,
-			},
-		},
-		IsGenerated: true,
-		BackendInfo: originalMethod.BackendInfo,
+func (s *ServiceInfo) addOptionMethod(originalMethod *methodInfo, path string) error {
+	genOperation := fmt.Sprintf("%s.%s_CORS_%s", originalMethod.ApiName, util.AutogeneratedOperationPrefix, originalMethod.ShortName)
```
qiwzhang, line 411: My concern is: this generated operation may not be unique since it only use apiName + ShortName
If it collapse with other cors method, if the other method has different backend,  they this CORS will route to a wrong backend.
I am not sure if ApiName + ShortName is unique enough
My understanding is ApiName +  ServiceName + Method,  are unique,  ShortName is Method. 
If there are two Rpc Services with same Method name under same Api,  we may have problem.

TAOXUY: +1. IMHO, the route won't be messed up since it is using regex path match -> cluster but the name may be duplicate.

qiwzhang: My concern is invalid.  Api is the same as gRPC service
ApiName + Method_name should be unique.

qiwzhang (approved):

src/go/configinfo/service_info.go
nareddyt: Wayne and I discussed offline since I didn't fully understand, there should be no problems with that.

@TAOXUY can you clarify what problems you see? Keep in mind, `ApiName + . + MethodName` forms a selector, which is guaranteed to be unique. In fact, it's the regex path match that may not be unique, which is why PR #255 de-dups them. This PR still keeps that logic, as you can see by the new test case that was added.

TAOXUY: Same with Wayne concern, didn't see the ApiName included the serviceName. 

For the regex, I mean being unique after dedup in #255.

TAOXUY (approved):

PR #335, opened 2020-09-17
b/145520483: Fix Envoy route ordering for wildcard dynamic routing

nareddyt:
**Background**: We have two path matchers in ESPv2 - the path matcher filter and Envoy's route config. Envoy's route config matches in first-to-win ordering. This can cause problems when wildcard paths are involved. If the wildcard path comes before more specific paths, then we may do host-rewrite incorrectly for backend routing. PR #322 has more description of this.

**Fix**: Modify the deterministic method ordering to always use the same ordering as the incoming service config. Just remove the line that sorts methods alphabetically, so we always output configs (route config, filter config, etc) in the service config method order.

**Testing**:
- Fix lots of unit tests.
- Add integration test that will fail host rewrite when method ordering is changed.


tests/env/env.go
```diff
@@ -258,9 +258,10 @@ func (e *TestEnv) SkipHealthChecks() {
 // In the service config for each backend, the backend port is represented with a "-1".
 // Example: Address: "https://localhost:-1/"
 // During env setup, replace the -1 with the actual dynamic routing port for the test.
+// To force routing to a non-existent backend host, use "9".
 func addDynamicRoutingBackendPort(serviceConfig *confpb.Service, port uint16) error {
 	for _, rule := range serviceConfig.Backend.GetRules() {
-		if !strings.Contains(rule.Address, "-1") {
+		if !strings.Contains(rule.Address, "-1") && !strings.Contains(rule.Address, "9") {
```
nareddyt, line 264: This is really messy and I have a commit to extract this out into a constant variable. But the diff is big, prefer to make a follow-up PR instead.

src/go/configinfo/service_info.go
```diff
@@ -40,9 +39,12 @@ type ServiceInfo struct {
 
 	// An array to store all the api names
 	ApiNames []string
-	// A sorted array to store all the method name for this service.
-	// Should always iterate this array to avoid test fail due to order issue.
+
+	// A ordered slice of operation names. Follows the same order as the service config.
+	// Ordering is important, both for route matching and testability.
+	// All functions that output order-dependent configs should use this ordering.
```
TAOXUY, line 45: Could you also comment the first-win match of order dependency of route matching?

Now we are depending on the order from service config(OpenAPI compiler), which is likely to break. Can we we sort order by ourselves?

qiwzhang: 1) I am not clear what order in service configs we are following?  apiMathod?  HttpRules?  Backend Rules? 
2) for allow=all wildcard,  are we sure they are in the last with such ordering?

examples/testdata/path_matcher/envoy_config.json
```diff
@@ -491,34 +491,34 @@
...
-                              "hostRewriteLiteral": "http-bookstore-abc9876-uc.a.run.app",
+                              "cluster": "backend-cluster-http-bookstore-edf123456-uc.a.run.app:443",
+                              "hostRewriteLiteral": "http-bookstore-edf123456-uc.a.run.app",
                               "timeout": "15s"
                             }
                           },
                           {
                             "decorator": {
-                              "operation": "ingress Google_Autogenerated_Unrecognized_Put_Method_Call"
+                              "operation": "ingress ESPv2_Autogenerated_CORS_Google_Autogenerated_Unrecognized_Get_Method_Call"
```
qiwzhang, line 516: This is good.  It shows that /** routes are after other normal routes.  and /** for OPTIONS are after.  This should work.
<pre>
Normal routes:
allow_all wildcard routes for GET, POST ...  except OPTIONS
OPTIONS for normal http rules
OPTIONS for allow_all wildcard
</pre>
I am just nervous, that could this be accidentally,  or this order is guaranteed  by api-compler

src/go/configinfo/service_info.go
nareddyt: Done. See other comment about OpenAPI compiler. 

I highly recommend that we don't sort them, I feel that is even more brittle. And what if a user manually specifies their own http rule? Then our sorting could break them. It's best to take the order of the service config, which the user can easily modify if needed.

examples/testdata/path_matcher/envoy_config.json
nareddyt: Right now, the OpenAPI compiler guarantees the `Google_Autogenerated_Unrecognized_...` will show up last.

I understand why you are nervous though, a future change in the OpenAPI compiler could break this. I plan to add a test to the OpenAPI compiler to guarantee the autogenerated methods are always last. This should make it safe to depend upon.

qiwzhang (approved):

src/go/configinfo/service_info.go
TAOXUY: SG

TAOXUY (approved):

PR #336, opened 2020-09-17
Enable gRPC when downstream is HTTP/1.1

qiwzhang:

In order to support gRPC when downstream is HTTP/1.1,   
1) Envoy Http_connection_manager Http1ProtocolOptions need to set enable_trailers=true
2) Need to add grpc_metadata_scrubber to remove Content-Length for gRPC reqeuests.

Add a configMgr option and flag EnableGrpcForHttp1.  
Default is on, since it doesn't harm anything.


qiwzhang:
PTAL

nareddyt (approved):

toddbeckman-g (approved):

src/go/configmanager/flags/flags.go
```diff
@@ -160,6 +162,7 @@ func EnvoyConfigOptionsFromFlags() options.ConfigGeneratorOptions {
 		SuppressEnvoyHeaders:                    *SuppressEnvoyHeaders,
 		UnderscoresInHeaders:                    *UnderscoresInHeaders,
 		ServiceControlNetworkFailOpen:           *ServiceControlNetworkFailOpen,
+		EnableGrpcForHttp1:                      *EnableGrpcForHttp1,
```
JLXIA, line 165: Do we need this flag here, or just default to false? I thought this is only needed for ApiGateway when it runs on Titanium.

TAOXUY: +1.

qiwzhang: Decided to add a Option to ConfigMgr.  We don't need a flag.  But for now,  every ConfigMgr option has a flag, so just add a flag.

PR #339, opened 2020-09-18
Fix quota test

TAOXUY:
The quota test often [failed](https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/GoogleCloudPlatform_esp-v2/337/ESPv2-cloud-run-e2e-cloud-run-http-bookstore/1306763442949984257) because read on null response returned by exception. Fix it by removing first redundant attempt.


tests/e2e/client/apiproxy_bookstore_quota_test.py
```diff
@@ -57,14 +57,12 @@ def _try_call_quota_read():
 
     # exhaust the quota in the current window.
     print("Exhaust current quota...");
-    response = _try_call_quota_read()
-    if response.status_code != 429:
```
TAOXUY, line 61: This is buggy line.

PR #344, opened 2020-09-22
Add flag `--connection_buffer_limit_bytes`

### LINKED ISSUE #341: Add start-up flag to set per_connection_buffer_limit_bytes setting in envoy
### Description of problem
It seems that the response size of json-transcoded responses processed by espv2 is limited by the **per_connection_buffer_limit_bytes** setting parameter in envoy. This defaults to 1mb, and all json transcoded responses that are larger than this will fail. GRPC responses are 

See envoy related issue in gitlab for more info: [Size limit on gRPC-JSON transcoder with google.api.HttpBody as output?](https://github.com/envoyproxy/envoy/issues/6441)

Posted a more detailed description of the problem in the endpoints discussion group here:
[Increasing max size of response for requests that are passed through json transcoding](https://groups.google.com/g/google-cloud-endpoints/c/v1k5zTd4LBw/m/MN19CEzkAgAJ)

### Proposed resolution

Add a startup flag for espv2 that allows to set the **per_connection_buffer_limit_bytes** setting parameter in envoy to a custom buffer limit. This will allow for the espv2 gateway container to process and return larger responses.

nareddyt:
This flag is passed through to Envoy. It allows response bodies to be larger, which is especially useful during gRPC transcoding.

Testing done: Added integration test that reproduces the transcoding problem in #341 when the limit is too small.

Fixes #341

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -548,6 +548,13 @@ environment variable or by passing "-k" flag to this script.
         Otherwise use ignored_query_parameters. Defaults to false.
         ''')
 
+    parser.add_argument(
+        '--connection_buffer_limit_bytes', action=None,
+        help='''
+        Configure the maximum amount of data that is buffered for each 
+        request/response body. This should be increased for large bodies used 
+        by gRPC transcoding. Default is 1 MB.
+        ''')
 
     # Start Deprecated Flags Section
 
@@ -979,6 +986,10 @@ def gen_proxy_config(args):
 
     if args.enable_debug:
         proxy_conf.append("--suppress_envoy_headers=false")
+
+    if args.connection_buffer_limit_bytes:
+        proxy_conf.extend(["--connection_buffer_limit_bytes",
+                           args.connection_buffer_limit_bytes])
 
     return proxy_conf
 
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -301,6 +301,16 @@ environment variable or by passing "-k" flag to this script.
         sidecar deployments and 0 for serverless deployments.''')
 
     parser.add_argument(
+        '--envoy_connection_buffer_limit_bytes', action=None,
+        help='''
+        Configure the maximum amount of data that is buffered for each 
+        request/response body, in bytes. If not set, default is decided by
+        Envoy.
+        
+        https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto
+        ''')
+
+    parser.add_argument(
         '--log_request_headers',
         default=None,
         help='''Log corresponding request headers through
@@ -546,14 +556,6 @@ environment variable or by passing "-k" flag to this script.
         corresponding protobuf field in grpc-json transcoding. Use this if you
         cannot control the query parameters and do not know them beforehand.
         Otherwise use ignored_query_parameters. Defaults to false.
-        ''')
-
-    parser.add_argument(
-        '--connection_buffer_limit_bytes', action=None,
-        help='''
-        Configure the maximum amount of data that is buffered for each 
-        request/response body, in bytes. If not set, default is decided by
-        Envoy (currently 1 MiB). 
         ''')
 
     # Start Deprecated Flags Section
@@ -987,9 +989,9 @@ def gen_proxy_config(args):
     if args.enable_debug:
         proxy_conf.append("--suppress_envoy_headers=false")
 
-    if args.connection_buffer_limit_bytes:
+    if args.envoy_connection_buffer_limit_bytes:
         proxy_conf.extend(["--connection_buffer_limit_bytes",
-                           args.connection_buffer_limit_bytes])
+                           args.envoy_connection_buffer_limit_bytes])
 
     return proxy_conf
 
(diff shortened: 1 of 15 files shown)


docker/generic/start_proxy.py
```diff
@@ -548,6 +548,13 @@ def make_argparser():
         Otherwise use ignored_query_parameters. Defaults to false.
         ''')
 
+    parser.add_argument(
+        '--connection_buffer_limit_bytes', action=None,
+        help='''
+        Configure the maximum amount of data that is buffered for each 
+        request/response body. This should be increased for large bodies used 
+        by gRPC transcoding. Default is 1 MB.
```
qiwzhang, line 556: Do we need to mention for gRPC?  what happen to HTTP?

src/go/configmanager/flags/flags.go
```diff
@@ -97,6 +97,9 @@ var (
 
 	EnableGrpcForHttp1 = flag.Bool("enable_grpc_for_http1", true, `Enable gRPC when the downstream is HTTP/1.1. The default is on.`)
 
+	ConnectionBufferLimitBytes = flag.Int("connection_buffer_limit_bytes", 1*1024*1024, `Configure the maximum amount of data that is buffered for each request/response body. 
+			This should be increased for large bodies used by gRPC transcoding. Default is 1 MB.`)
```
qiwzhang, line 101: I am debating: if this flag is not set,  should we introduce espv2 default, or just use envoy default. 

IMO, we (espv2) is not in the business of deciding such default.  we should let envoy to decide it.

docker/generic/start_proxy.py
nareddyt: It's very unclear to me how this buffering works in Envoy. When I set this flag to a small value and ran all our integration tests, only the gRPC transcoding tests failed. All gRPC and HTTP pass-through tests worked just fine. Hence it seems this will only be a problem for gRPC transcoding, which is why I explicitly mention it here.

src/go/configmanager/flags/flags.go
nareddyt: Well Envoy [defaults to 1 MB](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto). We should document this for our users, right? Otherwise they don't know what the default will be.

qiwzhang: Sorry,  I did not state it clearly.  I would prefer

If this flag is not set, ESPv2 don't set this setting in envoy config.  it will use envoy default value.
We don't want to define our own default value.  which may be different with future envoy default value.

docker/generic/start_proxy.py
qiwzhang: In theory, it should apply to http too. I prefer not to mention grpc here.

src/go/configmanager/flags/flags.go
nareddyt: I think we should explicitly set a default. What if Envoy lowers it in the future? It could break our clients.

I doubt Envoy would change it anyways, so I don't think we need to worry. It's easier to unit test if we always include it.

qiwzhang: If envoy ever change its default, it must be a reason, to fix somethings. we may want to benefits from that changes too. For backward compatibility,  I don't think Envoy will change a default to break anybody.  that change will not be a breaking change for any users, including ours. 
For your unit-test argument, I don't think we have unit-test to test 1MB default. with actual data flow.  You may mean the unit-test verifying envoy config setting.  "not setting it" can also be verified in unit-test too.

docker/generic/start_proxy.py
nareddyt: Done. What do you think about documenting the default as I did here?

src/go/configmanager/flags/flags.go
nareddyt: Makes sense, done.

docker/generic/start_proxy.py
```diff
@@ -548,6 +548,13 @@ def make_argparser():
         Otherwise use ignored_query_parameters. Defaults to false.
         ''')
 
+    parser.add_argument(
+        '--connection_buffer_limit_bytes', action=None,
+        help='''
+        Configure the maximum amount of data that is buffered for each 
+        request/response body, in bytes. If not set, default is decided by
+        Envoy (currently 1 MiB).
```
TAOXUY, line 556: Maybe no need to mention the number, otherwise:
- we need to keep it sync with envoy
- possibly mislead our users if we forget

qiwzhang: Agreed with Jason, not to mention the actual default value. maybe add a link to envoy doc for this setting.

qiwzhang: Wondering if we should prefix the flag name with --envoy-.  since this is specific envoy setting.  just like envoy-xff
and create a envoy specific flag section, and move it there.

nareddyt: Done. I changed the start_proxy flag to prefix with Envoy. The config manager flag does not have that prefix for now.

TAOXUY (approved):

qiwzhang (approved):

Issue #345, opened 2020-09-22
Change --service_control_network_fail_open flag


qiwzhang:
This flag default is true already,  by specifying it,  it is also true.  So it is confusing. 
  
Please see this https://github.com/cloudendpoints/esp/issues/816 for detail

Suggested changes: keep this flag for backward compatible, but mark it as deprecated
add a new flag
--service_control_network_fail_policy=[open|close]  default is open.

PR #348, opened 2020-09-23
Deprecated flag --service_control_network_fail_open

qiwzhang:

Replaced it with --service_control_network_fail_policy=[open|close]

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -328,13 +328,11 @@ environment variable or by passing "-k" flag to this script.
         if the fields are available. The value must be a primitive field,
         JSON objects and arrays will not be logged.
         ''')
-    parser.add_argument(
-        '--service_control_network_fail_open',
-        default=True,
-        action='store_true',
-        help='''
-        In case of network failures when connecting to Google service control,
-        the requests will be allowed if this flag is on. The default is on.
+    parser.add_argument('--service_control_network_fail_policy',
+        default='open',  choices=['open', 'close'], help='''
+        Specify the policy to handle the request in case of network failures when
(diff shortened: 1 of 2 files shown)


qiwzhang:
https://github.com/GoogleCloudPlatform/esp-v2/issues/345

TAOXUY (approved):

nareddyt (approved):

docker/generic/start_proxy.py
```diff
@@ -328,13 +328,11 @@ def make_argparser():
...
         JSON objects and arrays will not be logged.
         ''')
-    parser.add_argument(
-        '--service_control_network_fail_open',
-        default=True,
-        action='store_true',
-        help='''
-        In case of network failures when connecting to Google service control,
-        the requests will be allowed if this flag is on. The default is on.
+    parser.add_argument('--service_control_network_fail_policy',
```
TAOXUY, line 331: This isn't backward compatible, right? If users setting `service_control_network_fail_open=False`. This flag is undefined? 

Actually I just realized for this [pr](https://github.com/cloudendpoints/esp/pull/818), look neither. `service_control_network_fail_open` will always be true even if users setting `service_control_network_fail_open=false` without setting service_control_network_fail_policy, which makes backward uncompatible?

qiwzhang: For ESPv2,  since it is in beta,  we don't need to worry about backward compatibility.  
Beside you could not use flag --service_control_network_fail_open=False.  as it is tested by users. Please see
[here](https://github.com/cloudendpoints/esp/issues/816) So with old flag, you could not turn if off.

TAOXUY: SG

TAOXUY (approved):

PR #349, opened 2020-09-24
b/162392946: #1 add response code detail in service control report

TAOXUY:
This series of PRs are adding rc_details including from ESPv2 filters and upstream ones to service control report. In this PR,
- add rc_detail into report as `response_code_detail`
- expend service control filter rc_details:
   - service_control_check_error{`CHECK_RESPONSE_ERROR_NAME`}, ex. service_control_check_error{SERVICE_NOT_ACTIVATED}
   - service_control_check_network_failure{`GRPC_STATUS_NAME`}, ex. service_control_check_call_failure{UNAVAILABLE}
   - service_control_quota_error{`QUOTA_RESPONSE_ERROR_NAME`}
   - service_control_quota_network_failure{`GRPC_STATUS_NAME`}
   - service_control_bad_request{MISSING_API_KEY}
   - service_control_bad_request{UNDEFINED_REQUEST}


TAOXUY:
Ready to review. Will fix the integration test later.

src/envoy/http/service_control/client_cache.cc
```diff
@@ -80,6 +81,14 @@ constexpr uint32_t kReportDefaultNumberOfRetries = 5;
 // The default value for network_fail_open flag.
 constexpr bool kDefaultNetworkFailOpen = true;
 
+// The service control call 4xx failure name, used in response code detail
+// later.
+constexpr char kScCallFail4XX[] = "4XX";
```
nareddyt, line 86: 4xx, lowercase. Then it will be consistent with our report labels

https://github.com/GoogleCloudPlatform/esp-v2/blob/fa28e9a6c337f82394f6fd4157c42a76417fbb6f/tests/utils/service_control_utils.go#L238

src/envoy/http/service_control/client_cache.cc
```diff
@@ -442,18 +452,20 @@ void ClientCache::callQuota(const AllocateQuotaRequest& request,
...
+            *response, config_.service_name(), &response_info);
 
     collectScResponseErrorStats(response_info.error_type);
-    on_done(quota_status);
+    on_done(quota_status, response_info);
   } else {
     // Most likely an auth error in ESPv2 or API producer deployment.
     filter_stats_.filter_.denied_producer_error_.inc();
-    on_done(http_status);
+    response_info.error_name = kScCallFail5XX;
```
nareddyt, line 466: This could either be a 4xx or 5xx, you need some logic to differentiate between it. Or we could just put the full status code, or we could just generalize to `http_call_fail`.

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -54,6 +55,26 @@ const Envoy::Http::LowerCaseString kAndroidCertHeader{"x-android-cert"};
...
+}
+
+std::string QuotaErrorNameToRcDetail(absl::string_view rc_detail) {
+  return absl::StrCat("service_control_quota_error{", rc_detail, "}");
+}
+
+struct RcDetailsValues {
+  // The request isn't defined in the service config.
+  const std::string UndefinedRequest =
+      "service_control_bad_request{UNDEFINED_REQUEST}";
```
nareddyt, line 69: A little confusing. How about `UNDEFINED_OPERATION`, `OPERATION_NOT_MATCHED`, etc.

src/envoy/http/service_control/client_cache.cc
nareddyt: Can we rename it to `http_call_4xx`?

These are used to say the actual HTTP call failed, not that Check returned a 4xx error status. I think just saying `4xx` is too general.

nareddyt: On that note, why do we need to generalize to 4xx and 5xx? Why can't we just put the full status code?

`http_call_403`, `http_call_504`.

TAOXUY: Done. Use the status name instead of status code.

src/envoy/http/service_control/handler_impl.cc
TAOXUY: I was trying to not use implementation terms, especially in the view of API Gateway, like what we did in https://github.com/GoogleCloudPlatform/esp-v2/pull/350.

Relatively, `OPERATION` is unknown to customers and they only know requests.

TAOXUY:
PTAL

src/api_proxy/service_control/request_info.h
```diff
@@ -141,18 +141,34 @@ enum ScResponseErrorType {
...
+
   // The trust level of the API Key that was checked.
-  api_key::ApiKeyState api_key_state = api_key::ApiKeyState::NOT_CHECKED;
+  api_key::ApiKeyState api_key_state;
+
+  CheckResponseInfo()
+      : consumer_project_number(),
+        consumer_type(),
+        consumer_number(),
+        error({"", false, ScResponseErrorType::ERROR_TYPE_UNSPECIFIED}),
```
qiwzhang, line 170: can we define a default constructor for ScResponseError? move these default values there

src/api_proxy/service_control/request_info.h
```diff
@@ -141,18 +141,34 @@ enum ScResponseErrorType {
   CONSUMER_QUOTA = 5,
 };
 
+// Store the info of service control http call failure/call error response.
+struct ScResponseError {
+  std::string name;
```
qiwzhang, line 146: name is kind of confusing?   is this "message"?

src/api_proxy/service_control/request_info.h
```diff
@@ -141,18 +141,34 @@ enum ScResponseErrorType {
...
+
+  bool is_error_from_http_call;
```
qiwzhang, line 148: is_network_error?

we may switch to grpc.

src/api_proxy/service_control/request_info.h
```diff
@@ -166,7 +182,12 @@ struct QuotaRequestInfo : public OperationInfo {
 
 // Stores the information extracted from the quota response.
 struct QuotaResponseInfo {
-  ScResponseErrorType error_type = ScResponseErrorType::ERROR_TYPE_UNSPECIFIED;
+  ScResponseErrorType type;
```
qiwzhang, line 185: why do we still need "type"?  is it covered by "error"?

src/envoy/http/service_control/client_cache.cc
```diff
@@ -80,6 +81,13 @@ constexpr uint32_t kReportDefaultNumberOfRetries = 5;
 // The default value for network_fail_open flag.
 constexpr bool kDefaultNetworkFailOpen = true;
 
+// Convert http error status into the error name.
+std::string httpFailStatusToErrorName(const Status& status) {
+  return absl::StrCat(
+      "HTTP_CALL_",
+      absl::StatusCodeToString(static_cast<absl::StatusCode>(status.code())));
```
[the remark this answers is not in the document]
qiwzhang: Do users need to know we are using Http call?

src/envoy/http/service_control/client_cache.cc
```diff
@@ -387,6 +393,13 @@ void ClientCache::handleCheckResponse(const Status& http_status,
                 "Google Service Control Check is unavailable, and the "
                 "request is denied due to network fail closed, with error: {}",
                 final_status.error_message());
+
+      // If http_status is not ok, the Code::UNAVAILABLE is from http_status.
+      if (!http_status.ok()) {
+        response_info.error = {httpFailStatusToErrorName(http_status),
```
qiwzhang, line 399: prefer to define a function to take http_status as input and return a "error" struct.

src/envoy/http/service_control/client_cache.cc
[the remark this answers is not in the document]
TAOXUY: Yes, they are. 

https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.h

src/api_proxy/service_control/request_info.h
TAOXUY: This is the error name of service control response, which will be the response code detail, like `SERVICE_NOT_ACTIVATED` or `UNAVAILABLE`.

src/api_proxy/service_control/request_info.h
TAOXUY: Yes, I think so. @nareddyt 

This PR is very big already. We can refactor in the next PR.

src/api_proxy/service_control/request_info.h
qiwzhang: how about error_from_server?

src/api_proxy/service_control/request_info.h
```diff
@@ -167,7 +176,7 @@ struct CheckResponseInfo {
       : consumer_project_number(),
         consumer_type(),
         consumer_number(),
-        error({"", false, ScResponseErrorType::ERROR_TYPE_UNSPECIFIED}),
+        error(),
```
qiwzhang, line 179: do we need to list it out if a member variable is using default constructor? 

can we remove this constructor and use initializer "=" for api_key_state

src/api_proxy/service_control/request_info.h
```diff
@@ -145,9 +145,18 @@ enum ScResponseErrorType {
 struct ScResponseError {
   std::string name;
 
-  bool is_error_from_http_call;
+  bool is_network_error;
 
   ScResponseErrorType type;
```
qiwzhang, line 150: could you add comment on each member variable?

src/api_proxy/service_control/request_info.h
```diff
@@ -145,9 +145,18 @@ enum ScResponseErrorType {
 struct ScResponseError {
   std::string name;
 
-  bool is_error_from_http_call;
+  bool is_network_error;
```
qiwzhang, line 148: I am just wondering if we can just put rpc_status here.   if rpc_status is not OK, it means nework failure.  otherwise it is error from server.

src/api_proxy/service_control/request_info.h
nareddyt: Seems you wrapped `ScResponseErrorType` inside `ScResponseError`, so you should be able to remove it. This is the same thing you did for CheckResponseInfo

src/envoy/http/service_control/client_cache.cc
```diff
@@ -387,6 +394,11 @@ void ClientCache::handleCheckResponse(const Status& http_status,
                 "Google Service Control Check is unavailable, and the "
                 "request is denied due to network fail closed, with error: {}",
                 final_status.error_message());
+
+      // If http_status is not ok, the Code::UNAVAILABLE is from http_status.
+      if (!http_status.ok()) {
```
nareddyt, line 399: move this up to line 370? No need for duplicate if statement

src/api_proxy/service_control/request_info.h
nareddyt: It can be either from SC server or it can be a network error. I'm ok with `name`, maybe `display_name`. This is essentially what gets exposed in the log entry.

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -54,6 +56,32 @@ const Envoy::Http::LowerCaseString kAndroidCertHeader{"x-android-cert"};
 constexpr char JwtPayloadIssuerPath[] = "iss";
 constexpr char JwtPayloadAudiencePath[] = "aud";
 
+std::string CheckErrorToRcDetail(const ScResponseError& error) {
+  return error.is_network_error
+             ? absl::StrCat("service_control_check_call_failure{", error.name,
```
nareddyt, line 61: I like this logic, just a small NIT on naming. Maybe `service_control_check_network_failure`?

src/envoy/http/service_control/handler_impl.cc
nareddyt: Ack

src/envoy/http/service_control/handler_impl.cc
```diff
@@ -323,6 +359,8 @@ void ServiceControlHandlerImpl::callReport(
   }
   info.response_size = stream_info_.bytesSent() + response_header_size;
 
+  info.response_code_detail = stream_info_.responseCodeDetails().value_or("");
```
nareddyt, line 362: So what are the cases where it will be empty? Only on successful requests now?

src/api_proxy/service_control/request_info.h
TAOXUY: This is not as straightforward and we don't need more information from rpc_status.

src/envoy/http/service_control/handler_impl.cc
TAOXUY: For successful one, it will be "via_upstream".

src/envoy/http/service_control/client_cache.cc
TAOXUY: It is not duplicate. 

It needs satisfy both `http_status` is `UNAVALIABLE` and `network_fail_open_` is `false`. Moving to 370 needs checking the latter.

qiwzhang (approved):

src/envoy/http/service_control/client_cache.cc
nareddyt: Ack

tests/integration_test/websocket_test.go
```diff
@@ -83,6 +83,7 @@ func TestWebsocket(t *testing.T) {
 					LogMessage:                   "1.echo_api_endpoints_cloudesf_testing_cloud_goog.WebsocketEcho is called",
 					StatusCode:                   "0",
 					ResponseCode:                 101,
+					ResponseCodeDetail:           "via_upstream",
```
nareddyt, line 86: Remove default value?

nareddyt (approved):
After seeing the integration tests, this looks like a very useful feature. Thanks!

PR #350, opened 2020-09-24
b/169095072: fix path matcher misleading error message

TAOXUY:
For mismatched request, update error message from `"Path does not match any requirement URI template."`
to `` "Request`METHOD PATH` is not defined by this API"``.

Please see the bug for the detailed discussion.

### DIFF
--- a/src/envoy/http/path_matcher/filter.cc
+++ b/src/envoy/http/path_matcher/filter.cc
@@ -77,7 +77,7 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
     rejectRequest(
         Envoy::Http::Code::NotFound,
         absl::StrCat("`", method, " ", path, "`"
-                     " does not match any requirement URI template."));
+                     " does not match any defined routes."));
     return Envoy::Http::FilterHeadersStatus::StopIteration;
   }
 
--- a/src/envoy/http/path_matcher/filter.cc
+++ b/src/envoy/http/path_matcher/filter.cc
@@ -76,7 +76,7 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
   if (rule == nullptr) {
     rejectRequest(
         Envoy::Http::Code::NotFound,
-        absl::StrCat(method, " ", path,
+        absl::StrCat("`", method, " ", path, "`"
                      " does not match any requirement URI template."));
     return Envoy::Http::FilterHeadersStatus::StopIteration;
   }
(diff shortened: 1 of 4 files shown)


src/envoy/http/path_matcher/filter.cc
```diff
@@ -76,7 +76,7 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
   if (rule == nullptr) {
     rejectRequest(
         Envoy::Http::Code::NotFound,
-        absl::StrCat(method, " ", path,
+        absl::StrCat("`", method, " ", path, "`"
                      " does not match any requirement URI template."));
```
qiwzhang, line 80: How about:
Request with (POST /path) is not defined in the service configuration.

TAOXUY: How about `` `GET /pets/spot` does not match any defined routes`` ? 

Currently, the service configuration is transparent to downstream. It is unnecessary to mention this.

src/envoy/http/path_matcher/filter_test.cc
```diff
@@ -166,7 +166,7 @@ TEST_F(PathMatcherFilterTest, DecodeHeadersNoMatch) {
   EXPECT_CALL(
       mock_cb_,
       sendLocalReply(Envoy::Http::Code::NotFound,
-                     "`POST /bar` does not match any requirement URI template.",
+                     "`POST /bar` does not match any defined routes.",
```
qiwzhang, line 169: "defined routes" is confusing too.  When users see this?  what can they do?

I think we can mention "service config".  For now, consumer client (or users) are the same as producer, they know about "service config"

nareddyt: Well, not really. For Cloud API Gateway, API Producers don't know about service config either. They only know about OpenAPI spec, service config should be transparent to them.

nareddyt (approved):

src/envoy/http/path_matcher/filter_test.cc
qiwzhang: how about
.... is not defined in service spec.  

OpenAPI is a service spec. 
It will also applies to gRPC too,  grpc spec is another service spec.

TAOXUY: Make sense.

qiwzhang (approved):

TAOXUY:
Hi @nareddyt , please take another look.

nareddyt (approved):

PR #354, opened 2020-09-29
Envoy route config for sidecar with x-google-backend

nareddyt:
Configuration for ESPv2 in sidecar mode, but with `x-google-backend` defined to configure timeouts per route.

NOTE: Currently we **do not generate the correct config**, the `deadline` in `x-google-backend` is ignored in sidecar mode. But let's check in this config for now, it will make it easier for me to reason about what needs to change in config generator. And it will give me a good test case for per-route config implementation.

Ref #321 (does not fix the issue)

### DIFF
--- a/src/go/bootstrap/static/bootstrap_test.go
+++ b/src/go/bootstrap/static/bootstrap_test.go
@@ -25,9 +25,9 @@ import (
 	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
 	"github.com/GoogleCloudPlatform/esp-v2/tests/env/platform"
 	"github.com/golang/protobuf/jsonpb"
+	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
 
 	bootstrappb "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
-	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
 )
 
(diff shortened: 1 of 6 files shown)


src/go/bootstrap/static/bootstrap_test.go
```diff
@@ -95,61 +95,67 @@ func TestServiceToBootstrapConfig(t *testing.T) {
...
-			continue
-		}
+			var s confpb.Service
+			if err := unmarshaler.Unmarshal(bytes.NewBuffer(configBytes), &s); err != nil {
+				t.Errorf("Unmarshal() returned error %v, want nil", err)
+			}
 
-		opts := flags.EnvoyConfigOptionsFromFlags()
-		tc.opt_mod(&opts)
+			opts := flags.EnvoyConfigOptionsFromFlags()
```
TAOXUY, line 128: NIT: different indent?

TAOXUY (approved):

src/go/bootstrap/static/bootstrap_test.go
nareddyt: I moved everything into `t.Run`, everything should be indented. I don't think it's a different indent, i ran `make format`.

TAOXUY: SG

PR #358, opened 2020-09-30
b/170162456: Enable route config for local backend in sidecar mode

nareddyt:
Previously, we made a single route when running ESPv2 in sidecar mode. But now we plan to deprecate the path matcher filter and support configuring timeouts for the local backend (#321). To do this, we should switch to configuring the Envoy route table for individual routes for even the local backend.

There is one extra feature introduced by this PR: We need to support matching wildcard HTTP Methods when using a [Custom Http Pattern](https://github.com/googleapis/googleapis/blob/cb7fc620590382a4a2ea6ffdf6f51ae0e77bbbb5/google/api/http.proto#L343). This just required an if statement and extra test in `route_generator.go`.

**High-risk change**

Summarized comments from me and @qiwzhang:

> We have to use regex on “path” to support url_template.   But path_matcher supports many complicated forms that regex could not support.  Currently we use these regex to replace url templates, but not all templates are supported.

> I think it is fine,  our endpoints users are not using very complicated url templates.  We only need to support the url templates from openapi spec, and  options.http.rules from grpc proto file. We are still in Beta (not GA), so it should be fine.

**Testing Done**
- Fix lots of unit and integration tests.
- Micro benchmark of the performance impact of switching to per-route config instead of path prefix: https://github.com/envoyproxy/envoy/pull/13320


tests/integration_test/opencensus_tracing_test.go
```diff
@@ -87,7 +87,7 @@ func TestTracesServiceControlCheckWithRetry(t *testing.T) {
 				"Service Control remote call: Check",
 				"Service Control remote call: Check - Retry 1",
 				"Service Control remote call: Check - Retry 2",
-				"ingress",
+				"ingress ListShelves",
```
TAOXUY, line 90: This may be unrelated to this PR. Can we use more specific name? This look possibly duplicate for multi-backends.

nareddyt: This is intentional, span names have a length limit. If we include the API Name, we get very close to hitting it. Especially for OpenAPI Serverless, where the API Name is the entire cloud run URL.

https://github.com/GoogleCloudPlatform/esp-v2/blob/ea798af34f615caeeb016cded7977aafdd7ab6e6/src/go/configgenerator/route_generator.go#L223

TAOXUY (approved):

src/go/configgenerator/route_generator_test.go
```diff
@@ -186,49 +207,45 @@ func TestMakeRouteConfig(t *testing.T) {
 					Rules: []*annotationspb.HttpRule{
 						{
 							Selector: "endpoints.examples.bookstore.Bookstore.Foo",
-							Pattern: &annotationspb.HttpRule_Get{
-								Get: "/v1/{book_name=*}/test/**",
+							Pattern: &annotationspb.HttpRule_Custom{
+								Custom: &annotationspb.CustomHttpPattern{
+									Path: "/v1/{book_name=*}/test/**",
+									Kind: "*",
```
qiwzhang, line 213: This is not intended usage.   Users normally will not specify "*" in custom kind.

nareddyt: I think it's supported. https://github.com/googleapis/googleapis/blob/cb7fc620590382a4a2ea6ffdf6f51ae0e77bbbb5/google/api/http.proto#L340

Also we have an integration test for this: https://github.com/GoogleCloudPlatform/esp-v2/blob/96c61ffd01f1c6eb0a90aa96aa8d1e0143f147dd/tests/integration_test/grpc_fallback_test.go#L57

src/go/configinfo/service_info.go
```diff
@@ -515,6 +518,30 @@ func (s *ServiceInfo) processBackendRule() error {
 	return nil
 }
 
+// For methods that are not associated with any backend rules, create one
+// to associate with the local backend cluster.
+func (s *ServiceInfo) processLocalBackendOperations() error {
+	for _, operation := range s.Operations {
+		method, err := s.getOrCreateMethod(operation)
```
qiwzhang, line 525: only "get",  we should not be "creating" a new method here.

src/go/configgenerator/route_generator_test.go
qiwzhang: I see. Thanks

src/go/configinfo/service_info.go
nareddyt: Done

qiwzhang (approved):

nareddyt:
Breakages in the API regression test are expected, the pre-existing span names changed.

PR #361, opened 2020-10-02
Fix api regression test

TAOXUY:
Currently, we are using the latest envoy with the master configmanager to do the api compatibility test for API Gateway uses case. However, its rollout order requires the opposite that uses the master old envoy and the latest configmanager.

### DIFF
--- a/prow/gcpproxy-api-regression.sh
+++ b/prow/gcpproxy-api-regression.sh
@@ -65,4 +65,4 @@ git checkout ${SHA}
 echo ${VERSION} > ${ROOT}/VERSION
 
 make depend.install
-make integration-test-without-envoy-build
+make integration-test-without-configmanager-build
--- a/prow/gcpproxy-api-regression.sh
+++ b/prow/gcpproxy-api-regression.sh
@@ -44,6 +44,11 @@ exit 1; }
 # The test will use the latest configmanager(configgen library) with the MASTER envoy to
 # run all the MASTER integrations.
 
+# Build latest configmanager.
+make depend.install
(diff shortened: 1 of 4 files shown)


prow/gcpproxy-api-regression.sh
```diff
@@ -58,4 +57,14 @@ git checkout ${SHA}
 echo ${VERSION} > ${ROOT}/VERSION
 
 make depend.install
-make integration-test-without-envoy-build
+make build build-envoy build-grpc-interop build-grpc-echo
+
+echo '======================================================='
+echo '============ Download latest configmanager ============'
```
nareddyt, line 63: Honestly we can remove this download dependency right? Building config manager is really quick, I don't think you're saving much time here. We can just build config manager before checking out `master` on line 48.

nareddyt (changes requested):

prow/gcpproxy-api-regression.sh
TAOXUY: Done.

scripts/all-utilities.sh
```diff
@@ -414,14 +414,16 @@ function sleep_wrapper() {
   echo "$(date): End sleep"
 }
 
-function envoy_binary_gcs_path() {
```
nareddyt, line 417: Can we remove these functions?

TAOXUY: Done.

nareddyt (approved):

qiwzhang (approved):

PR #366, opened 2020-10-07
b/162392946: #2 normalize rc details in the rest filters.

TAOXUY:
Add following response code details:
- `path_matcher_bad_request{MISSING_METHOD}`
- `path_matcher_bad_request{MISSING_PATH}`
- `path_matcher_bad_request{OVERSIZE_PATH}`
- `path_matcher_undefined_request`
- `backend_auth_missing_backend_token`
- `backend_auth_undefined_rqeuest`
- `backend_routing_bad_request{MISSING_PATH}`
- `backend_routing_bad_request{PATH_WITH_FRAGMENT_IDENTIFIER}`
- `backend_routing_undefined_rqeuest`
- `service_control_bad_request{UNDEFINED_REQUEST}` -> `service_control_undefined_request`

**Detail**:
- put rc detail related error names/util function in `rc_detail_utils.[h|cc]`
- generate the rc detail in the runtime
- add unit tests in all the filter_test


nareddyt (commented):
If we're using `missing_operation` in backend auth and backend routing, let's change SC to also use it.

https://github.com/GoogleCloudPlatform/esp-v2/blob/9c25358123da6c4fed660bed8a7264e800e36959/src/envoy/http/service_control/handler_impl_test.cc#L329

src/envoy/http/backend_auth/filter.cc
```diff
@@ -43,11 +43,11 @@ RegisterCustomInlineHeader<CustomInlineHeaderRegistry::Type::RequestHeaders>
     authorization_handle(CustomHeaders::get().Authorization);
 
 struct RcDetailsValues {
-  // The request is rejected due to missing backend auth token in internal
-  // filter config.
-  const std::string MissingBackendToken = "missing_backend_token";
-  // The request is rejected due to missing operation in internal filter state.
-  const std::string MissingOperation = "missing_operation";
+  // Missing backend auth token in internal filter config.
+  const std::string MissingBackendToken = "backend_auth_missing_backend_token";
```
nareddyt, line 47: Can we have some common utility function shared across the filters to form the RC detail?

Some function that does `absl::StrCat(filter_name, error_type, "{", error_name, "}").

I'm specifically worried about `error_type`, it should be shared across filters. `bad_request`, `missing_operation` are all common.

src/envoy/http/backend_routing/filter.cc
```diff
@@ -28,7 +28,15 @@ namespace backend_routing {
 
 namespace {
 struct RcDetailsValues {
-  const std::string BadRequest = "bad_request";
+  // The request doesn't contain PATH header.
+  const std::string MissingPath = "backend_routing_bad_request{MISSING_PATH}";
```
qiwzhang, line 32: Will these changed rc_detail show up in the Report call?   If so,  I don't see the change in integration tests

TAOXUY:
> If we're using `missing_operation` in backend auth and backend routing, let's change SC to also use it.
> 
> https://github.com/GoogleCloudPlatform/esp-v2/blob/9c25358123da6c4fed660bed8a7264e800e36959/src/envoy/http/service_control/handler_impl_test.cc#L329

All use UndefinedRequest now.

src/envoy/http/backend_routing/filter.cc
TAOXUY: There is one integration test change.

Most of them are prevented by path matcher filter.

qiwzhang (approved):

src/envoy/http/backend_auth/filter.cc
TAOXUY: Good point, done.

qiwzhang (approved):

nareddyt (approved):

PR #368, opened 2020-10-09
Use per-route config for ServiceControl filter

qiwzhang:
This is one of serveal PR to refactory envoy filters to remove path_matcher. 

This PR is to only change ServiceControl filter 
* Add PerRouteFilterConfig with just one field operation_name
* Change handler_impl to read per_route from stream_info->routeEntry().

PR is ready for review


src/envoy/http/service_control/config_parser.h
```diff
@@ -36,6 +38,10 @@ constexpr int64_t kDefaultMinStreamReportIntervalMs = 10000;
 constexpr int64_t kLowerBoundMinStreamReportIntervalMs = 100;
 }  // namespace
 
+// The filter name.
+const std::string ServiceControlFilterName =
```
nareddyt, line 42: go/totw/140

I think we build with c++ 17 now, right? Then `inline` should work.

```
inline constexpr absl::string_view kFilterName
```

src/envoy/http/service_control/handler_impl_test.cc
```diff
@@ -194,11 +196,27 @@ class HandlerTest : public ::testing::Test {
     counter.reset();
   }
 
+  void setPerRouteOperation(const std::string& operation) {
+    ::espv2::api::envoy::v9::http::service_control::PerRouteFilterConfig
+        per_route_cfg;
+    per_route_cfg.set_operation_name(operation);
+    auto per_route = std::make_shared<PerRouteFilterConfig>(per_route_cfg);
```
nareddyt, line 203: Should this be saved as a class member variable? It will be used in the lambda after this `setPerRouteOperation` function returns. I guess the lambda captures the shared_pointer, but it looks confusing.

nareddyt (commented):
Overall LGTM, but the integration test failures are concerning...

src/envoy/http/service_control/handler_impl_test.cc
qiwzhang: Yes, lambda capture add a reference to the shared_ptr.  
Not need to be a test class member since this is the only place using it.

src/envoy/http/service_control/config_parser.h
qiwzhang: No, in my local, we are not using c++17

TAOXUY (approved):
Nice to see how cool the perRouteConfig works!

src/envoy/http/service_control/filter.cc
```diff
@@ -35,14 +45,57 @@ Envoy::Http::FilterHeadersStatus ServiceControlFilter::decodeHeaders(
     Envoy::Http::RequestHeaderMap& headers, bool) {
   ENVOY_LOG(debug, "Called ServiceControl Filter : {}", __func__);
 
-  Envoy::Tracing::Span& parent_span = decoder_callbacks_->activeSpan();
+  if (!headers.Method()) {
```
nareddyt, line 48: Do you want to add these here already? I'd prefer to add these in when we actually remove path matcher.

If we add these in this PR, we should also copy in unit tests that cover these cases.

src/envoy/http/service_control/config_parser.cc
```diff
@@ -67,6 +67,7 @@ FilterConfigParser::FilterConfigParser(const FilterConfig& config,
   // Construct a requirement for non matched requests
   non_match_rqm_cfg_.set_service_name(first_srv_ctx->config().service_name());
   non_match_rqm_cfg_.set_operation_name(kUnrecognizedOperation);
+  non_match_rqm_cfg_.mutable_api_key()->set_allow_without_api_key(true);
```
nareddyt, line 70: Why is this needed? I see the changes to `handler_impl`, but it seems like it will work without this line (?)

qiwzhang: Agreed, it is no need. removed.

src/envoy/http/service_control/filter.cc
qiwzhang: Well, I just moved these checking and x-http-method-override here too.   I have added more unit-tests for them

PR #371, opened 2020-10-10
b/170423927: replace snakeName with jsonName in urlTemplate during config time

TAOXUY:
**Description** Replace snakeName with jsonName in the urlTemplates provided to path matcher filters during config time. It can help cut down config provided to filter and reduce computation in the runtime.

**Context** The variable names defined by users in service spec/OP config are in `jsonName` while the url templates generated in OP config all use `snakeName`(`name` field). In the current implementation, the path matcher filter will do the variable extraction with snakeName and transform variable name to jsonName in the run time.

**NextStep** 
- Remove variable mapping related codes in path matcher filter
- increment API version


src/go/bootstrap/static/bootstrap_test.go
```diff
@@ -153,7 +153,7 @@ func TestServiceToBootstrapConfig(t *testing.T) {
 				t.Fatal(err)
 			}
 			if err := util.JsonEqual(wantString, gotString); err != nil {
-				t.Fatalf("got err: %v", err)
+				t.Fatalf("Test(%s) fail, got err: %v", tc.desc, err)
```
nareddyt, line 156: No need, this is inside t.Run, it prints the test name.

src/go/util/url_util.go
```diff
@@ -198,6 +198,54 @@ func WildcardMatcherForPath(uri string) string {
...
+// TODO(taoxuy@): extract variable name by syntax parsing.
+func SnakeNamesToJsonNamesInPathParam(uri string, snakeNameToJsonName map[string]string) string {
+	findPathParamIndex := func(uri, snakeName string) int {
+		for {
+			index := strings.Index(uri, snakeName)
+			if index == -1 {
+				return -1
+			}
+
+			if index != 0 && (uri[index-1] == '{' || uri[index-1] == '.') && (uri[index+len(snakeName)] == '}' || uri[index+len(snakeName)] == '.' || uri[index+len(snakeName)] == '=') {
```
nareddyt, line 223: This is very confusing to me. Can we replace with a regex capture group? Or can you add a comment on what's being checked?

src/go/configinfo/service_info.go
```diff
@@ -716,9 +716,12 @@ func (s *ServiceInfo) processTypes() error {
 			}
 		}
 
-		// Store the mapping per selector.
+		// Replace the snake name with the json name in url template
 		if len(snakeToJson) > 0 {
-			mi.SegmentMappings = snakeToJson
+			for _, httpRule := range mi.HttpRule {
```
nareddyt, line 721: This is the wrong place to do it. This can lead to problems in route matching. We shouldn't modify the envoy route config with JSON name, only the backend routing filter.

- Example: /v1/{book=books_segment/*}
- This request should match: /v1/books_segment/42
- But you'll change the UriTemplate to /v1/{book=booksSegment/*}
- Now that request won't match.

nareddyt (changes requested):

src/go/configinfo/service_info.go
qiwzhang: We only change filed_name, not value.   In above case, we will not replace books_segment which is not a field name.

examples/testdata/path_matcher/envoy_config.json
```diff
@@ -206,14 +206,10 @@
                         "rules": [
                           {
                             "operation": "1.examples_path_matcher_endpoints_cloudesf_testing_cloud_goog.GetShelf",
-                            "pathParameterExtraction": {
-                              "snakeToJsonSegments": {
-                                "shelf_id": "shelfId"
-                              }
-                            },
+                            "pathParameterExtraction": {},
```
qiwzhang, line 209: Please change path_matcher filter config to remove snake_to_json_segments field

qiwzhang: well,  this require change envoy c++ code.   let us change it, we don't want dead code

src/go/configinfo/service_info_test.go
```diff
@@ -2445,11 +2445,16 @@ func TestProcessTypes(t *testing.T) {
 		desc                            string
 		fakeServiceConfig               *confpb.Service
 		fakeRequestTypeNamesByOperation map[string]string
+		urlTemplateByOperation          map[string]string
 		wantSegmentsByOperation         map[string]SnakeToJsonSegments
```
qiwzhang, line 2449: should we remove wantSegmentByOperation

tests/integration_test/dynamic_routing_integration_test.go
```diff
@@ -51,49 +51,49 @@ func TestDynamicRouting(t *testing.T) {
...
-			method:   "GET",
-			wantResp: `{"RequestURI":"/dynamicrouting/getpetbyid?lang=U%20S&zone=us%3Bwest&pet_id=a%20b&number=9%3B8"}`,
-		},
-		{
-			desc:     "Succeed, CONSTANT_ADDRESS path translation with snake case is correct",
-			path:     "/shelves/123/books/info/987",
-			method:   "GET",
-			wantResp: `{"RequestURI":"/dynamicrouting/bookinfo?SHELF=123&BOOK=987"}`,
-		},
+		//{
```
qiwzhang, line 54: why are they been commented out?

src/go/configinfo/service_info.go
nareddyt: Ack got it, was confused. Then this is fine, i don't see any other downsides to changing it here.

examples/testdata/path_matcher/envoy_config.json
TAOXUY: It would be a lot to remove the dead code. I intend to put in another PR as I put in the description.

src/go/util/url_util.go
TAOXUY: I described the matching logic in the line 207 and also commented.  Is it clear to you? 

The regex I think of is not supported in go lib(not sure if there is equivalent one). As it is not that complicated, we implement here.

qiwzhang (approved):

nareddyt (approved):

src/go/util/url_util.go
nareddyt: Thanks, makes sense

TAOXUY:
`ESPv2-anthos-cloud-run-e2e-anthos-cloud-run-http-bookstore` is broken.

PR #372, opened 2020-10-12
b/170423927: #2cleanup dead code for snakeName to jsonName conversion

TAOXUY:
*Description*: after #371, cleanup dead code for snakeName to jsonName conversion.

*Followup*: increment API version.


api/envoy/v9/http/path_matcher/config.proto
```diff
@@ -25,16 +25,6 @@ message PathMatcherRule {
...
-  // If set, the path parameters in the `pattern` must be extracted. Ref:
-  // https://cloud.google.com/endpoints/docs/openapi/openapi-extensions#understanding_path_translation
-  PathParameterExtractionRule path_parameter_extraction = 3;
-}
-
-message PathParameterExtractionRule {
-  // Each rule has a mapping of snake_case segments to jsonCase segments.
-  // If the map is empty, the path params still need to be extracted.
-  map<string, string> snake_to_json_segments = 2;
 }
```
qiwzhang, line 28: We could not completely remove path_parameter_extraction.  Even it is an empty one, it indicates CONSTANT_ADDRESS and require to do variable names to query parameter translation,  
otherwise it is not needed.

We may use a bool to replace it.

src/envoy/http/path_matcher/filter.cc
```diff
@@ -94,19 +93,14 @@ Envoy::Http::FilterHeadersStatus Filter::decodeHeaders(
...
-    std::vector<VariableBinding> variable_bindings;
-    config_->findRule(method, path, &variable_bindings);
-
-    if (!variable_bindings.empty()) {
-      const std::string query_params = VariableBindingsToQueryParameters(
-          variable_bindings, param_rule.snake_to_json_segments());
-      utils::setStringFilterState(filter_state, utils::kFilterStateQueryParams,
-                                  query_params);
-    }
+  if (!variable_bindings.empty()) {
```
qiwzhang, line 99: This is not correct.  We only do path translation for CONSTANT case.

api/envoy/v9/http/path_matcher/config.proto
TAOXUY: ACK

src/envoy/http/path_matcher/filter.cc
TAOXUY: ACK

qiwzhang (approved):

examples/testdata/path_matcher/envoy_config.json
```diff
@@ -248,8 +248,8 @@
                             }
                           },
                           {
+                            "extractPathParameters": true,
                             "operation": "1.examples_path_matcher_endpoints_cloudesf_testing_cloud_goog.ESPv2_Autogenerated_CORS_GetShelf",
-                            "pathParameterExtraction": {},
                             "pattern": {
                               "httpMethod": "OPTIONS",
                               "uriTemplate": "/libraries/{library_id}/shelves/{shelf_id}"
```
nareddyt, line 255: This should also be `shelfId`? Can you check how the uri template is propagated to these auto-generated methods? Perhaps you do the replacement too late.

nareddyt (approved):

examples/testdata/path_matcher/envoy_config.json
TAOXUY: Yes, it should be `shelfId`. 
It should be a [existing bug](https://github.com/GoogleCloudPlatform/esp-v2/blob/bd0fa867a0eefd8bec9de0db2bfe3c5253c0a448/examples/testdata/path_matcher/envoy_config.json#L259) in our code, where auto generated cors operation doesn't have snakeName to jsonName mapping.

TAOXUY: Created  b/170691898.

TAOXUY:
It is expected to break the API regression test.

PR #377, opened 2020-10-13
b/170691898: fix path extraction for auto-gen cors methods

TAOXUY:
**Description**: Currently, the auto-generated cors methods doesn't correctly set the snakeName to jsonName mapping correctly. For example, this is the original [method](https://github.com/GoogleCloudPlatform/esp-v2/blob/bd0fa867a0eefd8bec9de0db2bfe3c5253c0a448/examples/testdata/path_matcher/envoy_config.json#L216) and this is generated [method](https://github.com/GoogleCloudPlatform/esp-v2/blob/bd0fa867a0eefd8bec9de0db2bfe3c5253c0a448/examples/testdata/path_matcher/envoy_config.json#L259).Therefore, the generated path after backend routing will use snakeName, which is unknown to api producers.


TAOXUY:
Main part is ready to view. Only need to fix some unit tests.

qiwzhang (commented):
So it was not working even before moving replacing code to configmgr.  
The snake_to_json map was not generated for OPTIONS rules

qiwzhang (approved):

TAOXUY:
> So it was not working even before moving replacing code to configmgr.
> The snake_to_json map was not generated for OPTIONS rules

Correct.

src/go/configinfo/method_info.go
```diff
@@ -40,6 +40,10 @@ type methodInfo struct {
 
 	// The request type name (not the entire type URL).
 	RequestTypeName string
+
+	// The auto-generated methods, used to replace snakeName with jsonName in their
```
nareddyt, line 44: Does this need to be a slice? There should only be 1 method generated for each one.

src/go/configinfo/service_info.go
```diff
@@ -722,6 +724,12 @@ func (s *ServiceInfo) processTypes() error {
 				httpRule.UriTemplate = util.SnakeNamesToJsonNamesInPathParam(httpRule.UriTemplate, snakeToJson)
 			}
 
+			for _, gen := range mi.GeneratedCorsMethods {
+				for _, httpRule := range gen.HttpRule {
+					httpRule.UriTemplate = util.SnakeNamesToJsonNamesInPathParam(httpRule.UriTemplate, snakeToJson)
```
nareddyt, line 729: can we add a unit test to verify the uri rewrite in the CORS method?

nareddyt (approved):

TAOXUY:
The PR doesn't involve CC code so ignore the timeout presubmit-asan.

PR #378, opened 2020-10-13
b/170723966: Add option to disable OpenID Connect Discovery

nareddyt:
**Description**: Requirement from Cloud API Gateway, they do not want to make remote calls in config generator. OpenID Connect Discovery is the only place config generator will make remote calls (and we don't plan to add any other calls in the near future), so just add a flag to disable this specifically.

**Testing done**: Unit and integration test. Verified logs in integration test.

### DIFF
--- a/src/go/configmanager/flags/flags.go
+++ b/src/go/configmanager/flags/flags.go
@@ -69,6 +69,10 @@ var (
   omitted, the proxy contacts the metadata service to fetch an access token`)
 	TokenAgentPort = flag.Uint("token_agent_port", 8791, "Port that configmanager use to setup server to provide envoy with access token using service account credential, for accessing servicecontrol.")
 
+	// Flags for external calls.
+	DisableOidcDiscovery = flag.Bool("disable_oidc_discovery", false, `Disable OpenID Connect Discovery, prevents config generator from making external calls to determine the JWKS URI. 
+	This should be disabled when the URLs configured by the API Producer cannot be trusted.`)
+
 	// Envoy configurations.
 	AccessLog       = flag.String("access_log", "", "Path to a local file to which the access log entries will be written")
 	AccessLogFormat = flag.String("access_log_format", "", `String format to specify the format of access log.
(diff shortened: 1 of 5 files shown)


src/go/options/configgenerator.go
```diff
@@ -119,6 +122,7 @@ func DefaultConfigGeneratorOptions() ConfigGeneratorOptions {
 		ListenerAddress:                  "0.0.0.0",
 		ListenerPort:                     8080,
 		TokenAgentPort:                   8791,
+		DisableOidcDiscovery:             false,
```
qiwzhang, line 125: How about "EnableOidcDiscovery"?

nareddyt: We want to keep it turned on by default and give the option to disable it, right? In this case, a `--disable` flag/option works better.

qiwzhang: Yes, we want OIDC discovery to be on by default. I don't have strong preference.

src/go/configmanager/flags/flags.go
```diff
@@ -69,6 +69,10 @@ var (
   omitted, the proxy contacts the metadata service to fetch an access token`)
 	TokenAgentPort = flag.Uint("token_agent_port", 8791, "Port that configmanager use to setup server to provide envoy with access token using service account credential, for accessing servicecontrol.")
 
+	// Flags for external calls.
+	DisableOidcDiscovery = flag.Bool("disable_oidc_discovery", false, `Disable OpenID Connect Discovery, prevents config generator from making external calls to determine the JWKS URI.
```
qiwzhang, line 73: If enabled, the config generator will make external calls ...

nareddyt: Sorry, could you clarify what you mean. I clarified what disabling oidc would do.

src/go/configmanager/flags/flags.go
```diff
@@ -70,7 +70,8 @@ var (
 	TokenAgentPort = flag.Uint("token_agent_port", 8791, "Port that configmanager use to setup server to provide envoy with access token using service account credential, for accessing servicecontrol.")
 
 	// Flags for external calls.
-	DisableOidcDiscovery = flag.Bool("disable_oidc_discovery", false, `Disable OpenID Connect Discovery, prevents config generator from making external calls to determine the JWKS URI. 
+	DisableOidcDiscovery = flag.Bool("disable_oidc_discovery", false, `Disable OpenID Connect Discovery. 
+  When disabled, config generator will not make external calls to determine the JWKS URI.
```
qiwzhang, line 74: adding following
"and `x-google-jwks_uri` field in the OpenAPI spec must not be empty".

nareddyt: Done. Prefer not to refer to OpenAPI spec here, should be generic / service config based.

qiwzhang (approved):

TAOXUY (approved):

PR #381, opened 2020-10-14
b/170679093: Format with `clang-format-10`

nareddyt:
No other changes in this PR other than formatting with `clang-format-10`.


qiwzhang (changes requested):
What is the motivation to move to clang-10?

nareddyt:
> What is the motivation to move to clang-10?

Envoy no longer builds with `clang-8`. We updated to `clang-10` in #375, may as well use `clang-format-10`. It does a better job formatting includes and lambdas. Do you prefer to keep using `clang-format-7`?

qiwzhang:
Do our build scripts auto install clang-10?  or do we need to install it in our local manually?

qiwzhang (approved):

nareddyt:
> Do our build scripts auto install clang-10? or do we need to install it in our local manually?

You need to install it manually: `sudo apt install clang-format-10`

I don't want to put it in our Makefile since it requires sudo. Then you'll have to type in your password every time you run the target, even if it's already installed.

PR #382, opened 2020-10-14
Update DEVELOPER.md to install `lld-10`

nareddyt:
PR #375 changes `.bazelrc` to link with `lld-10` instead of just `lld`. ESPv2 devs need to install `lld-10` specifically, otherwise an older version of the linker may be used.


qiwzhang (approved):

nareddyt:
Skip tests, this just updates a .md file.

PR #383, opened 2020-10-14
Fix serverless janitor not cleaning up Cloud Functions

nareddyt:
Cloud Functions names are full resource ids. Example:
```
projects/cloudesf-testing/locations/us-central1/functions/e2e-test-e25fc32-bookstore--nk2r7
```

Fix the filter so it doesn't check from the start of the name.


TAOXUY (approved):

nareddyt:
Skip tests, does not affect any of them.

PR #384, opened 2020-10-16
Path rewrite filter: add envoy filter related files.

qiwzhang:
This is second PR to add path_rewrite filter.

### DIFF
--- a/src/envoy/http/backend_auth/filter.cc
+++ b/src/envoy/http/backend_auth/filter.cc
@@ -63,6 +63,8 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
                      "` is not defined by this API."),
         utils::generateRcDetails(utils::kRcDetailFilterBackendAuth,
                                  utils::kRcDetailErrorTypeUndefinedRequest));
+    decoder_callbacks_->streamInfo().setResponseFlag(
+        Envoy::StreamInfo::ResponseFlag::NoRouteFound);
     return FilterHeadersStatus::StopIteration;
   }
 
@@ -107,8 +109,6 @@ void Filter::rejectRequest(Envoy::Http::Code code, absl::string_view error_msg,
   ENVOY_LOG(debug, "{}", error_msg);
   decoder_callbacks_->sendLocalReply(code, error_msg, nullptr, absl::nullopt,
                                      details);
-  decoder_callbacks_->streamInfo().setResponseFlag(
-      Envoy::StreamInfo::ResponseFlag::UnauthorizedExternalService);
 }
 
 }  // namespace backend_auth
--- a/src/envoy/http/path_rewrite/filter.cc
+++ b/src/envoy/http/path_rewrite/filter.cc
@@ -32,6 +32,15 @@ using Envoy::Http::FilterHeadersStatus;
 using Envoy::Http::FilterTrailersStatus;
 using Envoy::Http::RequestHeaderMap;
 
+namespace {
+
+// Half of the max header value size Envoy allows.
+// 4x the standard browser request size.
+// This is need for path_matcher lookup.
+constexpr uint32_t PathMaxSize = 8192;
+
+}  // namespace
+
 FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
   if (headers.Path() == nullptr) {
     // NOTE: this shouldn't happen in practice because ServiceControl filter
@@ -42,6 +51,15 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
                                            utils::kRcDetailErrorTypeBadRequest,
                                            utils::kRcDetailErrorMissingPath));
     return FilterHeadersStatus::StopIteration;
+  } else if (headers.Path()->value().size() > PathMaxSize) {
+    config_->stats().denied_by_oversize_path_.inc();
+    rejectRequest(Envoy::Http::Code::BadRequest,
+                  absl::StrCat("Path is too long, max allowed size is ",
+                               PathMaxSize, "."),
+                  utils::generateRcDetails(utils::kRcDetailFilterPathRewrite,
+                                           utils::kRcDetailErrorTypeBadRequest,
+                                           utils::kRcDetailErrorOversizePath));
+    return Envoy::Http::FilterHeadersStatus::StopIteration;
   }
 
   absl::string_view original_path = headers.Path()->value().getStringView();
@@ -78,7 +96,8 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
       route->routeEntry()->perFilterConfigTyped<PerRouteFilterConfig>(
           kFilterName);
   if (per_route == nullptr) {
-    ENVOY_LOG(debug, "no per-route path_rewrite config");
+    ENVOY_LOG(debug,
+              "no per-route config, request is passed through unmodified");
     config_->stats().path_not_changed_.inc();
     return FilterHeadersStatus::Continue;
   }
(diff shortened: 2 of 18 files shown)


src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,119 @@
...
+    config_->stats().denied_by_invalid_path_.inc();
+    rejectRequest(
+        Envoy::Http::Code::BadRequest,
+        "Path cannot contain fragment identifier (#)",
+        utils::generateRcDetails(utils::kRcDetailFilterPathRewrite,
+                                 utils::kRcDetailErrorTypeBadRequest,
+                                 utils::kRcDetailErrorFragmentIdentifier));
+    return FilterHeadersStatus::StopIteration;
+  }
+
```
nareddyt, line 61: We should move this check from SC filter to here

https://github.com/GoogleCloudPlatform/esp-v2/blob/641ce1d5c177401e424f2b27dd45de1bf797530b/src/envoy/http/service_control/filter.cc#L62

We enforce this max length because it results in recursion in the path matcher tree. SC filter doesn't actually need the check because it doesn't build the tree, this filter does.

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,119 @@
...
+        utils::generateRcDetails(utils::kRcDetailFilterPathRewrite,
+                                 utils::kRcDetailErrorTypeUndefinedRequest));
+    return FilterHeadersStatus::StopIteration;
+  }
+
+  const auto* per_route =
+      route->routeEntry()->perFilterConfigTyped<PerRouteFilterConfig>(
+          kFilterName);
+  if (per_route == nullptr) {
+    ENVOY_LOG(debug, "no per-route path_rewrite config");
```
nareddyt, line 81: ", request is passed through unmodified"

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,119 @@
...
+namespace http_filters {
+namespace path_rewrite {
+
+using Envoy::Http::FilterDataStatus;
+using Envoy::Http::FilterHeadersStatus;
+using Envoy::Http::FilterTrailersStatus;
+using Envoy::Http::RequestHeaderMap;
+
+FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
+  if (headers.Path() == nullptr) {
```
nareddyt, line 36: Validate `:method` header is exists as well?

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  std::string new_path;
+  if (!per_route->config_parser().rewrite(original_path, new_path)) {
+    config_->stats().denied_by_url_template_mismatch_.inc();
+    rejectRequest(
+        Envoy::Http::Code::InternalServerError,
+        absl::StrCat("Request `", utils::readHeaderEntry(headers.Method()), " ",
+                     utils::readHeaderEntry(headers.Path()),
+                     "` is mismatched with url_template."),
```
nareddyt, line 93: Can we include `per_route.DebugString()` as well in the rejection message?

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,119 @@
...
+        utils::generateRcDetails(utils::kRcDetailFilterPathRewrite,
+                                 utils::kRcDetailErrorTypeUndefinedRequest));
+    return FilterHeadersStatus::StopIteration;
+  }
+
+  config_->stats().path_changed_.inc();
+  if (!headers.EnvoyOriginalPath()) {
+    headers.setEnvoyOriginalPath(headers.getPathValue());
```
nareddyt, line 101: What is this for? Does backend routing already do this, or is this new functionality?

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,119 @@
...
+  return FilterHeadersStatus::Continue;
+}
+
+void Filter::rejectRequest(Envoy::Http::Code code, absl::string_view error_msg,
+                           absl::string_view details) {
+  ENVOY_LOG(debug, "{}", error_msg);
+  decoder_callbacks_->sendLocalReply(code, error_msg, nullptr, absl::nullopt,
+                                     details);
+  decoder_callbacks_->streamInfo().setResponseFlag(
+      Envoy::StreamInfo::ResponseFlag::UnauthorizedExternalService);
```
nareddyt, line 113: Better response flag we can use? This doesn't make much sense

nareddyt (approved):

nareddyt (changes requested):
Approved by accident, sorry. Did not take a close look at the tests.

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  std::string new_path;
+  if (!per_route->config_parser().rewrite(original_path, new_path)) {
+    config_->stats().denied_by_url_template_mismatch_.inc();
+    rejectRequest(
+        Envoy::Http::Code::InternalServerError,
+        absl::StrCat("Request `", utils::readHeaderEntry(headers.Method()), " ",
+                     utils::readHeaderEntry(headers.Path()),
+                     "` is mismatched with url_template: ",
```
TAOXUY, line 93: This shouldn't happen. We use the url template to generate the route regex but now the request matched route regex doesn't match url template. It means we have bug in generating route regex(most likely). Could you add the comment here?

We shouldn't specify the url template. For downstream especially for api gateway users, they shouldn't know url template so it will confuse them. Please use other error message and also use new rc details detailed error instead of 
kRcDetailErrorTypeUndefinedRequest for differentiating for upstream.

src/envoy/http/path_rewrite/filter.cc
qiwzhang: Done

src/envoy/http/path_rewrite/filter.cc
qiwzhang: Done

src/envoy/http/path_rewrite/filter.cc
qiwzhang: We don't use method in this filter.

src/envoy/http/path_rewrite/filter.cc
qiwzhang: Done. exposed url_template() and write it in the log.  It can only due to url_template

src/envoy/http/path_rewrite/filter.cc
qiwzhang: copied from backend_routing.  if path is modified, keep the old one in x-enovy-origin-path

src/envoy/http/path_rewrite/filter.cc
qiwzhang: This is really for sanity checking if our Regex in RouteMatch works as intended. I even think about building a path_matcher_sanity envoy filter, to add method + url_template to verify it.  If it doesn't match,  reject the request to expose the problem.

qiwzhang: I see, the detail error should be in the rc_detail. will change

src/envoy/http/path_rewrite/filter.cc
qiwzhang: I could not find a better flag than this except ResponseFlag::NoRouteFound for not-route case. My take is we (service control, path-rewrite, and backend-auth) are external sever (for envoy) since we could not define our own flag

qiwzhang: OK, for backend-auth and path-rewrite, I changed it to.  not call setResponseFlag for most error, except NotRouteFound.

src/envoy/http/backend_auth/filter.cc
```diff
@@ -107,8 +109,6 @@ void Filter::rejectRequest(Envoy::Http::Code code, absl::string_view error_msg,
   ENVOY_LOG(debug, "{}", error_msg);
   decoder_callbacks_->sendLocalReply(code, error_msg, nullptr, absl::nullopt,
                                      details);
-  decoder_callbacks_->streamInfo().setResponseFlag(
```
TAOXUY, line 110: In L85, for the case `token found`, the response flag is not set.

Can we always put the `setResponseFlag` in `rejectRequest` like filters ?

qiwzhang: But we don't know which ResponseFlag to set.

nareddyt: Where do these response flags show up? Do our clients care about them? Seems like we dont need them

qiwzhang:
PTAL

nareddyt (approved):

src/envoy/http/path_rewrite/filter.cc
nareddyt: Makes sense, LGTM

TAOXUY (approved):

src/envoy/http/backend_auth/filter.cc
TAOXUY: They can be seen in the access logger.

PR #385, opened 2020-10-16
b/170497119: ordering route match#1 add uriTemplate

TAOXUY:
*Description*: import uriTemplate class from [path matcher filter](https://github.com/GoogleCloudPlatform/esp-v2/blob/641ce1d5c177401e424f2b27dd45de1bf797530b/src/api_proxy/path_matcher/http_template.h#L1). It is required to order route match sequence based on path matcher logic.

*Followup*:
- add `SequenceGenerator` under src/go/util/httppattern based on [path_matcher](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/src/api_proxy/path_matcher/path_matcher.h)
- apply the route match ordering in `RouteGenerator`

### DIFF
--- a/src/go/util/httppattern/uri_template.go
+++ b/src/go/util/httppattern/uri_template.go
@@ -1,0 +1,397 @@
+package httppattern
+
+import (
+	"bytes"
+)
+
+// UriTemplate is used to syntax pairse uri templates. It is based on the grammar
+// on https://github.com/googleapis/googleapis/blob/e5211c547d63632963f9125e2b333185d57ff8f6/google/api/http.proto#L224.
+type UriTemplate struct {
+	Segments  []string
+	Verb      string
+	Variables []*variable
+}
+
+// The info about a variable binding {variable=subpath} in the template.
+type variable struct {
+	// Specifies the range of segments [start_segment, end_segment) the
+	// variable binds to. Both start_segment and end_segment are 0 based.
+	// end_segment can also be negative, which means that the position is
+	// specified relative to the end such that -1 corresponds to the end
+	// of the path.
+	StartSegment int
+	EndSegment   int
+
+	// The path of the protobuf field the variable binds to.
+	FieldPath []string
+
+	// Do we have a ** in the variable template?
+	HasWildcardPath bool
+}
+
+// HTTP Template Grammar:
+// Questions:
+//
+// Template = "/" | "/" Segments [ Verb ] ;
+// Segments = Segment { "/" Segment } ;
+// Segment  = "*" | "**" | LITERAL | Variable ;
+// Variable = "{" FieldPath [ "=" Segments ] "}" ;
+// FieldPath = IDENT { "." IDENT } ;
+// Verb     = ":" LITERAL ;
+type parser struct {
+	input      string
+	tb         int
+	te         int
+	inVariable bool
+	segments   []string
+	verb       string
+	variables  []*variable
+}
+
+func Parse(ht string) *UriTemplate {
+	if ht == "/" {
+		return &UriTemplate{}
+	}
+
+	p := parser{
+		input: ht,
+	}
+	if !p.parse() || !p.validateParts() {
+		return nil
+	}
+
+	return &UriTemplate{
+		Segments:  p.segments,
+		Verb:      p.verb,
+		Variables: p.variables,
+	}
+}
+
+func (p *parser) parse() bool {
+	if !p.parseTemplate() || !p.consumeAllInput() {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


qiwzhang (approved):
Looking very good.  Thanks

src/go/util/httppattern/uri_template.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+const (
+	SingleParameterKey  = "/."
+	WildCardPathPartKey = "*"
+	WildCardPathKey     = "**"
+	InvalidChar         = byte(0)
+)
+
+// only constant path segments are allowed after '**'.
```
nareddyt, line 100: Are you sure about this, i think ** has to be last

https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#path-template-syntax

> The syntax ** matches zero or more URL path segments, which must be the last part of the URL path except the Verb.

src/go/util/httppattern/uri_template.go
```diff
@@ -0,0 +1,411 @@
...
+func Parse(ht string) *UriTemplate {
+	if ht == "/" {
+		return &UriTemplate{}
+	}
+
+	p := parser{
+		input: ht,
+	}
+	if !p.parse() || !p.validateParts() {
+		return nil
```
nareddyt, line 74: We need some form of error reporting. At the very least, we can just return an error:

```
return nil, fmt.Errorf("Invalid http template (%v)", ht)
```

src/go/util/httppattern/uri_template.go
```diff
@@ -0,0 +1,411 @@
...
+func (p *parser) consumeAllInput() bool {
+	return p.tb >= len(p.input)
+}
+
+func (p *parser) currentChar() byte {
+	if p.tb < p.te && p.te <= len(p.input) {
+		return p.input[p.te-1]
+	}
+
+	return InvalidChar
```
nareddyt, line 343: Callers of `currentChar()` do not handle this `InvalidChar`, is that expected? Will parsing succeed with these 0 characters in the template now?

src/go/util/httppattern/uri_template_test.go
```diff
@@ -0,0 +1,1008 @@
...
+		"/a/b{x}c/s",
+		"/a/{x}b{y}/s",
+		// No verb test cases.
+		":",
+		"/:",
+		"/a/:",
+		"/a/*:",
+		"/a/**:",
+		"/a/{b=*}/**:",
+	}
```
nareddyt, line 998: A few more cases I'd like to verify:
- /a/**/b/*/c

src/go/util/httppattern/uri_template.go
TAOXUY: Yes.

Nop, I think ""  are in the testcases.

src/go/util/httppattern/uri_template_test.go
TAOXUY: Done.

src/go/util/httppattern/uri_template.go
TAOXUY: The error info not under parsing environment will be meaningless and trivial. For example, 
- under such parsing stage, missing some components
- don't meet some expected part.

As for the `Invalid http template (%v)`, caller can add too if it is not successfully parsed.

src/go/util/httppattern/uri_template.go
TAOXUY: According to the spec, it is invalid template.

I think it should be fine as an enhancement on template syntax. In addition, it is aligned to path matcher,  for which we don't want to break our users. WDYT? @qiwzhang

qiwzhang: I know our existing c++ http_template did not following that statement: ** has to be the last part.  Either way is fine for me.  If it is easy,  we just support it.  If it brings too much work for us,  we can drop it.

nareddyt: We should be careful about deviating from the spec. If we support something undocumented now, we have to keep supporting it forever.

If we do decide to support this, I would also make sure the grpc transcoding filter supports this, and update the docs.

nareddyt (approved):

src/go/util/httppattern/uri_template.go
nareddyt: Well i guess if espv1 supports this (which is GA), we should support it to for making migration easy.

TAOXUY: ACK.

Issue #363, opened 2020-10-05
Server failed to listen: listen tcp 127.0.0.1:8790: bind: address already in use


zhaytee:
Just upgraded our ESPv1 sidecar deployment to ESPv2. Pretty excited about it, thanks for getting the sidecar mode working!

We serve multiple Cloud Endpoints services from each k8s pod. This pattern worked with ESPv1. However, ESPv2 seems to listen on a port called DiscoveryPort (default 8790). This causes a conflict with other ESPv2 instances running on the same pod:

```
Server failed to listen: listen tcp 127.0.0.1:8790: bind: address already in use
```

I'm unable to find a way to manually set that port, e.g. via a cli option or similar.

Is there a recommended path forward here? The only thing I can think of is creating a completely separate k8s Deployment for each of our Cloud Endpoints services, which would be undesirable in our particular case. So, hoping there's a better way forward.

Thank you!

TAOXUY:
Hi zhaytee,

You are right. Right now, ESPv2 uses fixed 8790 as the ads port, which is a port used by the control plane of ESPv2. We can expose a flag to use other port to fix this problem.

qiwzhang:
My understanding is: 8790 is a local port inside the ESPv2 container.  It should not conflict with other ESPv2 containers in the same pod.  Am i wrong?

TAOXUY:
The problem is that multi-containers in one pod share one IP address and same `localhost` so ESPv2 may have port conflicts with other containers. [REF](https://linchpiner.github.io/k8s-multi-container-pods.html).

qiwzhang:
I see, all containers in a pod share  same IP address and port space. Thanks for the reference. 

In this case, we need to expose a flag for users to change ads port

zhaytee:
We do plan to switch to a multi-Deployment strategy, but until we do, this could help us out a lot. Thank you!

nareddyt:
Hi @TAOXUY and @qiwzhang, a slightly different idea:

Why don't we change ADS to use unix domain sockets instead of an TCP/IP port? Then we won't introduce internal implementation detail, and it will reduce the number of flags / configuration complexity.

Currently, we use an `core.SocketAddress` in our Envoy bootstrap config. We can change that to [core.Pipe](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/address.proto#core-pipe) instead.

[...]

PR #386, opened 2020-10-16
b/171074292: Switch ads to unix domain socket

nareddyt:
**Problem**: In k8s, the network stack is shared across containers in a single pod. This can result in port conflicts, especially when ESPv2's internal ports are hardcoded. #363 is an example where a port conflict occurred due to the internal ADS port.

**Solution**: Switch ESPv2 to use a unix domain socket for ADS instead of a TCP/IP port. This will prevent port conflicts and allows the ADS traffic to be internal to the ESPv2 docker container (filesystem is not shared across containers by default).

**Implementation detail**: Use an [abstract namespace](https://utcc.utoronto.ca/~cks/space/blog/linux/SocketAbstractNamespace) so we don't have to manage the underlying inode. Switch ADS cluster to be `static`, no DNS required.

**Testing done**:
- Fix ads bootstrapper unit tests.
- Switch integration test framework to generate separate (but deterministic) pipe names for each test. Verify all tests still function.

**Follow-up PRs**: This will fix all port conflicts in the majority of cases. However, when ESPv2 is deployed with a customer service account, we will reserve TCP/IP port `8791` for the local token agent server. I need to look into the feasibility of switching this to a unix domain socket. Worst case, i'll just expose a flag to configure this. Most people don't use this, so it's not a big issue.

### DIFF
--- a/src/go/util/xds_name.go
+++ b/src/go/util/xds_name.go
@@ -67,6 +67,9 @@ const (
 	// The service control server cluster name.
 	ServiceControlClusterName = "service-control-cluster"
 
+	// The ADS cluster name.
+	AdsClusterName = "ads-cluster"
+
 	IngressListenerName  = "ingress_listener"
 	LoopbackListenerName = "loopback_listener"
 )
--- a/src/go/util/xds_name.go
+++ b/src/go/util/xds_name.go
@@ -67,9 +67,6 @@ const (
 	// The service control server cluster name.
 	ServiceControlClusterName = "service-control-cluster"
 
-	// The ADS cluster name.
-	AdsClusterName = "ads-cluster"
-
 	IngressListenerName  = "ingress_listener"
 	LoopbackListenerName = "loopback_listener"
 )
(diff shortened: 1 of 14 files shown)


src/go/util/xds_name.go
```diff
@@ -67,6 +67,9 @@ const (
 	// The service control server cluster name.
 	ServiceControlClusterName = "service-control-cluster"
 
+	// The ADS cluster name.
+	AdsClusterName = "ads-cluster"
```
qiwzhang, line 71: Where is this used?  I thought we are using Pipe name as the cluster name

nareddyt: Thanks, forgot to remove this. Originally was going to hardcode the pipe name, decided to make it a flag.

qiwzhang (approved):

TAOXUY (approved):

nareddyt:
Ignore API regression test, Cloud API Gateway doesn't use the ADS server.

PR #387, opened 2020-10-17
b/170497119: ordering route match#2 add Sort

TAOXUY:
**Problem**: There can be multiple url templates matching one url successfully. The current route-match is  executed in sequence and the first-match wins. We need to ensure the precise ones matched first and then the vague ones. 

**Details**: It is described in the [design doc](https://docs.google.com/document/d/1bLDfCpSmId7_GcGLq2UX1oB1Dtm8bL8f5pn1E-TAPJw/edit#heading=h.xgjl2srtytjt)

**Follow-up**: apply `MatchSequenceGenerator` to generate the right route matcher


TAOXUY:
PTAL, will merge the urlTemplate PR after it is submitted.

qiwzhang:
I don't understand what is the purpose of this code

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,193 @@
...
+	"strings"
+)
+
+// MatchSequenceGenerator generate the match sequence of added methods based on their
+// http pattern. The sequence should be same as the matching order of
+// https://github.com/GoogleCloudPlatform/esp-v2/blob/641ce1d5c177401e424f2b27dd45de1bf797530b/src/api_proxy/path_matcher/path_matcher.h#L1
+type MatchSequenceGenerator struct {
+	RootPtr     *matchSequenceGeneratorNode
+	CustomVerbs map[string]bool
+	Methods     []*methodData
```
qiwzhang, line 28: Methods is not needed.  In c++, it was used to own all methodData.  But in go,  objects are owned by ref_count. methodData is owned in the node, I believe.

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,193 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// MatchSequenceGenerator generate the match sequence of added methods based on their
+// http pattern. The sequence should be same as the matching order of
+// https://github.com/GoogleCloudPlatform/esp-v2/blob/641ce1d5c177401e424f2b27dd45de1bf797530b/src/api_proxy/path_matcher/path_matcher.h#L1
+type MatchSequenceGenerator struct {
```
qiwzhang, line 25: I suggest we can call it PathMatcher  or PathMetricTrie  or PathMetricTree

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,193 @@
...
+		h.CustomVerbs[ht.Verb] = true
+	}
+
+	return nil
+}
+
+type MatchSequence []*Method
+
+// Return a sorted slice of methods, used to match incoming request in sequence.
+func (h *MatchSequenceGenerator) Generate() *MatchSequence {
```
qiwzhang, line 77: s/Generate/getSortedMethods

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	h.Methods = append(h.Methods, methodData)
+	if ht.Verb != "" {
+		h.CustomVerbs[ht.Verb] = true
+	}
+
+	return nil
+}
+
+type MatchSequence []*Method
```
qiwzhang, line 74: MethodList

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Return a sorted slice of methods, used to match incoming request in sequence.
+func (h *MatchSequenceGenerator) Generate() *MatchSequence {
+	result := &MatchSequence{}
+	h.RootPtr.traverse(result)
+	return result
+
+}
+
+type variableBinding struct {
```
qiwzhang, line 84: Add a bit section divider
///////////////////////
// following lookup code are for test-only
//

src/go/util/httppattern/match_sequence_generator_node.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	return child.insertTemplate(pathParts, pathPartsIdxCur+1, httpMethod, methodData, markDuplicate)
+}
+
+func (hn *matchSequenceGeneratorNode) insertPath(pathParts []string, httpMethod string, methodData *methodData, markDuplicate bool) bool {
+	return hn.insertTemplate(pathParts, 0, httpMethod, methodData, markDuplicate)
+}
+
+func (hn *matchSequenceGeneratorNode) lookupPath(pathParts []string, pathPartsIdxCur int, httpMethod string, result *lookupResult) {
```
qiwzhang, line 70: Again, move "lookup" code to the bottom of the file with a big divider

src/go/util/httppattern/match_sequence_generator_test.go
```diff
@@ -0,0 +1,1014 @@
...
+)
+
+func parsePattern(pattern string) (string, string) {
+	s := strings.Split(pattern, " ")
+	return s[0], s[1]
+}
+
+// Though lookup method is not exposed so far, this test helps validate the
+// underlying implementation(trie insertion) of MatchSequenceGenerator.
+func TestMatchSequenceGeneratorRequestMatch(t *testing.T) {
```
qiwzhang, line 33: Separate into two test files, one for sorting and anther for lookup

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	pathInfo := transferFromUriTemplate(ht)
+	methodData := &methodData{
+		Method:   method,
+		Variable: ht.Variables,
+	}
+
+	if !h.RootPtr.insertPath(pathInfo, httpMethod, methodData, true) {
+		return fmt.Errorf("duplicate http pattern `%s %s`", httpMethod, uriTemplate)
```
qiwzhang, line 64: We should allow duplication I think.  The purpose of is object is just for sorting.  duplication should be allowed.

qiwzhang (changes requested):

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Agree, semantically it look bad to deduplicate in sorting at the first glance. I kind of prefer the current approach because 
- we can consider the non-duplication as the requirement/spec of this sorter and it is raising the error in the preprocess stage.
- it is convenient to raise this error here instead of adding extra logic to convey this duplicate information to outside and raise outside, which brings no value.

src/go/util/httppattern/match_sequence_generator_node.go
TAOXUY: Done

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Will do if we go with struct name as PathMatcher

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Will do if we go with struct name as PathMatcher

src/go/util/httppattern/match_sequence_generator.go
qiwzhang: One thing I don't like about current approach is: we are building so big that most of features are not needed.  http_method is not needed in the tree, the whole idea of the tree is to detect regex pattern overlap and generate proper order for these overlapped url_template.

Definitely, we don't want to check in and maintain dead code (not used code).  "lookup" is dead code.   http_method is not needed in the structure,  we only need to sort url_template. Ideally, only need to sort these "overlapped" ones,  don't care about the ones that not overlapped.

qiwzhang: can we do this: to simplify this tree
* only put url_template,  not http_method.  not need to store variable binding, not need to deduplicate. 
* this tree is only used for sorting,  not for lookup.  only unit-test the final sorting result.
Based on this url_template order to generate RouteMatch,  
just be care, each url_template may have multiple http_rules, since each http_rule has http_match + url

TAOXUY: ack

nareddyt: I agree this can be simplified, but we need `http_method` in the tree, right? Because we support wildcard HTTP methods, it must be used as a key and matched against.

Here is an example path config:
1. GET /a/**/c
2. *      /a/b/c

The request `GET a/b/c` should match path 1. The request `POST /a/b/c` should match path 2.

If the method is not stored in the tree, how would we determine this?

src/go/util/httppattern/match_sequence_generator.go
nareddyt: How would this work if duplication is allowed? Then you have 1 path mapping to two different operations, that should not be allowed definitely.

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,194 @@
...
+	isMultiple bool
+}
+
+func (sr *MatchSequence) appendMethod(m *Method) {
+	if m != nil {
+		*sr = append(*sr, m)
+	}
+}
+
+/////////////////////////////////////////////
```
nareddyt, line 104: Follow golang style, this will make documentation in intellij look weird

```
/*
 * lookup is for test-only
 */
```

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func NewMatchSequenceGenerator() *MatchSequenceGenerator {
+	return &MatchSequenceGenerator{
+		RootPtr:     newMatchSequenceGeneratorNode(),
+		CustomVerbs: make(map[string]bool),
+	}
+}
+
+func (h *MatchSequenceGenerator) Register(method *Method) error {
```
nareddyt, line 43: Nit: why `h` to represent MatchSequenceGenerator?

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	if ht.Verb != "" {
+		h.CustomVerbs[ht.Verb] = true
+	}
+
+	return nil
+}
+
+type MatchSequence []*Method
```
nareddyt, line 73: Nit: Prefer if all public types are defined at the top of the file. It's easy to miss this.

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,194 @@
...
+	if result.data == nil {
+		return nil, nil
+	}
+
+	return result.data.Method, extractBindingsFromPath(result.data.Variable, parts)
+}
+
+func extractRequestParts(path string, customVerbs map[string]bool) []string {
+	// Remove query parameters.
+	if idx := strings.Index(path, "?"); idx != -1 {
```
nareddyt, line 125: Can we make use of the http_template parser instead of doing this manual string manipulation?

Or i guess I'm confused about the order. When will the http template be parsed, when will this generator be used?

src/go/util/httppattern/match_sequence_generator_node.go
```diff
@@ -0,0 +1,204 @@
...
+		traverseChildren()
+		// Post-order traverse.
+		appendMethodOnCurrentNode()
+	} else {
+		appendMethodOnCurrentNode()
+		traverseChildren()
+	}
+}
+
+/////////////////////////////////////////////
```
nareddyt, line 147: Ditto

nareddyt (commented):
DId not take a close look at the `node` code yet, but looks good for trie traversal 👍

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Yes, urlTemplate deduplication + appending wildcard method in the end

Right now they are done in the lib and I guess Wayne wants to move them out of this library. It should be straightforward to be done by scanning once.

src/go/util/httppattern/match_sequence_generator_node.go
TAOXUY: `lookup` is removed.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: `lookup` is removed. So no need now.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: After checking the code, I prefer to keep the current implementation.

Deduplication out of this library still requires the segments produced by it(serialize it somehow and deduplicate in map). Plus, we also need to handle the `wildcard` method order. The current approach of sorting httpMethod + urlTemplate is relatively clean.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: How about `HttpPatternSorter`?

qiwzhang: Sorter is shorter, is better.  Not sure if we still need HttpPattern since the package name is already called that name

src/go/util/httppattern/match_sequence_generator.go
qiwzhang: I am fine. I can see the benefits of keeping the code in this shape;  it is directly ported from c++,  it should be less buggy than writing something new.

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Removes all trailing empty parts caused by extra "/".
+	for len(result) > 0 && result[len(result)-1] == "" {
+		result = result[:len(result)-1]
+	}
+
+	return result
+}
+
+func extractBindingsFromPath(vars []*variable, parts []string) []*variableBinding {
```
qiwzhang, line 134: do we need this?

src/go/util/httppattern/match_sequence_generator.go
```diff
@@ -0,0 +1,175 @@
...
+	isMultiple bool
+}
+
+func (sr *MatchSequence) appendMethod(m *Method) {
+	if m != nil {
+		*sr = append(*sr, m)
+	}
+}
+
+func extractRequestParts(path string, customVerbs map[string]bool) []string {
```
qiwzhang, line 104: do we need this function?

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done.

src/go/util/httppattern/match_sequence_generator.go
TAOXUY: Done.

TAOXUY:
PTAL

TAOXUY:
PTAL @qiwzhang @nareddyt

src/go/util/httppattern/sort_impl.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// If the current node is wildcard(**), its children has higher priority.
+	if hn.WildCard {
+		// Pre-order traverse.
+		traverseChildren()
+		// Post-order traverse.
+		appendMethodOnCurrentNode()
+	} else {
+		appendMethodOnCurrentNode()
```
qiwzhang, line 221: can we always do traverseChidlren first for all cases?

src/go/util/httppattern/sort_impl.go
```diff
@@ -0,0 +1,231 @@
...
+		// Visit vague match children after.
+		for _, child := range []*httpPatternTrieNode{singleParameterChild, wildCardPathPartChild, wildCardPathChild} {
+			if child != nil {
+				child.traverse(result)
+			}
+		}
+	}
+
+	// If the current node is wildcard(**), its children has higher priority.
+	if hn.WildCard {
```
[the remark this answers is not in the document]
nareddyt: In general, can we go through this package and clarify all comments/names that just say wildcard? It's a little confusing when it's talking about single or double.

nareddyt (approved):

src/go/util/httppattern/sort_impl.go
TAOXUY: Not quite understand your examples.

If there is wildcard underneath, it will cause problem.

For example, it will make like:
/a/**
/a

apparently, it should be
/a
/a/**

where `appendMethodOnCurrentNode()` should be called first

qiwzhang: My understanding is:  **  mean 1 or more.  so

1) /a/**
2) /a

Above order should be.   "/a" should not match 1)

TAOXUY: Refer the [spec](https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#path-template-syntax)
`The syntax ** matches zero or more URL path segments, which must be the last part of the URL path except the Verb`

qiwzhang: Well,  in actual implementation (both regex and path_matcher),  it means "1 and more".  You can try it.

"/a" will not match pattern "/a/**",    and regex for it (it has second slash). 

Well,  I just want you to justify why we need two kind of sorting order, one for double_wildcard and another for others.

TAOXUY: OK, I take back the previous example. 

How about 
/**
/**/a (not allowed in spec but we support)
/**:verb(this is allowed)

We have to put /** in the end as it is more vague.

qiwzhang: Yeap,  only in /** case, it means 0 or more. Agreed that /** should be after /.
In c++, we have special code to handle root path "/".   Do we need that here?

OK,  could you add comment on this special case to justify that order has to be this way?
Could you add test cases for this special case?

src/go/util/httppattern/sort_impl.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	traverseChildren := func() {
+		var singleParameterChild *httpPatternTrieNode
+		var wildCardPathPartChild *httpPatternTrieNode
+		var wildCardPathChild *httpPatternTrieNode
+		var exactMatchChildKey []string
+		for key, child := range hn.Children {
+			switch key {
+			case SingleParameterKey:
```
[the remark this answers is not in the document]
TAOXUY: done

src/go/util/httppattern/sort_impl.go
```diff
@@ -0,0 +1,231 @@
...
+		if wildMethodResult != nil {
+			result.appendMethod(wildMethodResult.data.Method)
+		}
+
+	}
+
+	traverseChildren := func() {
+		var singleParameterChild *httpPatternTrieNode
+		var wildCardPathPartChild *httpPatternTrieNode
+		var wildCardPathChild *httpPatternTrieNode
```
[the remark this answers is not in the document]
TAOXUY: done

src/go/util/httppattern/sort_impl.go
```diff
@@ -0,0 +1,231 @@
...
+// limitations under the License.
+
+package httppattern
+
+import (
+	"fmt"
+	"sort"
+)
+
+const HttpMethodWildCard = "*"
```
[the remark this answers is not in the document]
TAOXUY: done

TAOXUY:
Resolved all the comments. PTAL

qiwzhang (approved):

nareddyt (approved):

TAOXUY:
Th failed test is unrelated.

PR #388, opened 2020-10-17
Change configmgr to replace backend_routing with path_rewrite filter

qiwzhang:
This is the 3rd PR to use per-route config for backend_routing. 
This PR remove all backend_routing in go code
Change path_matcher filter config, not need to set extractPathParameters.

Two changes make this PR bigger
* move old TestBackendRouting code from listener_generator_test.go to router_generator_test.go with same service configs
* need to expose methodinfo in configinfo package to configgenerator  so uppercase it to MethodInfo. 

Next PR is to remove backend_routing envoy filter in c++.


src/go/configgenerator/route_generator.go
```diff
@@ -144,6 +147,63 @@ func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfigu
...
+					PathPrefix: method.BackendInfo.Path,
+				},
+			}
+		}
+	}
+	if method.BackendInfo.TranslationType == confpb.BackendRule_CONSTANT_ADDRESS {
+		constPath := &prpb.ConstantPath{
+			Path: method.BackendInfo.Path,
+		}
+		if strings.ContainsRune(httpRule.UriTemplate, '{') {
```
TAOXUY, line 168: This check can be done based on the new-added `UrlTemplate` parsing, by if the`variables` size > 0.

TAOXUY, line 168: Add a unit test case of `CONSTANT_ADDRESS` with `constPath.UrlTemplate` is empty if it is possible?

src/go/configgenerator/route_generator_test.go
```diff
@@ -308,6 +319,738 @@ func TestMakeRouteConfig(t *testing.T) {
...
+              "@type": "type.googleapis.com/espv2.api.envoy.v9.http.service_control.PerRouteFilterConfig",
+              "operationName": "testapi.bar"
+            }
+          }
+        }
+      ]
+    }
+  ]
+}`,
+		},
```
nareddyt, line 1053: Can we add a test case where a single method has multiple different HTTP rules? I would like to see multiple routes created for the method with the `path_rewrite` config in each one.

src/go/configgenerator/listener_generator.go
```diff
@@ -365,8 +358,16 @@ func makePathMatcherFilter(serviceInfo *sc.ServiceInfo) *hcmpb.HttpFilter {
 	return pathMatcherFilter
 }
 
-func hasPathParameter(httpPattern string) bool {
-	return strings.ContainsRune(httpPattern, '{')
+func needPathRewrite(serviceInfo *sc.ServiceInfo) bool {
+	for _, method := range serviceInfo.Methods {
+		// Determined by the method only. Hence just pass in the first httpRule.
```
nareddyt, line 363: Sorry, it's not clear to me why we only use the first method. 

Here's an example where i think this doesn't work: The first http rule could APPEND with empty path, so path rewrite is not needed. But the second rule may be something with a path prefix, requiring path rewrite.

If this example is correct, try to add a unit test for it.

src/go/configgenerator/route_generator.go
nareddyt: +1, recommend checking the parsed struct instead of individual characters

tests/utils/stats_fetcher.go
```diff
@@ -27,7 +27,7 @@ type StatHistograms map[string][]float64
 
 const (
 	// Path with filtering for ESPv2 stats.
-	ESpv2FiltersStatsPath = "/stats?format=json&usedonly&filter=http.ingress_http.(path_matcher|backend_auth|service_control|backend_routing)"
```
nareddyt, line 30: Do we have any stats integration tests for backend routing that need to be modified for path_rewrite?

src/go/configgenerator/route_generator.go
TAOXUY: Maybe you can add a TODO here. I can replace later when applying `UrlTemplate`.

tests/utils/stats_fetcher.go
qiwzhang: Yes, there is only on in TestStatistics.  It has been modified.

src/go/configgenerator/listener_generator.go
qiwzhang: Hmm.  I feel it is pre-mature optimization.  ConfigMgr doens't need to be fast.  So I change it to check every httpRule

src/go/configgenerator/route_generator_test.go
qiwzhang: Added

src/go/configgenerator/route_generator.go
qiwzhang: Added

qiwzhang: Ack.  UrlTemplate parsing code is in, but not hooked up yet.  I will just add a /TODO here

src/go/configgenerator/route_generator_test.go
```diff
@@ -308,6 +319,949 @@ func TestMakeRouteConfig(t *testing.T) {
...
+		{
+			desc: "Success, generate backend routing filter with allow Cors",
```
nareddyt, line 1055: Nit: Fix test descriptions (all test cases below)

nareddyt (approved):

TAOXUY (approved):

src/go/configgenerator/route_generator_test.go
qiwzhang: Done

PR #392, opened 2020-10-22
b/170497119: ordering route match#3 sort route match

TAOXUY:
**Description**: Order route match by applying `httppattern.Sort`

**Details:**
- sort methods by its http patterns during building router config
- reorder the routeConfig in the existing test cases
- remove `regexOverSize` test for building route config by `uriTemplate`(still keep the regex size test for cors origin)  ,as the `httppattern` lib enforce the [syntax](https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#path-template-syntax) so no more uri templates(before like `/**/**../**`) will cause this problem


qiwzhang (approved):

src/go/configgenerator/route_generator_test.go
```diff
@@ -1277,6 +1248,400 @@ func TestMakeRouteConfig(t *testing.T) {
...
 }`,
 		},
+		// In this test, the route configs will be in the order of
+		//		GET /foo/bar
+		//		* /foo/bar,
+		//		GET /foo/*
+		//		GET /foo/*/bar
+		//		GET /foo/**/bar
+		//		GET /foo/**:verb
+		//    GET /foo/**
```
qiwzhang, line 1258: NIT: please fit indent

qiwzhang (approved):

TAOXUY:
@nareddyt PTAL

tests/env/testdata/fake_echo_service_config_for_dynamic_routing.go
```diff
@@ -103,21 +103,15 @@ var (
 					{
 						Name: "dynamic_routing_FieldPath",
 					},
-
-					// Regression test for operation ordering: b/145520483.
-					// Config manager should not change method ordering.
-					// We test that the third method below will never be routed to.
 					{
-						Name: "operation_order_first_matched",
+						Name: "route_match_order_exact_match",
```
nareddyt, line 107: Rearrange the order in the service config to be incorrect, but double wildcard first and exact route last. Then this will truly ensure config manager will sort them correctly for the integration test.

nareddyt (approved):

tests/env/testdata/fake_echo_service_config_for_dynamic_routing.go
TAOXUY: Done.

PR #393, opened 2020-10-23
b/170423927: #3 replace snakeName with jsonName using syntax parsing

TAOXUY:
**Description**:  This PR will do syntax parsing the uri template and rebuilding it along with replacement of snakeName for jsonName, which makes more theoretically correct.

**Context**: The issue has been described in #371 and in that PR we workaround it by regex replacement. 

**Note**:  `{var}`, `{var=*}` are equal in syntax and which type expression is used is dropped during parsing and no points to stick on it. `ReplaceVariableFieldName` will all use `{var=*}`, which is also consistent with other syntaxes `{var=a/b/*}` or `{var=**}`.

**Details**:
- expose `ReplaceVariableFieldInUriTemplate` to realize this functionality and add its tests
- use it in service_info.go and remove the regex replacement related code


src/go/util/httppattern/uri_template_impl.go
```diff
@@ -0,0 +1,437 @@
...
+	inVariable bool
+	segments   []string
+	verb       string
+	variables  []*variable
+}
+
+// `serializeUriTemplate` serialize the UriTemplate object into the string representation.
+// It assumes the input is valid uriTemplate generated from `Parse()` and it
+// won't do any validation.
+func serializeUriTemplate(input *UriTemplate, fieldNameMapping map[string]string) string {
```
qiwzhang, line 61: This is not just serialize?  It is serialize + replacement.   I am wondering if it is possible 
1) doing the replacement in UrlTemplate
2) then serailize() just do a searialize.

I think the code will be cleaner

src/go/util/httppattern/uri_template_impl.go
```diff
@@ -0,0 +1,437 @@
...
+	input      string
+	tb         int
+	te         int
+	inVariable bool
+	segments   []string
+	verb       string
+	variables  []*variable
+}
+
+// `serializeUriTemplate` serialize the UriTemplate object into the string representation.
```
qiwzhang: Also can we move serialize and replacement code into a separate file.  Not to mix with url_template parsing code.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -962,13 +962,13 @@ func TestPathMatcherFilter(t *testing.T) {
 						{
 							Selector: "1.cloudesf_testing_cloud_goog.Foo",
 							Pattern: &annotationspb.HttpRule_Get{
-								Get: "foo/{foo_bar}",
```
nareddyt, line 965: Is this a breaking change? We allowed HttpRule without a prefix `/`, but now we enforce it?

src/go/configinfo/service_info.go
```diff
@@ -724,19 +724,28 @@ func (s *ServiceInfo) processTypes() error {
 			}
 		}
 
+		snakeNameToJsonNameForUriTemplates := func(m *MethodInfo, snakeNameToJsonName map[string]string) error {
+			for _, httpRule := range m.HttpRule {
+				if uriTemplate, err := httppattern.ReplaceVariableFieldName(httpRule.UriTemplate, snakeNameToJsonName); err != nil {
+					return fmt.Errorf("fail to replace snake name with json name in uri template, %v", err)
```
nareddyt, line 730: Include the the original template or operation name in the error

src/go/util/httppattern/uri_template_impl.go
```diff
@@ -0,0 +1,454 @@
+// Copyright 2019 Google LLC
```
nareddyt, line 1: Splitting into impl is c++ style, not golang style. We don't split the public interface from the implementation like this. How about change this file as `uri_template_parser.go`, and then move the 'serialize`  to `http_template.go`?

src/go/util/httppattern/uri_template_impl.go
TAOXUY: Done.

src/go/util/httppattern/uri_template_impl.go
TAOXUY: Keep parsing code into uri_template_parser.go

src/go/configgenerator/listener_generator_test.go
TAOXUY: Starting with '/' is required. 

https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#path-template-syntax

src/go/util/httppattern/uri_template.go
```diff
@@ -45,365 +47,100 @@ type variable struct {
...
-		Verb:      p.verb,
-		Variables: p.variables,
-	}
+	return uriTemplate.String(), nil
 }
 
-func (p *parser) parse() bool {
-	if !p.parseTemplate() || !p.consumeAllInput() {
-		return false
+func (u *UriTemplate) String() string {
```
qiwzhang, line 64: How about ToString()?

src/go/util/httppattern/uri_template_test.go
```diff
@@ -1000,10 +999,136 @@ func TestUriTemplateParseError(t *testing.T) {
...
-			if Parse(tc) != nil {
+			if ParseUriTemplate(tc) != nil {
 				t.Fatalf("succeed parsing %s but expect to fail", tc)
 			}
 		})
 
 	}
 }
+
+func TestReplaceVariableFieldInUriTemplateRebuild(t *testing.T) {
```
qiwzhang, line 1010: How about split into two test files,  one for url_template and another for parer.

src/go/util/httppattern/uri_template.go
TAOXUY: `String()` is the the `ToString()` in [golang](https://stackoverflow.com/questions/41232904/is-there-a-go-equivalent-of-tostring-that-fmt-print-will-use).

src/go/util/httppattern/uri_template_test.go
TAOXUY: Done.

TAOXUY:
PTAL

qiwzhang (approved):

nareddyt (approved):

PR #394, opened 2020-10-27
b/171745015:  Use syntax parsing to generate route match instead of regex

TAOXUY:
**Issue**: Right now, ESPv2 is using [regex replacement](https://github.com/GoogleCloudPlatform/esp-v2/blob/eb124fc57ef3dd26fb307634b6093286e211d7f2/src/go/util/httppattern/util.go#L45) based on uri template to get the route match, which may not cover all the uri template cases and lacks the theory proof of converting correctness.

**Solution**: Use uri template parser to parse and regenerate its regex. It is more correct in theory.

**Details**:
- Under `MethodInfo`, replace `commonpb.Pattern` with `httppattern.Pattern`, which has the complete information of the former one and functions in several places:
   - generate routeMatch when adding `OPTIONS` to deduplicate during `processHttpRules` in `service_info.go`
   - replace snakeName with jsonName during `processHttpRules` in `service_info.go`
   - sort when generating route match in `route_generator.go`
   - generate route match regex from http pattern in `route_generator.go`
- Add `Origin` under `UriTemplate` to error log original uri template.


TAOXUY:
PTAL, all the unit tests are fixed except `uri_template_test.go`.

qiwzhang (approved):

src/go/configinfo/service_info.go
```diff
@@ -380,18 +380,21 @@ func (s *ServiceInfo) processHttpRule() error {
...
-					matcher := httppattern.WildcardMatcherForPath(httpRule.UriTemplate)
-					if matcher == "" {
-						matcher = httpRule.UriTemplate
+					newHttpRule := &httppattern.Pattern{
+						HttpMethod:  util.OPTIONS,
+						UriTemplate: httppattern.ParseUriTemplate(httpRule.UriTemplate.Origin),
 					}
-					if _, exist := httpMatcherWithOptionsSet[matcher]; !exist {
-						if err := s.addOptionMethod(method, httpRule.UriTemplate); err != nil {
+					routeMatch := httpRule.UriTemplate.Regex()
```
nareddyt, line 387: Will this return exact match if there is no need for regex?

src/go/configinfo/service_info.go
```diff
@@ -419,7 +422,7 @@ func (s *ServiceInfo) processHttpRule() error {
 	return nil
 }
 
-func (s *ServiceInfo) addOptionMethod(originalMethod *MethodInfo, path string) error {
+func (s *ServiceInfo) addOptionMethod(originalMethod *MethodInfo, httpRule *httppattern.Pattern) error {
 	genOperation := fmt.Sprintf("%s.%s_CORS_%s", originalMethod.ApiName, util.AutogeneratedOperationPrefix, originalMethod.ShortName)
```
nareddyt, line 426: Maybe add a sanity check to ensure the HTTP method in the httpRule is OPTION. Worried what will happen if its called with a non-option method.

src/go/util/httppattern/sort_test.go
```diff
@@ -25,33 +25,6 @@ func parsePattern(pattern string) (string, string) {
 	return s[0], s[1]
 }
 
-func TestSortErrorHttpPattern(t *testing.T) {
```
nareddyt, line 28: Why remove?

src/go/util/httppattern/uri_template_test.go
```diff
@@ -137,10 +134,80 @@ func TestReplaceVariableFieldInUriTemplate(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.desc, func(t *testing.T) {
-			if getUriTemplate, _ := ReplaceVariableFieldInUriTemplate(tc.uriTemplate, tc.varReplace); getUriTemplate != tc.wantUriTemplate {
+			uriTemplate := ParseUriTemplate(tc.uriTemplate)
```
nareddyt, line 137: There's a lot of tests where ParseUriTemplate is being called and the result being used without checking for `nil`. It works now, but as soon as one test breaks, the tests will panic instead of failing normally.

This is why I would prefer having ParseUriTemplate return an error, as mentioned in a previous PR. Otherwise every caller would have to check for `nil`, very easy to forget. Not going to forget if you have to handle `err` explicitly.

I'm fine with doing this in a follow-up PR. Please double check all production code is checking for nil when using ParseUriTemplate.

nareddyt (approved):

src/go/configgenerator/route_generator.go
```diff
@@ -279,32 +278,26 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, erro
...
+func makeHttpRouteMatcher(httpRule *httppattern.Pattern) (*routepb.RouteMatch, error) {
 	if httpRule == nil {
 		return nil, fmt.Errorf("httpRule is nil")
 	}
 	var routeMatcher routepb.RouteMatch
 
-	regex := httppattern.WildcardMatcherForPath(httpRule.UriTemplate)
-	if regex == "" {
-		// Match with HttpHeader method. Some methods may have same path.
+	if httpRule.UriTemplate.IsExactMatch() {
```
nareddyt, line 287: Nit: If `httpRule.UriTemplate.Regex()` returns an exact match anyways, maybe no need to check and handle this case specifically.

src/go/util/httppattern/sort_test.go
TAOXUY: Because right now the sort interface is changed from the url template strings to url template objects so the invalid ones are prevented when it is constructed, mainly in `service_info.go`

src/go/configinfo/service_info.go
TAOXUY: Done.

TAOXUY:
> Why do we need this? can we check all the callers to make sure they are calling with OPTIONS?

This is a sanity check to ensure it is `OPTIONS`. 

We are calling this method with `OPTIONS` now.

src/go/util/httppattern/uri_template_test.go
TAOXUY: ACK.

PR #395, opened 2020-10-28
b/171326338: Add extra validation to filter config

nareddyt:
One customer issue that came up is due to query params being included in the backend auth JWT audience. I went through our filter configs and added more validation, especially for URI-related fields.

Follow-up PRs:
- For Cloud API Gateway usability, we also need to duplicate this validation logic into config generator. Ideally we should have the same validation reject the service config before Envoy ever sees it. As Jason suggested, the first iteration of this could include running protoc-gen-validate on the Envoy config that is outputted.


qiwzhang (approved):

nareddyt:
Accidently pushed wrong commit to this branch, reverted. Diff should be the same as when Wayne reviewed.

TAOXUY (approved):

PR #397, opened 2020-10-29
[WIP] Run protoc-gen-validate on generated envoy config

nareddyt:
TODO: Decide what the best way to test this is. Some new unit tests that run on a full config with random errors.

TODO: When synced into google3, we need to modify the BUILD rules to depend on the generated google3 validate protos. 


nareddyt:
We decided against this approach in go/espv2-config-validation-google3

PR #398, opened 2020-10-29
b/171924804:[Clean up] use error instead of nil to denote ParseUriTemplate failure

TAOXUY:
Using an associated error in function return is a golang style and it helps remind caller not to forget to check the function call failures. This PR makes `ParseUriTemplate` follow this.


qiwzhang (approved):

nareddyt (approved):

TAOXUY:
This PR only involves a go util function change and a bunch of unit tests. Passing the presubmit test should be enough.

PR #399, opened 2020-10-30
Remove white space in rc details

TAOXUY:
**Descriptions**: as pointed out by [the upstream envoy](https://screenshot.googleplex.com/8Bux9SN4R58SaDU), response code details shouldn't have white space. Inspect all the rc details in our code and find one. Fix it by replacing them with dash and underscore.


src/envoy/http/path_rewrite/filter.cc
```diff
@@ -112,13 +112,8 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
         absl::StrCat("Request `", utils::readHeaderEntry(headers.Method()), " ",
                      utils::readHeaderEntry(headers.Path()),
                      "` is getting wrong route config"),
-        utils::generateRcDetails(
-            utils::kRcDetailFilterPathRewrite,
-            utils::kRcDetailErrorTypeWrongRouteConfig,
-            absl::StrCat(
-                "Request path: ", utils::readHeaderEntry(headers.Path()),
-                " is mismatched with url_template: ",
-                per_route->config_parser().url_template())));
```
qiwzhang, line 121: We may want to keep url_template  in rc_detail. There could be multiple routes matched with a path,  we don't know which route is selected.   By printing the url_template we will know

TAOXUY: ACK

TAOXUY: Done.

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -105,6 +105,12 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
   }
 
   std::string new_path;
+
+  // This route config mismatch shouldn't happen as the current route match is
+  // regenerated and ordered in the control planeafter syntax parsing uri
+  // template in the same logic of path rewrite filter , to ensure the request
+  // is specifically matched to targeted uri template according to its grammar.
+  // It would be a bug in the route match generation if t happens.
```
qiwzhang, line 113: Just say:  It should be a bug in Envoy RouteMatch generated by config-mgr if url_template is mismatched with the request path

TAOXUY: Done.

qiwzhang (approved):

PR #401, opened 2020-11-03
Update envoy to per-route jwt_authn

qiwzhang:

This is just c++ envoy changes.  use separate PR for golang configmgr change.


src/envoy/http/service_control/handler_utils.cc
```diff
@@ -153,10 +153,10 @@ void fillLoggedHeader(
...
   }
   for (const auto& log_header : log_headers) {
-    auto* entry = headers->get(Envoy::Http::LowerCaseString(log_header));
-    if (entry) {
+    auto entry = headers->get(Envoy::Http::LowerCaseString(log_header));
+    if (!entry.empty()) {
       absl::StrAppend(&info_header_field, log_header, "=",
-                      entry->value().getStringView(), ";");
+                      entry[0]->value().getStringView(), ";");
     }
```
TAOXUY, line 160: We can call `getAllofHeaderAsString` here to make all header values be logged

src/envoy/utils/http_header_utils.cc
```diff
@@ -38,19 +38,22 @@ absl::string_view readHeaderEntry(const Envoy::Http::HeaderEntry* entry) {
 
 absl::string_view extractHeader(const Envoy::Http::HeaderMap& headers,
                                 const Envoy::Http::LowerCaseString& header) {
-  const auto* entry = headers.get(header);
-  return readHeaderEntry(entry);
+  const auto result = headers.get(header);
+  if (result.empty()) {
+    return Envoy::EMPTY_STRING;
+  }
+  return readHeaderEntry(result[0]);
```
TAOXUY, line 45: Prefer to wrap with `getAllOfHeaderAsString` to get a string list of all header values instead of only the first one.

TAOXUY (commented):
Please add 2020-11-02 in the PR title

src/envoy/utils/http_header_utils.cc
nareddyt: +1, it would match our previous behavior correct? Previously I believe all duplicate headers were concat into a single string.

nareddyt (approved):

src/envoy/http/service_control/handler_utils.cc
qiwzhang: Done

TAOXUY (approved):
LGTM.
Again, prefer to add the date 2020-11-02 in the PR title.

PR #402, opened 2020-11-03
b/171326666 #1: Add config generator option to allow startup on errors

nareddyt:
**Background**: Our Service Control and Backend Auth filter depend on external dependencies (IMDS and/or IAM) to retrieve access and ID tokens on filter config initialization. The current behavior is to block initialization (Envoy will not listen on the listener port) until ALL external dependencies are satisfied. When there are errors contacting external services, it will prevent the entire proxy from starting up.

**Problem**: Cloud API Gateway's deployment architecture does not support the same types of health checks that k8s and cloud run support. For them, blocking startup makes monitoring harder. For us, we can go with either option, but blocking startup will make our proxy easier to use in production Cloud Run deployments.

**Solution**: Add an option to config generator to make this behavior configurable. Default value ensures there is no behavior change on our side (but we can flip the value in the future if we decide to). Follow-up PR will make the changes to our filters and TokenSubscriber.

**Testing**:
- Unit tests in this PR
- Integration tests will be in the follow-up PR that has the envoy changes.


src/go/configgenerator/listener_generator_test.go
```diff
@@ -605,32 +615,68 @@ func TestBackendAuthFilter(t *testing.T) {
...
+							Address:         "https://testapipb.com/foo",
+							PathTranslation: confpb.BackendRule_CONSTANT_ADDRESS,
+							Authentication: &confpb.BackendRule_JwtAudience{
+								JwtAudience: "bar.com",
+							},
+						},
+					},
+				},
+			},
+			wantError: "unknown value for DependencyErrorBehavior",
```
TAOXUY, line 641: Please use full error for better demonstration

TAOXUY (approved):

src/go/configgenerator/listener_generator_test.go
```diff
@@ -584,6 +593,7 @@ func TestBackendAuthFilter(t *testing.T) {
    "name":"com.google.espv2.filters.http.backend_auth",
    "typedConfig":{
       "@type":"type.googleapis.com/espv2.api.envoy.v9.http.backend_auth.FilterConfig",
+		  "depErrorBehavior":"ALWAYS_INIT",
```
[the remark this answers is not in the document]
nareddyt: Done

src/go/configgenerator/listener_generator_test.go
nareddyt: Done

src/go/configgenerator/listener_generator.go
```diff
@@ -835,3 +857,11 @@ func makeRouterFilter(opts options.ConfigGeneratorOptions) *hcmpb.HttpFilter {
 	}
 	return routerFilter
 }
+
+func parseDepErrorBehavior(stringVal string) (commonpb.DependencyErrorBehavior, error) {
+	depErrorBehaviorInt, ok := commonpb.DependencyErrorBehavior_value[stringVal]
```
qiwzhang, line 862: I feel very strange to use proto generated enum directly.  prefer to use string as flag and variable,  and convert the string to proto enum when filling the proto.

nareddyt: That is what is happening, right? The flag and option are declared as string. This will parse the string option into the proto enum, which is then filled in the filter config.

src/go/configgenerator/listener_generator_test.go
```diff
@@ -605,32 +615,68 @@ func TestBackendAuthFilter(t *testing.T) {
...
+							Address:         "https://testapipb.com/foo",
+							PathTranslation: confpb.BackendRule_CONSTANT_ADDRESS,
+							Authentication: &confpb.BackendRule_JwtAudience{
+								JwtAudience: "bar.com",
+							},
+						},
+					},
+				},
+			},
+			wantError: "unknown value for DependencyErrorBehavior (UNKNOWN_ERROR_BEHAVIOR), accepted values are: map[0:UNSPECIFIED 1:BLOCK_INIT_ON_ANY_ERROR 2:ALWAYS_INIT]",
```
qiwzhang, line 641: It is not clear to:me:   should it "2:ALWAYS_INIT"  or just "ALWAYS_INIT"?

src/go/configmanager/flags/flags.go
```diff
@@ -76,6 +78,8 @@ var (
   When disabled, config generator will not make external calls to determine the JWKS URI, 
 	but the 'jwks_uri' field must not be empty in any authentication provider. 
 	This should be disabled when the URLs configured by the API Producer cannot be trusted.`)
+	DependencyErrorBehavior = flag.String("dependency_error_behavior", commonpb.DependencyErrorBehavior_BLOCK_INIT_ON_ANY_ERROR.String(),
+		`The behavior all Envoy filter will adhere to when waiting for external dependencies during filter config.`)
```
qiwzhang, line 82: Can we list out all option strings that users can use?

src/go/configgenerator/listener_generator.go
qiwzhang: I see.  You are using the proto map to to the convert.  OK, then.

src/go/configgenerator/listener_generator_test.go
nareddyt: Should be ALWAYS_INIT, fixed

src/go/configmanager/flags/flags.go
nareddyt: Sure, i'll link to the proto so it won't go out of sync.

src/go/configgenerator/listener_generator_test.go
nareddyt: Done

qiwzhang (approved):

PR #414, opened 2020-11-10
b/172866817: Fix timing issue causing ASan test to fail

nareddyt:
- Add extra sleep time to integration tests due to ASan being slow.

### DIFF
--- a/tests/integration_test/iam_imds_data_path_test.go
+++ b/tests/integration_test/iam_imds_data_path_test.go
@@ -113,6 +113,7 @@ func TestIamImdsDataPath(t *testing.T) {
 			}
 
 			if tc.wantErr != "" {
+				// When we expect a Envoy startup error, we must skip health checks. Otherwise they will prevent the test from running.
 				s.SkipHealthChecks()
 			}
 
@@ -122,6 +123,7 @@ func TestIamImdsDataPath(t *testing.T) {
 			}
 
 			if tc.wantErr != "" {
+				// When health checks are skipped (above), we need to manually sleep some time. Otherwise Envoy will not have time to try starting up.
 				glog.Infof("Sleeping to ensure Envoy is starting")
 				time.Sleep(10 * time.Second)
 			}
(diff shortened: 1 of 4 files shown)


qiwzhang (approved):

TAOXUY (approved):

nareddyt:
> Should not it be
> 
> time.Duration(tc.wantNumFails * time.Second * 4)
> 
> What does time.Duration(3) mean? 3 seconds?

No, that expression will fail to compile. `tc.wantNumFails` is an int, cannot be multiplied by `time.Second` until it's converted to a duration.

`time.Duration(3)` does not have any unit associated with it, you don't know what unit it is. But when you multiply that by ` * time.Second`, it associates a unit with the duration.

qiwzhang (approved):

nareddyt:
> Why?
> 
> In this ALWAYS_INIT mode, we should mark it ready when fetching fails.
> 
> In the new mode, START_UP_ASAP, we just don't register, so init_manager will not wait for us.

Ah right, they want a mode that will startup and suspend requests. OK, in that case I'll revert the change here, we can preserve this mode as is.

tests/integration_test/iam_imds_data_path_test.go
```diff
@@ -113,17 +113,19 @@ func TestIamImdsDataPath(t *testing.T) {
...
-				glog.Infof("Sleeping to ensure Envoy is starting")
-				time.Sleep(7 * time.Second)
 			}
 
 			defer s.TearDown(t)
 			if err := s.Setup(tc.confArgs); err != nil {
 				t.Fatalf("fail to setup test env, %v", err)
 			}
 
+			if tc.wantErr != "" {
```
qiwzhang, line 124: Don't understand why we only need to sleep for wantErr case. Could you add comment on that?

nareddyt: Done.

qiwzhang (approved):

PR #415, opened 2020-11-10
b/172671425: work around tsan false alarm on atomic race

qiwzhang:

Our tsan log has a log of false "data race".  It is due to https://github.com/google/sanitizers/issues/953


nareddyt (commented):
Might be better to set it in the .bazelrc like Envoy does. Keeps everything in sync

https://github.com/envoyproxy/envoy/blob/67609bc22f68cd3e05f5c01264a33932377955c7/.bazelrc#L97

nareddyt (approved):
Ack, nevermind about previous comment. Bazelrc will only work if we use bazel to run test.

PR #418, opened 2020-11-14
Fix transcoding_options_test

qiwzhang:

This test is not wrong,  but it is confusing.   the proto_field_name normally is snake_case, but it is defined as camelCase.  But this test reverse it, causing a lot of confusing.

In `tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto`,  field name should be `price_in_usd`, but it is `priceInUsd`.  The json name is forced to be `price_in_usd`.  Actually,  it will be auto converted to `priceInUsd`.  It is causing a lot of confusing.

This PR changes it back.


tests/integration_test/transcoding_options_test.go
```diff
@@ -73,9 +73,9 @@ func TestTranscodingPrintOptions(t *testing.T) {
 			clientProtocol:                     "http",
 			httpMethod:                         "POST",
 			method:                             "/v1/shelves/100/books?key=api-key",
-			bodyBytes:                          []byte(`{"id": 4, "price_in_usd": 100}`),
+			bodyBytes:                          []byte(`{"id": 4, "priceInUsd": 100}`),
```
nareddyt, line 76: Can the request message also be in snake_case? I assume so, just want to double-confirm

nareddyt (approved):
Thanks, much clearer

tests/integration_test/transcoding_options_test.go
qiwzhang: Yes,  the old code was using snake_case in the request JSON, it works.  This flag only control transcoding respond from proto to JSON.

TAOXUY (approved):

Issue #369, opened 2020-10-09
Allow for some clock skew in JWT Validation


someone1:
While developing, I randomly encounter errors such as `rpc error: code = Unauthenticated desc = {"code":401,"message":"Jwt not yet valid"}` - ESPv2 logs this and it seems doesn't allow for any skew.

```
Message logged: 2020-10-09 01:50:53.281 EDT (Taken from Cloud Run Logs)
JWT IAT:        2020-10-09 01:51:00 EDT
```

My PC seems to be ahead of Google's by ~7 seconds - is this beyond the allowed skew from ESP?

Google's [official guidance](https://cloud.google.com/iap/docs/signed-headers-howto) specifies allowing a 30 second skew when validating the IAT and EXP claims (this guidance is inconsistent throughout various JWT sources best practice documents).

Does ESPv2 allow for a skew when validating the JWT IAT and EXP claims? If not, can we specify one via the config, or can ESPv2 be updated to allow for some skew?

A side note but related - logs from ESPv2 don't always reflect the actual response status of the request, although I got a 401 response (gRPC), the log reflects a 200 response:

```
2020-10-09 01:50:53.281 EDT
D1009 05:50:53.280 28 envoy] [28][jwt][bazel-out/k8-opt/bin/external/envoy/source/extensions/filters/http/jwt_authn/_virtual_includes/filter_config_interface/extensions/filters/http/jwt_authn/filter_config.h:117] use filter state value <REMOVED> to find verifier.
Default
2020-10-09 01:50:53.281 EDT
D1009 05:50:53.280 28 envoy] [28][jwt][external/envoy/source/extensions/filters/http/jwt_authn/extractor.cc:188] extract authorizationBearer
Default
2020-10-09 01:50:53.281 EDT
D1009 05:50:53.280 28 envoy] [28][jwt][external/envoy/source/extensions/filters/http/jwt_authn/extractor.cc:188] extract x-goog-iap-jwt-assertion
Default
2020-10-09 01:50:53.281 EDT
D1009 05:50:53.280 28 envoy] [28][jwt][external/envoy/source/extensions/filters/http/jwt_authn/authenticator.cc:127] google_service_account: JWT authentication starts (allow_failed=false), tokens size=1
Default
2020-10-09 01:50:53.281 EDT
[...]

PR #420, opened 2020-11-17
Update envoy to  jwt clock_skew change

qiwzhang:

It is to fix: https://github.com/GoogleCloudPlatform/esp-v2/issues/369

### DIFF
--- a/tests/integration_test/jwt_auth_integration_test.go
+++ b/tests/integration_test/jwt_auth_integration_test.go
@@ -255,12 +255,12 @@ func TestAuthAllowMissing(t *testing.T) {
 			wantError:      `401 Unauthorized, {"code":401,"message":"Jwks remote fetch is failed"}`,
 		},
 		{
-			desc:           "Failed, token with wrong issuer",
+			desc:           "Successed, token with wrong issuer is considered as missing",
 			clientProtocol: "http",
 			httpMethod:     "GET",
 			method:         "/v1/shelves?key=api-key",
 			token:          testdata.Rs256Token,
-			wantError:      `401 Unauthorized, {"code":401,"message":"Jwt issuer is not configured"}`,
+			wantResp:       `{"shelves":[{"id":"100","theme":"Kids"},{"id":"200","theme":"Classic"}]}`,
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


TAOXUY (approved):

tests/integration_test/jwt_auth_integration_test.go
```diff
@@ -255,12 +255,12 @@ func TestAuthAllowMissing(t *testing.T) {
 			wantError:      `401 Unauthorized, {"code":401,"message":"Jwks remote fetch is failed"}`,
 		},
 		{
-			desc:           "Failed, token with wrong issuer",
+			desc:           "Successed, token with wrong issuer is considered as missing",
```
nareddyt, line 258: Hi @qiwzhang, was there a change in upstream envoy that changes this test? Just curious why it changed, it's not obvious.

qiwzhang: Yes, it is due to my recently submitted pr https://github.com/envoyproxy/envoy/pull/13839
If a Jwt has "iss" and "iss" doesn't matched the one specified in the `provider`,  it will return JwtUnknownIssuer.   
In this PR, JwtUnkownIssuer is treated as JwtMissing. 

This is the part I am debating,   if the spec wants to verify JWT with "iss1"  and allow_missing.  But a JWT with "iss2" comes,   should we reject it?  Is this consider as "missing"?

For now,  it is considered as "missing" and it is accepted.

nareddyt: I see, that is interesting. JWT requirement config allows both `allow_missing` and `allow_missing_or_failed`. Wouldn't the issuer mismatch be a verification failure? Then `allow_missing` should reject it, but `allow_missing_or_failed` should pass through.

https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/jwt_authn/v2alpha/config.proto#config-filter-http-jwt-authn-v2alpha-jwtrequirement

nareddyt: Well, maybe you could clarify what the term "verification" means in this documentation. Verifying the entire token (including the issuer), or just verifying the signature matches?

qiwzhang: We verify signature, "aud" and "iss".

PR #425, opened 2020-12-01
b/173810289: fix request header size

TAOXUY:
Right now, the service control filter report the size of request headers passed from upstream. It is incorrect behavior as we need the size of request headers from client and they can be modified by upstream filter like the backend auth filter.

**Fix**: use the size of request headers from downstream 

**Follow-up**: add integration tests for request/response header size

### DIFF
--- a/src/envoy/http/service_control/handler_impl.h
+++ b/src/envoy/http/service_control/handler_impl.h
@@ -117,7 +117,9 @@ class ServiceControlHandlerImpl
   std::string uuid_;
   std::string api_key_;
 
-  // The size of request headers.
+  // Considering the request headers can be modified, the original downstream
+  // header should be used as request_header_size. This variable is used to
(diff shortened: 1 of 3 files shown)


src/envoy/http/service_control/handler_impl.h
```diff
@@ -117,6 +117,9 @@ class ServiceControlHandlerImpl
   std::string uuid_;
   std::string api_key_;
 
+  // The size of request headers.
+  int request_header_size_;
```
qiwzhang, line 121: add comment on 
// If request header is modified, the original downstream header should be used as request_header_size
// this variable is used to remember the header size when HandlerImpl object is created.

TAOXUY: done

qiwzhang (approved):

nareddyt (approved):

Issue #430, opened 2020-12-03
503 error returned from Cloud Run running "gcr.io/endpoints-release/endpoints-runtime-serverless:2" as base image


mejuhi:
I have configured cloud run  and cloud functions with endpoint using the following [documentation](https://cloud.google.com/endpoints/docs/openapi/get-started-cloud-functions). In which the ESP image is running on Cloud run  

Base image used by cloud run is ```gcr.io/endpoints-release/endpoints-runtime-serverless:2``` which gets build using the following [script](https://github.com/GoogleCloudPlatform/esp-v2/blob/40caa16ea9b4231ab9272bce2c9dfcec95dee3e5/docker/serverless/gcloud_build_image) 

Cloud run is running with following configuration:

- CPU allocated: 1
- Memory allocated: 1024
- Concurrency: 80
- Request timeout: 800 seconds
- Auto-scaling:  Min instances: 5
- Auto-scaling: Max instances: 1,000

Quite recently since last week, cloud run has started to return with `503` HTTP status code and response body as `{"message":"upstream connect error or disconnect/reset before headers. reset reason: connection termination","code":503}`

Also adding full error response as seen from SoapUI:
```
HTTP/1.1 503 Service Unavailable
content-type: application/json
x-envoy-decorator-operation: ingress *********************
X-Cloud-Trace-Context: *********************;o=1
Date: Mon, 30 Nov 2020 11:28:12 GMT
Server: Google Frontend
Content-Length: 121

{"message":"upstream connect error or disconnect/reset before headers. reset reason: connection termination","code":503}
```

This error seems usual, there is no additional logs generated on cloud run side to help debug or understand what happened,  it error seems to occur random requests, also checked if there was any spike in memory or CPU utilization, but they seems normal as well.

Looking at the error, it seems like it is related to grpc error, is it related to any new changes made on the image code? Since we have started to get this 503 error from cloud run running ESPv2 quite recently.

nareddyt:
How often have you seen this occur? Did it resolve itself automatically, or did you need to deploy a new Cloud Run revision to fix it?

The error message `upstream connect error` indicates a network outage. The ESPv2 Cloud Run container could not make network connection to the backend Cloud Function. It could be a temporary problem with Cloud Run. I'm not sure what 

I suggest looking at the ESPv2 Cloud Run logs again around the timestamp you provided (`Mon, 30 Nov 2020 11:28:12 GMT`). Are there any other logs like the ones below?

```
[...]

PR #436, opened 2020-12-10
b/174860114: backend retry config options

TAOXUY:
Expose two flags `--backend_retry_ons` and `--backend_retry_num` to config retry on the backends. It is mainly based on the envoy's [RetryPolicy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-retrypolicy) and all the retryOn conditions are in [x-envoy-retry-on](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on) and [x-envoy-retry-grpc-on](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on). The default `retryOns` is `reset,connect-failure,refused-stream` and `retryNum` is `1`. 

Please note this setting will be applied to all the backends if you have multiple ones.

_Risk:_ Low as these default `retryOns` are all for TCP connection issues.

_Integration Testing:_
- under the default `backend_retry_ons`, ESPv2 will do retry on
   - the backend send RST to the connection, which causes https://github.com/GoogleCloudPlatform/esp-v2/issues/430
   - the backend hasn't been started
- the `backend_retry_ons` works
- the `backend_retry_num` works

_Follow Up:_ add backoff options


tests/endpoints/echo/server/app.go
```diff
@@ -120,6 +134,51 @@ func main() {
...
+}
+
+func (l *RstListener) Close() error {
+	return l.l.Close()
+}
+
+func RejectMiddleWare(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if *rejectRequestNum > 0 {
+			*rejectRequestNum -= 1
```
nareddyt, line 174: Do we need a mutex here? I assume it's fine, our backends don't receive parallel requests.

tests/integration_test/backend_retry_test.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			if tc.backendRetryOnsFlag != defaultBackendRetryOns {
+				args = append(args, "--backend_retry_ons="+tc.backendRetryOnsFlag)
+			}
+
+			if tc.backendRetryNumFlag != defaultBackendRetryNum {
+				args = append(args, fmt.Sprintf("--backend_retry_num=%v", tc.backendRetryNumFlag))
+			}
+
+			s := env.NewTestEnv(platform.TestBackendRetry, platform.EchoSidecar)
```
nareddyt, line 174: Do you think we should have a test for dynamic routing backends too, since remote backends is the primary use case where we need retries?

I assume the logic should be the same, so probably doesn't matter.

docker/generic/start_proxy.py
```diff
@@ -409,6 +409,25 @@ def make_argparser():
...
+        All the retryOn conditions are defined in the 
+        x-envoy-retry-on(https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on) and 
+        x-envoy-retry-grpc-on(https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on).
+        ''')
+    parser.add_argument(
+        '--backend_retry_num',
+        default=None,
+        help='''
+        The allowed number of retries. Must be >= 0 and defaults to 1. 
+        If backend_retry_ons is not set, retry_num is effectless.
```
qiwzhang, line 429: remove line 429 since we have default for "backend_retry_ons"

examples/auth/envoy_config.json
```diff
@@ -330,6 +330,10 @@
                             },
                             "route": {
                               "cluster": "backend-cluster-examples-auth.endpoints.cloudesf-testing.cloud.goog_local",
+                              "retryPolicy": {
+                                "numRetries": 1,
+                                "retryOn": "reset,connect-failure,refused-stream"
```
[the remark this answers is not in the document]
qiwzhang: Let us use default for now. we can add it later if requested.

src/go/configmanager/flags/flags.go
```diff
@@ -134,6 +134,20 @@ var (
 	TranscodingPreserveProtoFieldNames      = flag.Bool("transcoding_preserve_proto_field_names", false, "Whether to preserve proto field names for grpc-json transcoding")
 	TranscodingIgnoreQueryParameters        = flag.String("transcoding_ignore_query_parameters", "", "A list of query parameters(separated by comma) to be ignored for transcoding method mapping in grpc-json transcoding.")
 	TranscodingIgnoreUnknownQueryParameters = flag.Bool("transcoding_ignore_unknown_query_parameters", false, "Whether to ignore query parameters that cannot be mapped to a corresponding protobuf field in grpc-json transcoding.")
+
+	BackendRetryOns = flag.String("backend_retry_ons", "reset,connect-failure,refused-stream",
```
qiwzhang, line 138: How to disable retry?  set it to empty value?

docker/generic/start_proxy.py
```diff
@@ -409,6 +409,25 @@ def make_argparser():
         Set the retry times for service control Report request.
         Must be >= 0 and the default is 5 if not set.
         ''')
+    parser.add_argument(
```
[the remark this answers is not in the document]
TAOXUY: I think we get aligned in the meeting. If not, as it is the design issue, we can discuss in the design doc.

tests/endpoints/echo/server/app.go
TAOXUY: ACK. We are sending sequential requests so it should be fine. Would be easy to add if needed.

examples/auth/envoy_config.json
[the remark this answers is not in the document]
TAOXUY: ACK. We can add if needed.

tests/env/components/echo_http_server.go
```diff
@@ -27,7 +27,7 @@ type EchoHTTPServer struct {
 	*Cmd
 }
 
-func NewEchoHTTPServer(port uint16, enableHttps bool, enableRootPathHandler, useWrongCert bool, mtlsCertFile string, disableHttp2 bool) (*EchoHTTPServer, error) {
+func NewEchoHTTPServer(port uint16, enableHttps bool, enableRootPathHandler, useWrongCert bool, mtlsCertFile string, disableHttp2, backendAlwaysRespondRST bool, backendRejectRequestNum, backendRejectRequestStatus int) (*EchoHTTPServer, error) {
```
qiwzhang, line 30: too many parameters,   can we use a struct for passing them?  it will be more reabable

tests/integration_test/backend_retry_test.go
TAOXUY: Done.

docker/generic/start_proxy.py
TAOXUY: Done.

tests/env/components/echo_http_server.go
TAOXUY: Done.

src/go/configmanager/flags/flags.go
TAOXUY: It can be done by empty `RetryOns` or 0 `retryNum`. Put it in the flag explanation.

qiwzhang (approved):

tests/integration_test/backend_retry_test.go
```diff
@@ -203,10 +282,23 @@ func TestBackendRetry(t *testing.T) {
 
 			time.Sleep(time.Second * 5)
 			gotSpanNames, _ := s.FakeStackdriverServer.RetrieveSpanNames()
-			if !reflect.DeepEqual(gotSpanNames, tc.wantSpanNames) {
-				t.Errorf("got span names: %+q, want span names: %+q", gotSpanNames, tc.wantSpanNames)
+			realWantSpanNames := tc.wantSpanNames
+			if tc.backendTypeUsed == remoteBackend {
+				realWantSpanNames = replaceBackendPort(tc.wantSpanNames, strconv.Itoa(int(s.Ports().DynamicRoutingBackendPort)))
```
nareddyt, line 287: Good idea. I think you need to add the same logic to `opencensus_tracing_test.go`. Probably best to put in some utility file.

nareddyt (approved):

tests/integration_test/backend_retry_test.go
TAOXUY: You mean the port replacement part?

nareddyt: Yes, but nevermind. Turns out I added this logic there as well:

https://github.com/GoogleCloudPlatform/esp-v2/blob/9fbb88c56b3b7b7e14a43b64b77ca2d8e7d2a0e0/tests/integration_test/opencensus_tracing_test.go#L406

TAOXUY: Cool!

PR #440, opened 2020-12-12
b/174609312 #2: Handle trailing slashes for exact match paths

nareddyt:
Follow-up on #435 to accept trailing slash for all remaining paths. The trailing slash will **not** be stripped from the request unless `CONSTANT_ADDRESS` path rewrite is configured.

This is implemented by creating a duplicate route, where the only difference is the exact match path.

Testing done:
- Updated config manager unit tests
- Added new integration tests

### DIFF
--- a/src/go/configgenerator/route_generator.go
+++ b/src/go/configgenerator/route_generator.go
@@ -315,6 +315,14 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, erro
 	return backendRoutes, nil
 }
 
+func makeHttpExactPathRouteMatcher(path string) *routepb.RouteMatch {
+	return &routepb.RouteMatch{
+		PathSpecifier: &routepb.RouteMatch_Path{
+			Path: path,
+		},
+	}
+}
+
 func makeHttpRouteMatchers(httpRule *httppattern.Pattern) ([]*routepb.RouteMatch, error) {
 	if httpRule == nil {
 		return nil, fmt.Errorf("httpRule is nil")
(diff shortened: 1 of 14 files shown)


qiwzhang (approved):

src/go/util/httppattern/uri_template.go
```diff
@@ -62,7 +62,7 @@ type variable struct {
 	HasDoubleWildCard bool
 }
 
-func (u *UriTemplate) String() string {
+func (u *UriTemplate) ExactMatchString(acceptTrailingBackslash bool) string {
 	if len(u.Segments) == 0 {
 		return "/"
```
TAOXUY, line 67: Please add unit test of calling this method  on the uriTemplate "/" with true/false, which is the only one we accept with trailing slash/

src/go/configgenerator/route_generator.go
```diff
@@ -167,7 +167,7 @@ func MakePathRewriteConfig(method *configinfo.MethodInfo, httpRule *httppattern.
 		}
 
 		if uriTemplate := httpRule.UriTemplate; uriTemplate != nil && len(uriTemplate.Variables) > 0 {
-			constPath.UrlTemplate = uriTemplate.String()
+			constPath.UrlTemplate = uriTemplate.ExactMatchString(false)
```
TAOXUY, line 170: NIT: add `acceptTrailingBackslash` comment before argument for readability.

src/go/configgenerator/route_generator.go
```diff
@@ -255,100 +255,116 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, erro
...
-	var routeMatcher routepb.RouteMatch
+	var routeMatchers []*routepb.RouteMatch
 
 	if httpRule.UriTemplate.IsExactMatch() {
-		routeMatcher = routepb.RouteMatch{
-			PathSpecifier: &routepb.RouteMatch_Path{
-				Path: httpRule.UriTemplate.String(),
-			},
+		pathNoTrailingSlash := httpRule.UriTemplate.ExactMatchString(false)
+		pathWithTrailingSlash := httpRule.UriTemplate.ExactMatchString(true)
```
TAOXUY, line 334: DITTO

src/go/util/httppattern/uri_template.go
```diff
@@ -62,7 +62,7 @@ type variable struct {
 	HasDoubleWildCard bool
 }
 
-func (u *UriTemplate) String() string {
```
TAOXUY, line 65: Please check all the place using uriTemplate as Stringer interface otherwise the log may be messed up. Or implement this method with ExactMatchString(false) just in cash.

nareddyt: Nowhere else is using it, it would be a compile time error since Stringer is now longer implemented.

I reimplemented stringer as you suggested.

src/go/configgenerator/route_generator.go
nareddyt: See above. Let's keep it as a boolean for now, if we need to support more args, we can move into a struct for readability.

src/go/configgenerator/route_generator.go
nareddyt: I assume you're talking about this in C++? http://clang.llvm.org/extra/clang-tidy/checks/bugprone-argument-comment.html

This is not only for readability, it's actually checked by `clang-tidy`.

I suggest not doing this in golang:
- It's not part of the golang style guide
- No golang linter will check to ensure they're correct (especially after refactoring, they can go out of sync)

src/go/util/httppattern/uri_template.go
nareddyt: Done

TAOXUY (approved):

PR #444, opened 2020-12-16
b/175134875: Revamp status codes in access log

nareddyt:
**Key Behavior Changes:**
- For all requests, a new log entry field `http_status_code` is set.
- For gRPC requests that had no HTTP errors, a new log entry field `grpc_status_code` is set.
- For gRPC requests that had no HTTP errors, the `http_request.status` field will reflect the canonical RPC status converted into the HTTP status code range. Note: We do not support access logging non-canonical/user-defined codes.
- When HTTP status cannot be determined (ie. cancellation on client side), do **not** populate `http_request.status`. But still set log entry field `http_status_code` to 0.

**Testing Done**
- Unit tests
- Integration tests for report with gRPC backend returning `grpc-status` trailer. gRPC status is in the report.
- Integration tests for report with client cancellation. Status is not populated.

**User-facing behavior**
For gRPC backend errors:
image

For client cancellation requests:
image


src/envoy/http/service_control/handler_impl_test.cc
```diff
@@ -1063,6 +1065,80 @@ TEST_F(HandlerTest, HandlerReportWithoutCheck) {
...
+                                 expected_report_info, headers,
+                                 response_headers, response_trailers)));
+    handler.callReport(&headers, &response_headers, &response_trailers);
+  }
+};
+
+TEST_F(HandlerReportStatusTest, GrpcStatusFromTrailers) {
+  // Test: When grpc status trailer is present, it's value will be used in
+  // addition to HTTP status.
+  runTest(200, {{"content-type", "application/grpc"}}, {{"grpc-status", "7"}},
```
qiwzhang, line 1101: I did not see tests for "trailers-only" where grpc-status is in the response headers.

src/envoy/http/service_control/handler_impl_test.cc
```diff
@@ -1063,6 +1065,80 @@ TEST_F(HandlerTest, HandlerReportWithoutCheck) {
...
+  // Remember: grpc-status will show up in the headers during a trailers-only
+  // response.
+  runTest(200, {{"content-type", "application/grpc"}}, {{"grpc-status", "14"}},
+          Code::UNAVAILABLE);
+}
+
+TEST_F(HandlerReportStatusTest, MissingGrpcStatus) {
+  // Test: When grpc status header/trailer is missing, we fallback to HTTP
+  // status.
+  runTest(200, {{"content-type", "application/grpc"}}, {}, Code::OK);
```
qiwzhang, line 1117: If there is not grpc-status, I prefer not to write "grpc_status_code" into log entry.   Now, we are still writting OK,  is that rigth?

TAOXUY (approved):

src/envoy/http/service_control/handler_impl_test.cc
nareddyt: Correct, fixed now. Use `absl::optional` based on guidance from go/totw/123. This should also reduce some duplicate logic in the request builder.

Also added an integration test for non-canonical rpc code.

src/envoy/http/service_control/handler_impl_test.cc
nareddyt: Oops, I mixed up the braces in a test. Fixed.

src/api_proxy/service_control/request_builder.cc
```diff
@@ -764,6 +776,9 @@ constexpr char kLogFieldNameConfigId[] = "service_config_id";
 constexpr char kLogFieldNameTimestamp[] = "timestamp";
 constexpr char kLogFieldNameApiKeyState[] = "api_key_state";
 constexpr char kLogFieldNameResponseCodeDetail[] = "response_code_detail";
+;
```
qiwzhang, line 779: why here is extra ";" here?

nareddyt: Fixed

src/envoy/http/service_control/handler_impl_test.cc
```diff
@@ -1063,6 +1078,80 @@ TEST_F(HandlerTest, HandlerReportWithoutCheck) {
...
+  // Remember: grpc-status will show up in the headers during a trailers-only
+  // response.
+  runTest(200, {{"content-type", "application/grpc"}, {"grpc-status", "14"}},
+          {}, Code::UNAVAILABLE);
+}
+
+TEST_F(HandlerReportStatusTest, MissingGrpcStatus) {
+  // Test: When grpc status header/trailer is missing, we fallback to HTTP
+  // status.
+  runTest(200, {{"content-type", "application/grpc"}}, {}, absl::nullopt);
```
qiwzhang, line 1130: The code is not easy to read.  It is possible to do followings:

testGrpcStatus( in = { http_status_code: 200,
                                   response_headers: {{}},
                                   response_trailer_headers: 
                                 },
                          expected_grpc_status:  
                   );

nareddyt: Hmm, I cannot think of a good way to do this. If we build with c++20, we could use designated initializers (go/totw/172). But we build with c++17.

I can change it to a struct input, but then we need multiple lines to fill in the struct:
```
TestInput input;
input.http_status_code = 200;
input.response_headers = {{...}};
...
testGrpcStatus(input, Code::UNAVAILABLE)
```

What do you think?

qiwzhang: Yeap,  I don't have a good way to do that either.  So I guess it is fine then.

qiwzhang (approved):

PR #445, opened 2020-12-17
Update Envoy to v1.17.0

nareddyt:
Due to a bug in clang, I had to disable UbSan builds. See #462 for more context. We can work with Nighthawk to find a good solution.

`ESPv2-API-regression-test` test failure is incorrect, there is no API change. We updated the bazel version and some bazel flags changed. Because the test checks out an older commit, the build with old bazel flags fails.


TAOXUY:
IIRC, we were aligned with pinning to the latest envoy release(it is 12/7 for now) instead of HEAD, though I understand you wanna include the latest trace change. Can we keep following this?

Last time, it was an exception for the per-route-config change which is a GA blocker.

nareddyt:
> IIRC, we were aligned with pinning to the latest envoy release(it is 12/7 for now) instead of HEAD, though I understand you wanna include the latest trace change. Can we keep following this to make the dependency less risky?
> 
> Last time, it was an exception for the per-route-config change which is a GA blocker.

Good point, this update is only for a FR, not any blocker. I am OK with waiting, but 1-2 months until the next release is a little too much.

I see we have not been following this pattern of pinning with an envoy release. For example, #420 updates Envoy to a non-release HEAD for a FR.

What do you think @qiwzhang? Any tips on how to balance developer productivity here?

qiwzhang:
How about following rules?
In principle, we try to pin to the release envoy.  We can pin to the latest envoy with following two exceptions:
1) if latest envoy release is too old( > 2 months),
2) if we urgently need a new feature from Envoy.

qiwzhang (approved):

TAOXUY (approved):

qiwzhang (approved):

PR #446, opened 2020-12-22
b/173024112: #1 move the NOT_FOUND rejection logic to router

TAOXUY:
Currently, the `NotFound` request is rejected by the service control filter. This PR will move this into [router](https://github.com/envoyproxy/envoy/blob/master/source/common/router/router.cc#L361) by its `directResponse`.

**Gains**
- trace annotation for 404 requests
- access logging 404 requests' api method and api key
- more clear implementation structures

**Follow-ups:**
- differentiate 404 and 405


examples/auth/envoy_config.json
```diff
@@ -442,6 +442,20 @@
...
+                          },
+                          {
+                            "decorator": {
+                              "operation": "ingress UnknownMethod"
+                            },
+                            "directResponse": {
+                              "body": {
+                                "inlineString": "The request is not defined by this API."
+                              },
+                              "status": 404
```
qiwzhang, line 454: Why do we need to add such catch all?   It is done automatically by envoy router [decodeHeader](https://github.com/envoyproxy/envoy/blob/master/source/common/router/router.cc#L361)

TAOXUY: Didnt release that. Thanks!

TAOXUY:
Hi @toddbeckman-g, there are two filter stats being removed. Please check if API Gateway depends on them.

src/envoy/http/backend_auth/README.md
```diff
@@ -22,7 +22,6 @@ This filter records statistics.
  missing a token needed for the request. Two possible causes: 1) the `jwt_audience` specified in the
  route entry perFilterConfig for this filter PerRouteFilerConfig is not in the `jwt_audience_list` in
  the FilterConfig. 2) fails to fetch ID token.
-- `denied_by_no_route`: Number of API Consumer requests that are denied due to not route configurated.
```
nareddyt, line 25: Instead of deleting the stat, we can change it to `passed_through_by_no_route`. May be useful for API Gateway team. @toddbeckman-g

tests/integration_test/service_control_failed_request_report_test.go
```diff
@@ -34,7 +34,7 @@ func TestServiceControlFailedRequestReport(t *testing.T) {
 	configId := "test-config-id"
 	args := []string{"--service_config_id=" + configId,
 		"--rollout_strategy=fixed", "--suppress_envoy_headers"}
-	s := env.NewTestEnv(platform.TestServiceControlFailedRequestReport, platform.GrpcBookstoreSidecar)
+	s := env.NewTestEnv(platform.TestServiceControlAccessTokenFromIam, platform.GrpcBookstoreSidecar)
```
nareddyt, line 37: TestServiceControlFailedRequestReport

tests/integration_test/service_control_failed_request_report_test.go
```diff
@@ -61,7 +61,7 @@ func TestServiceControlFailedRequestReport(t *testing.T) {
 			clientProtocol: "http",
 			httpMethod:     "GET",
 			method:         "/noexistoperation?key=api-key",
-			httpCallError:  "404 Not Found, {\"code\":404,\"message\":\"Request `GET /noexistoperation?key=api-key` is not defined by this API.\"}",
+			httpCallError:  "404 Not Found, {\"code\":404,\"message\":\"\"}",
```
nareddyt, line 64: @qiwzhang But now that we removed the catch-all route, we don't provide the correct error message in the response body. I assume this is important because we explicitly cared about this error message in b/169095072.

I would prefer to keep the catch-all route to provide a clearer error message. It will also make the trace spans have a nice annotation.

TAOXUY: +1

src/envoy/http/path_rewrite/filter.cc
```diff
@@ -79,19 +79,11 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
...
-                     utils::readHeaderEntry(headers.Path()),
-                     "` is not defined by this API."),
-        utils::generateRcDetails(utils::kRcDetailFilterPathRewrite,
-                                 utils::kRcDetailErrorTypeUndefinedRequest));
-    decoder_callbacks_->streamInfo().setResponseFlag(
-        Envoy::StreamInfo::ResponseFlag::NoRouteFound);
-    return FilterHeadersStatus::StopIteration;
+  // If no route is matched, pass through the request and it will be rejected by
+  // router.
+  if (route == nullptr || route->routeEntry() == nullptr) {
```
qiwzhang, line 85: There are two options to return both 404 and 405.
1) handled in service_control:  if could not find a route,  check all_configured_paths,  found 405, not found 404
2) handled by router.  we need to create one route_entry for each path, to return 405,  and catch_all to return 404.

I prefer 1) for:
* we are in control of how errors are returned, 
* less number of route entries. faster for 404

TAOXUY: I still prefer to put in the router since it is more reasonable to add there in terms of filter functionality, instead of plug in our custom filters.

- *we are in control of how errors are returned*
We can still add static direct response for different NotFound requests
- *we can create one stats for each case*
No idea for now if we reject in router but on the other hand, we have a better trace annotation. 
- *less number of route entries. faster for 404*
Not real. Only the 404 and 405 requests will fall into these route entries so it won't effect the normal request handling.

toddbeckman-g (approved):

src/envoy/http/path_rewrite/filter.cc
qiwzhang: In the route direct_response error message, can we print request info, such as path?  e.g. can we do
"405: path: /foo/bar is supported in the API, but method: POST is not supported"

nareddyt: Given that we're building the direct response, we can just put the `uri_template` as the error message. Not sure if we can put the request path, but URI template should be good enough.

"Path `/book/{shelf_id}` is supported, but method `POST` is not supported"

qiwzhang: In the direct_response route,  we could not print rejected request path and method.  The method could be any of POST, DELETE, PATCH, OPTIONS  if supported one is GET.
My point is:  if we handle the rejection in service_control, we have more flexible. 
This is inline with Josh request of moving jwt failure handling into service_control.

TAOXUY: I don't think we can pull the request info in the runtime. 

For the catchall route match configuration, We can only put the `path` in it considering the defined requests have been matched before, because
- it is impossible illustrate all the http methods including the custom one
- it can help us aggregate all the methods into several ones by paths

It can be `For path /book/{shelf_id}, the http method in the current request is not supported`.

TAOXUY: `This is inline with Josh request of moving jwt failure handling into service_control.`

IMHO, it is only for the request rejected before the service control filter, which doesn't mean we need to move all the rejections in the service control.

qiwzhang: OK, I pefer 1) approach,   but not strongly.   If you two prefer 2),  I am OK.  please go for it.

Hi  @TAOXUY , is this PR for both 404 and 405, or  just 404?

TAOXUY: This is the first part and only for 404. The followup would be 405.

src/envoy/http/backend_auth/filter.cc
```diff
@@ -53,6 +53,9 @@ const Envoy::Http::LowerCaseString kXForwardedAuthorization{
 FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
   // Make sure route is calculated
   auto route = decoder_callbacks_->route();
+
+  // This shouldn't happen as the fallback wildcard route match should catch all
+  // the undefined requests.
```
qiwzhang, line 58: Then, we should remove dead code.  and use ASSERT or  NOT_REACHABLE_CODE.

tests/integration_test/service_control_failed_request_report_test.go
TAOXUY: Done

src/envoy/http/backend_auth/filter.cc
nareddyt: Yes, why did we add this code back in? I thought the original commit deletes it.

src/go/configgenerator/route_generator.go
```diff
@@ -306,15 +300,64 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, erro
...
+				Status: http.StatusNotFound,
+				Body: &corepb.DataSource{
+					Specifier: &corepb.DataSource_InlineString{
+						InlineString: `The current request is not defined by this API.`,
+					},
+				},
+			},
+		},
+		Decorator: &routepb.Decorator{
+			Operation: fmt.Sprintf("%s %s", util.SpanNamePrefix, "UndefinedMethod"),
```
nareddyt, line 356: Would prefer `ingress UnknownOperationName` to keep it consistent with our access logs. Same decorator name can be used for 405 in follow-up PR, no need to differentiate.

image

tests/integration_test/service_control_failed_request_report_test.go
```diff
@@ -81,7 +81,7 @@ func TestServiceControlFailedRequestReport(t *testing.T) {
 					Platform:             util.GCE,
 					Location:             "test-zone",
 					BackendProtocol:      "grpc",
-					ResponseCodeDetail:   "service_control_undefined_request",
+					ResponseCodeDetail:   "direct_response",
```
nareddyt, line 84: Nice, the response code detail makes more sense now 👍

src/envoy/http/backend_auth/filter.cc
TAOXUY: This is to explain why all the requests should be matched

TAOXUY:
PTAL

nareddyt (approved):

src/envoy/http/backend_auth/filter.cc
```diff
@@ -53,20 +53,10 @@ const Envoy::Http::LowerCaseString kXForwardedAuthorization{
...
-                     "` is not defined by this API."),
-        utils::generateRcDetails(utils::kRcDetailFilterBackendAuth,
-                                 utils::kRcDetailErrorTypeUndefinedRequest));
-    decoder_callbacks_->streamInfo().setResponseFlag(
-        Envoy::StreamInfo::ResponseFlag::NoRouteFound);
-    return FilterHeadersStatus::StopIteration;
-  }
+  // This shouldn't happen as the fallback wildcard route match should catch all
+  // the undefined requests.
+  ASSERT(route != nullptr && route->routeEntry() != nullptr);
```
nareddyt, line 59: I'm confused now. Won't `route` be `nullptr` and we should return `FilterHeadersStatus::Continue`? This is what you had in your first commit, but it seems like you changed it to `ASSERT`.

And actually I see a few e2e tests failed with 503. A quick look at the logs show crashes, but I cannot find the full stacktrace. Assuming it's this `ASSERT`

image

nareddyt (changes requested):

src/envoy/http/service_control/filter_stats.h
```diff
@@ -37,6 +37,7 @@ namespace service_control {
   COUNTER(denied_consumer_error)         \
   COUNTER(denied_consumer_quota)         \
   COUNTER(denied_producer_error)         \
+  COUNTER(denied_by_no_route)            \
```
qiwzhang, line 40: why do we need to add this stat here?  I don't see where it is used

src/envoy/http/backend_auth/filter.cc
```diff
@@ -54,9 +54,14 @@ FilterHeadersStatus Filter::decodeHeaders(RequestHeaderMap& headers, bool) {
   // Make sure route is calculated
   auto route = decoder_callbacks_->route();
 
-  // This shouldn't happen as the fallback wildcard route match should catch all
+  // This shouldn't happen as the catch-all route match should catch all
   // the undefined requests.
-  ASSERT(route != nullptr && route->routeEntry() != nullptr);
+  ASSERT(route != nullptr);
```
qiwzhang, line 59: Actually,  I don't know when route() return null and when route()->routeEntry() return null.   I am not sure if it is safe to assert route != null, but not to assert route->routeEntry().

This behaviors may change from upstream envoy.

To be safe:  if either is null, just return continue.


Let us

qiwzhang (approved):

nareddyt (approved):

src/envoy/http/service_control/filter.cc
```diff
@@ -62,15 +62,11 @@ Envoy::Http::FilterHeadersStatus ServiceControlFilter::decodeHeaders(
 
   // Make sure route is calculated
   auto route = decoder_callbacks_->route();
+
+  // This shouldn't happen as the catch-all route match should catch all
+  // the undefined requests.
   if (route == nullptr) {
```
nareddyt, line 68:
````
if (route == nullptr || route->routeEntry() == nullptr) {
```

To match the other filters?

src/envoy/http/backend_auth/filter.cc
TAOXUY: Here actually the problem is routeEntry is null for the catch-all route match as we don't put any per-route-config there.

src/envoy/http/service_control/filter_stats.h
TAOXUY: Done.

src/envoy/http/backend_auth/filter.cc
TAOXUY: You are right. For `routeEntry` is null for catch-all as we don't put any per-route-config there. I think it is find to `continue`.

src/envoy/http/service_control/filter.cc
TAOXUY: This logic is left to the `handler_impl` which checks its nullness when constructed and set the operation.

TAOXUY:
`ESPv2-API-regression-test` failed as intended

Details:
- TestAccessLog: we now access log the api method and api key for 404 requests while they are skipped before because rejected early.
- TestTracesDynamicRouting: the new annotation for 404 requests
- TestServiceControlFailedRequestReport: because we use the Master envoy in this test, the rejection of 404 requests is changed from service control to [path matcher](https://github.com/GoogleCloudPlatform/esp-v2/blob/ae9a8567c04137a066ee43552c2b89ec57995681/src/envoy/http/path_rewrite/filter.cc#L87).

nareddyt (approved):

PR #450, opened 2020-12-28
b/172826703: Ensure service-config related errors are actionable

nareddyt:
**Goal**: Ensures error messages are actionable. This requires providing extra context (JWT provider id, operation name, etc.) with a clear error message.

**Scope**: Let's focus on the Cloud API Gateway use-case first. The majority of user-facing errors for Cloud API Gateway will come from the service config (examples: bad JWT locations, invalid backend rule address, etc.) So this PR primarily focuses on errors related to processing the service config and creating envoy configs.

**Implementation details**:
- Majority of changes are only to error messages (no error handling behavior change).
- Some JWT errors are currently not handled. These can be fixed in a follow-up PR (b/176432170). This PR logs the errors and changes the method structure, but does not change error handling behavior.
- When calling `getOrCreateMethod` in `service_info.go`, always handle the error. This is a minor behavior change, it should be safe. Follow-up PR can cleanup usage of this method (b/176433373)


TAOXUY (approved):

src/go/configgenerator/cluster_generator.go
```diff
@@ -122,7 +122,7 @@ func addDnsResolversToClusters(dnsResolverAddresses string, clusters []*clusterp
 func makeMetadataCluster(serviceInfo *sc.ServiceInfo) (*clusterpb.Cluster, error) {
 	scheme, hostname, port, _, err := util.ParseURI(serviceInfo.Options.MetadataURL)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fail to parse metadata cluster URI: %v", err)
```
qiwzhang, line 125: Should we also print the problem URL for easy debug?

qiwzhang (changes requested):
In most cases, we only print error messages, I think we should print error url too

src/go/util/url_util.go
```diff
@@ -150,7 +150,7 @@ func IamAccessTokenPath(IamServiceAccount string) string {
 	return fmt.Sprintf("/v1/projects/-/serviceAccounts/%s:generateAccessToken", IamServiceAccount)
 }
 
-func ExtraAddressFromURI(jwksUri string) (string, error) {
+func ExtractAddressFromURI(jwksUri string) (string, error) {
```
qiwzhang, line 153: s/jwksUri/uri

nareddyt: Done

src/go/configinfo/service_info.go
```diff
@@ -244,14 +251,18 @@ func (s *ServiceInfo) addGrpcHttpRules() error {
 	for _, api := range s.serviceConfig.GetApis() {
 		for _, method := range api.GetMethods() {
 			selector := fmt.Sprintf("%s.%s", api.GetName(), method.GetName())
-			mi, _ := s.getOrCreateMethod(selector)
+			mi, err := s.getOrCreateMethod(selector)
+			if err != nil {
+				return fmt.Errorf("error processing auto-generated gRPC http rule for operation (%s.%s): %v", api.GetName(), method.GetName(), err)
```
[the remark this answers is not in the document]
nareddyt: Rephrased. I would still like to mention we are creating a gRPC http rule, that is the main point of this method.

src/go/configinfo/service_info.go
```diff
@@ -210,13 +213,16 @@ func (s *ServiceInfo) processEmptyJwksUriByOpenID() error {
...
+func (s *ServiceInfo) processApis() error {
 	for _, api := range s.serviceConfig.GetApis() {
 		s.ApiNames = append(s.ApiNames, api.Name)
 
 		for _, method := range api.GetMethods() {
 			selector := fmt.Sprintf("%s.%s", api.GetName(), method.GetName())
-			mi, _ := s.getOrCreateMethod(selector)
+			mi, err := s.getOrCreateMethod(selector)
+			if err != nil {
+				return fmt.Errorf("error processing operation (%v): %v", selector, err)
```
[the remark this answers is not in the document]
nareddyt: Done. Let's not worry too much about the error messages for `getOrCreateMethod(...)`. I will work on b/176433373 next to cleanup usage of this.

src/go/configgenerator/cluster_generator.go
nareddyt: Done. I added this logic to the shared `ParseURI` function to reduce code duplication.

qiwzhang (approved):

PR #451, opened 2020-12-28
b/173024112: #2 add 405 directResponse in router

TAOXUY:
In #466, we move the 404 rejection logic in router and in this PR, we differentiate the rejection with 404 and 405. 

**Details:**
Duplicate the defined request path in the routes so the unmatched requests because of undefined methods will fall back to these routes, which will directly respond in 405.


src/go/configgenerator/route_generator.go
```diff
@@ -396,8 +450,12 @@ func makeHttpRouteMatchers(httpRule *httppattern.Pattern) ([]*routepb.RouteMatch
 		}
 	}
 
+	var methodNotAllowedRouteMatchers []*routepb.RouteMatch
 	if httpRule.HttpMethod != httppattern.HttpMethodWildCard {
 		for _, routeMatcher := range routeMatchers {
+			methodUndefinedRouterMatcherMsg := proto.Clone(routeMatcher)
```
nareddyt, line 456: Do we need a clone? Will anything happen if we don't explicitly clone? I'd like to remove the cast to `*routepb.RouteMatch` in the next line, it seems unsafe.

tests/integration_test/service_control_failed_request_report_test.go
```diff
@@ -85,6 +85,37 @@ func TestServiceControlFailedRequestReport(t *testing.T) {
 				},
 			},
 		},
+		{
```
nareddyt, line 88: Can you also add an integration test for a regex-based route with variable binding? The code is split into 2 cases, want to cover both.

examples/auth/envoy_config.json
```diff
@@ -443,6 +443,62 @@
                               }
                             }
                           },
+                          {
+                            "decorator": {
+                              "operation": "ingress UnknownOperationName"
```
qiwzhang, line 448: can we do 
ingress UnknownHttpMethodForPath_shelves

src/go/configgenerator/route_generator.go
TAOXUY: If we don't clone, I will expect the `methodNotAllowedRouteMatcher` shouldn't share the same object with the normal `routeMatcher`, which will add `Method` restriction.

examples/auth/envoy_config.json
```diff
@@ -443,6 +443,62 @@
...
+                            },
+                            "directResponse": {
+                              "body": {
+                                "inlineString": "The current request is matched to defined url template \"/shelves\" but the http method is not allowed"
```
qiwzhang, line 452: I feel Not need to print url template in the error message.  Just
The request path is defined in the API, but http method is not supported

src/go/configgenerator/route_generator.go
```diff
@@ -49,12 +50,20 @@ func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfigu
 		Domains: []string{"*"},
 	}
 
-	// Per-selector routes for both local and remote backends.
-	brRoutes, err := makeRouteTable(serviceInfo)
+	// The router matches request by sequence of routes and the first matched one will be picked, so the order of routes matters.
```
qiwzhang, line 53: Router always selects the first matched routes, so the order is important.

src/go/configgenerator/route_generator.go
```diff
@@ -271,18 +283,42 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, erro
 			respTimeout = method.BackendInfo.Deadline
 		}
 
-		var routeMatchers []*routepb.RouteMatch
+		// The `methodNotAllowedRouteMatchers` are the route matches covers all the defined uri templates
+		// but no specific methods. As all the defined requests are matched by `routeMatchers`, the rest
+		// matched by `methodNotAllowedRouteMatchers` fall in the category of `405 Method Not Allowed`.
+		var routeMatchers, methodNotAllowedRouteMatchers []*routepb.RouteMatch
```
qiwzhang, line 289: This code seems too complicated.   I feel it will be more cleaner if we can do followings
1) use a separate pass to loop through httpPatternsMethods, to get a map of path to bool.  If bool is true,  use regex, otherwise it is exact path match
2) then use that map to generate all 405 route matches

src/go/configgenerator/route_generator.go
TAOXUY: Done.

examples/auth/envoy_config.json
TAOXUY: IMHO, it is useful for users to debug by telling them which uri template the request is mapped to.

examples/auth/envoy_config.json
nareddyt: I don't agree with this. One tracing best practice is to **not** have paths in the trace name, it will make querying difficult. Especially if there is a variable template:

```
ingress UnknownHttpMethodForPath_books/3/
ingress UnknownHttpMethodForPath_books/42/
etc.
```

`UnknownOperationName` should be OK, no need to differentiate between 404 and 405 in the trace. Remember:
- Traces are used to tell latency. There will be no latency difference between 404 and 405, so this adds little value.
- Spans have attribute labels that contain HTTP status code and path. No need to include them in the span name, its duplicate info.

TAOXUY: Good point!

nareddyt: Reference that explains my first point better: https://docs.wavefront.com/tracing_best_practices.html#span-names-best-practices

> A poor choice for the span name is getShoppingMenu_123. If you incorporate the menu ID detail directly into the span names, you might end up with unique span names like getShoppingMenu_122, getShoppingMenu_123, getShoppingMenu_124, … getShoppingMenu_nnn, which all represent calls to the same piece of code. The result is a cardinality explosion!

TAOXUY: Done

src/go/configgenerator/route_generator.go
TAOXUY: IMHO, I feel it is better to put all the making routeMatch logic in one place, since the `methodNotAllowedRouteMatchers` are basically associated with `routeMatchers`.  It will be less easy to introduce bug later when we add more logic for routeMatchers.

tests/integration_test/service_control_failed_request_report_test.go
TAOXUY: Done.

examples/auth/envoy_config.json
TAOXUY: Actually I don't fully agree that.
- `Traces are used to tell latency` only. Here it can help provide debug information. 
- It is not duplicate info. For the regex, the trace space doesn't have the matched url template in the labels.

The example doesn't fit our case. The example is more about adding dynamic data from request into the span name while we are adding the url templates from service config. There won't be cardinality explosion since at most, the amount of span is just doubled.

TAOXUY:
PTAL. Only the unit tests under configmanager_test are not resolved.

examples/grpc_dynamic_routing/envoy_config.json
```diff
@@ -1370,6 +1370,202 @@
...
+                            },
+                            "directResponse": {
+                              "body": {
+                                "inlineString": "The current request is matched to the defined url template \"/echoreport\" but its http method is not allowed"
+                              },
+                              "status": 405
+                            },
+                            "match": {
+                              "path": "/echoreport/"
+                            }
```
TAOXUY, line 1427: Please note, for the exactMatch, we will add additional one with trailing slash for backward compatibility. Here, we use the user defined one(no trailing slash) for the both the original one and generated one, as they only know what they defined.

examples/auth/envoy_config.json
qiwzhang: SG

qiwzhang:
I feel this is really overkill. I am questioning the important of distinguishing 404 vs 405.  With current implementation, I am very concerns about:
1) added number of route matches, especially, we are adding two routes for "/foo" and "/foo/"
2) added so much code, just for this small feature.  It makes our route_generator code way too complicated. 

IMO,  It is not worth it.  ESPv1 did not do that.  We don't have customers asking for this.   It is not worth such effort to do this.

TAOXUY:
> I feel this is really overkill. I am questioning the important of distinguishing 404 vs 405. With current implementation, I am very concerns about:
> 
> 1. added number of route matches, especially, we are adding two routes for "/foo" and "/foo/"
> 2. added so much code, just for this small feature.  It makes our route_generator code way too complicated.
> 
> IMO, It is not worth it. ESPv1 did not do that. We don't have customers asking for this. It is not worth such effort to do this.

-  For added number of route matches,
    - They won't affect performance as they act as fallback match after the regular route matches
    - May not be many routes are added as only the unique url templates will be added 
- For added so much code,
    - most of code are from json format indentation
    - some are added 405 routes but they are very short as they have nothing but match and directResponse info
    - the code in `route_generator` is very limited

examples/auth/envoy_config.json
nareddyt: Good catch, the cardinality problem does not exist.

I still feel it's unnecessary. I don't see how adding the uri template to a span will help debug a 405. 405 means the method is wrong, the path is just extra noise.

I'm OK with keeping it if you think it's helpful. But please add some logic (and test) to truncate the string to 128 characters. Stackdriver will reject spans with names over 128 characters. This is another reason why I would prefer to keep the names simple if possible.

tests/integration_test/service_control_failed_request_report_test.go
```diff
@@ -55,29 +54,60 @@ func TestServiceControlFailedRequestReport(t *testing.T) {
 		httpCallError  string
 		wantScRequests []interface{}
 	}{
+		//{
```
nareddyt, line 57: Uncomment other tests

TAOXUY: Done.

examples/auth/envoy_config.json
TAOXUY: Done. 

Not sure if OP config has restriction on the `ShortName`s(the normal routes' span names) and if we also need to add the length check for them. Add a TODO so far.

nareddyt (approved):

src/go/configgenerator/route_generator.go
```diff
@@ -367,21 +415,25 @@ func makeHttpExactPathRouteMatcher(path string) *routepb.RouteMatch {
...
+		return nil, nil, fmt.Errorf("httpRule is nil")
 	}
 	var routeMatchers []*routepb.RouteMatch
+	var uriTemplates []string
 
 	if httpRule.UriTemplate.IsExactMatch() {
 		pathNoTrailingSlash := httpRule.UriTemplate.ExactMatchString(false)
 		pathWithTrailingSlash := httpRule.UriTemplate.ExactMatchString(true)
 
+		uriTemplates = append(uriTemplates, pathNoTrailingSlash)
```
qiwzhang, line 429: This is very buggy code, it relies on uriTemplate array align with routeMatchers array.   Please refractory this.

tests/integration_test/service_control_failed_request_report_test.go
```diff
@@ -85,6 +85,68 @@ func TestServiceControlFailedRequestReport(t *testing.T) {
...
+				},
+			},
+		},
+		{
+			desc:           "Request matches uri template(regex) but not method.",
+			url:            fmt.Sprintf("localhost:%v", s.Ports().ListenerPort),
+			clientProtocol: "http",
+			// "DELETE" is not defined for "/v1/shelves".
+			httpMethod:    "POST",
+			method:        "/v1/shelves/100?key=api-key",
```
qiwzhang, line 125: could we add two more cases for trailing_slash, one is exact match and another regex?

src/go/configgenerator/route_generator.go
```diff
@@ -330,13 +351,40 @@ func makeRoute(routeMatcher *routepb.RouteMatch, method *configinfo.MethodInfo,
...
 		},
 	}
 }
 
-func makeCatchAllUnmatchedRoute() *routepb.Route {
+func makeMethodNotAllowedRoute(methodNotAllowedRouteMatcher *routepb.RouteMatch, uriTemplateInSc string) *routepb.Route {
+	spanName := fmt.Sprintf("%s UnknownHttpMethodForPath_%s", util.SpanNamePrefix, uriTemplateInSc)
+
+	if len(spanName) > util.SpanNameMaxByteNum {
+		newSpanName := fmt.Sprintf("%s UnknownHttpMethod", util.SpanNamePrefix)
```
qiwzhang, line 365: How about just truncate the string to MaxByte and end it with "..." to indicate it has been truncated.

qiwzhang:
Well,  either way is fine.  Since we have been progressed so far now,  don't want to waste the effort.  So let us try to get this in then.

src/go/configgenerator/route_generator.go
TAOXUY: Done by wrapping them into a struct

tests/integration_test/service_control_failed_request_report_test.go
TAOXUY: Done.

src/go/configgenerator/route_generator.go
TAOXUY: Done

TAOXUY:
PTAL

nareddyt (approved):

qiwzhang (approved):

PR #452, opened 2020-12-30
Upgrade golang to latest stable(1.15.6)

TAOXUY:
Upgrade golang to latest stable 1.15 from 1.13 so that
- resolve the library difference on different golang sdk, ex. quotation mark for [the url in the error](https://github.com/GoogleCloudPlatform/esp-v2/blob/e16dc8bc9830ebbb6333035613d23775cab0276e/src/go/util/url_util_test.go#L202) returned by `net/url`
- align with our [developer guide](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/DEVELOPER.md#install-golang)

**Details**:
- Add SANs in ssl files because golang client need SANs in ssl certificate from 1.15.  For details, see [1](http://www.inanzzz.com/index.php/post/jo4y/using-tls-ssl-certificates-for-grpc-client-and-server-communications-in-golang-updated). It won't effect ESPv2's SSL behavior as it is only for the golang client side.


nareddyt (approved):

TAOXUY:
API regression test failed as intended. It uses the old golang code while the golang version in the base image has been updated.

qiwzhang (approved):

PR #454, opened 2021-01-05
b/176433373: Cleanup usages of `getOrCreateMethod`

nareddyt:
`getOrCreateMethod` is called from a few aspects where selectors are processed (http rules, usage rules, authentication rules, etc). However, these should only GET, not CREATE. If the selector doesn't exist, it should error instead of creating a new one. With this PR, `getOrCreateMethod` will only be used for:
- The initial processing of the API methods
- Auto-generated CORS operations
- Auto-generated HealthCheck operation

Also cleanup usages of directly accessing `Methods[selector]`. Some places do not check for `nil`. This can cause a crash in config generator if the service config has selectors with typos.

Diff is large because of test changes, feel free to ignore them. They are just normalizing the service config.

### DIFF
--- a/src/go/configinfo/service_info.go
+++ b/src/go/configinfo/service_info.go
@@ -253,9 +253,9 @@ func (s *ServiceInfo) addGrpcHttpRules() error {
 	for _, api := range s.serviceConfig.GetApis() {
 		for _, method := range api.GetMethods() {
 			selector := fmt.Sprintf("%s.%s", api.GetName(), method.GetName())
-			mi, err := s.getOrCreateMethod(selector)
+			mi, err := s.getMethod(selector)
 			if err != nil {
-				return fmt.Errorf("error creating auto-generated gRPC http rule for operation (%s.%s): %v", api.GetName(), method.GetName(), err)
+				return fmt.Errorf("error processing auto-generated gRPC http rule: %v", err)
 			}
 
 			path := fmt.Sprintf("/%s/%s", api.GetName(), method.GetName())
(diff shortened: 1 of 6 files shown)


TAOXUY:
Please fix the integration tests. I remember we have several test service configs only define httpRules but no apis .

nareddyt:
> Please fix the integration tests. I remember we have several test service configs only define httpRules but no apis .

Thanks, I thought they were already normalized but they are missing a few. I'll try to split this into another PR, there are a few things to clean up there.

src/go/configinfo/service_info.go
```diff
@@ -382,9 +391,9 @@ func (s *ServiceInfo) processHttpRule() error {
 	addedRouteMatchWithOptionsSet := make(map[string]bool)
 
 	for _, rule := range s.ServiceConfig().GetHttp().GetRules() {
-		method, err := s.getOrCreateMethod(rule.GetSelector())
+		method, err := s.getMethod(rule.GetSelector())
```
qiwzhang, line 394: I think we need getOrCreate here.

src/go/configinfo/service_info.go
```diff
@@ -251,9 +253,9 @@ func (s *ServiceInfo) addGrpcHttpRules() error {
 	for _, api := range s.serviceConfig.GetApis() {
 		for _, method := range api.GetMethods() {
 			selector := fmt.Sprintf("%s.%s", api.GetName(), method.GetName())
-			mi, err := s.getOrCreateMethod(selector)
+			mi, err := s.getMethod(selector)
```
qiwzhang, line 256: I think we need getOrCreate here.

qiwzhang (approved):

TAOXUY (approved):

nareddyt:
Had to force push because I split out some commits into separate PR #456

qiwzhang (approved):

PR #457, opened 2021-01-07
b/176912395: Automatically configure stream idle timeouts

### LINKED ISSUE #455: Configure the stream idle timeout
Feature request : Be able to configure the Envoy stream idle timeout :
https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#stream-timeouts

It appears that ESPv2 answers with an error after 5 minutes because of the default stream idle timeout of Envoy :
> The default stream idle timeout is 5 minutes.

Answer is :
```json
{
    "code": 408,
    "message": "stream timeout"
}
```

For now, we can only configure the backend deadline wich is something else :
https://cloud.google.com/endpoints/docs/openapi/openapi-extensions#deadline

Could we have a flag to be able to configure the stream idle timeout ?

```json
- name: esp
  image: gcr.io/endpoints-release/endpoints-runtime:2
  args: [
    "--stream-idle-timeout=1800s"
  ]
```

nareddyt:
**Background**: See discussion in #455 

**Implementation**: When `deadline` is set in a backend rule, override the `stream_idle_timeout` for that operation to match the deadline. Created a flag `--stream_idle_timeout` to hold the default value and increase testability, but the flag is **not exposed** to the API producer.

The current solution fixes both cases:
1) Non-streaming HTTP requests over 5m will be supported by increasing the `deadline`. HTTP 408 will never be returned. Only HTTP 504 will occur once the configured deadline is met. 
2) For non-unary gRPC methods, user can configure the `deadline`. We disable request timeouts for streaming methods. Instead, `deadline` will configure stream idle time. So there is no need for the API Producer to set a flag.

Fixes #455

### DIFF
--- a/examples/grpc_dynamic_routing/envoy_config.json
+++ b/examples/grpc_dynamic_routing/envoy_config.json
@@ -908,7 +908,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -942,7 +942,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -976,7 +976,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1010,7 +1010,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1044,7 +1044,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1082,7 +1082,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1120,7 +1120,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1154,7 +1154,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1188,7 +1188,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1222,7 +1222,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1256,7 +1256,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
@@ -1290,7 +1290,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "400s",
+                              "idleTimeout": "401s",
                               "retryPolicy": {
                                 "numRetries": 1,
                                 "retryOn": "reset,connect-failure,refused-stream"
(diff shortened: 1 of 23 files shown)


src/go/configgenerator/route_generator.go
```diff
@@ -336,15 +326,26 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, []*r
 	return backendRoutes, methodNotAllowedRoutes, nil
 }
 
-func makeRoute(routeMatcher *routepb.RouteMatch, method *configinfo.MethodInfo, respTimeout time.Duration) *routepb.Route {
+func makeRoute(routeMatcher *routepb.RouteMatch, method *configinfo.MethodInfo) *routepb.Route {
```
nareddyt, line 329: Up to line 399 is just moving some code into the function and cleaning up the function parameter. No behavior change.

tests/integration_test/http1_deadline_test.go
```diff
@@ -180,3 +184,78 @@ func TestDeadlinesForLocalBackend(t *testing.T) {
...
+			reqDuration:    time.Second * 10,
+			deadlineToTest: Default,
+		},
+		{
+			desc: "When deadline is specified, deadline overrides global idle timeout flag. Request still succeeds.",
+			confArgs: append([]string{
+				"--stream_idle_timeout=1s",
+			}, utils.CommonArgs()...),
+			reqDuration:    time.Second * 2,
+			deadlineToTest: Short,
```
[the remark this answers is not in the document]
nareddyt: Done

src/go/configinfo/service_info.go
```diff
@@ -606,10 +611,14 @@ func (s *ServiceInfo) processLocalBackendOperations() error {
 			continue
 		}
 
+		// Add an extra second to ensure errors are 504s, not 408s.
+		idleTimeout := util.MaxDuration(util.DefaultResponseDeadline+time.Second, s.Options.StreamIdleTimeout)
```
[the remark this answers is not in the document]
nareddyt: Done

src/go/configgenerator/route_generator.go
```diff
@@ -336,15 +326,26 @@ func makeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, []*r
...
+	// Response timeouts are not compatible with streaming methods (documented in Envoy).
```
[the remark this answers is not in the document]
nareddyt: From Envoy's perspective, `deadline` is **response** timeout. It is the amount of time that Envoy will wait for the upstream to respond after receiving a downstream request.

src/go/configinfo/service_info.go
```diff
@@ -556,12 +556,17 @@ func (s *ServiceInfo) addBackendInfoToMethod(r *confpb.BackendRule, scheme strin
 		deadline = time.Duration(deadlineMs) * time.Millisecond
 	}
 
+	// Allow per-route deadlines to override the global flag.
+	// Add an extra second to ensure errors are 504s, not 408s.
+	idleTimeout := util.MaxDuration(deadline+time.Second, s.Options.StreamIdleTimeout)
```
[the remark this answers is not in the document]
nareddyt: Done, does the new comment help?

src/go/configmanager/flags/flags.go
```diff
@@ -46,6 +46,8 @@ var (
 
 	// Envoy specific configurations.
 	ClusterConnectTimeout = flag.Duration("cluster_connect_timeout", 20*time.Second, "cluster connect timeout in seconds")
+	StreamIdleTimeout     = flag.Duration("stream_idle_timeout", util.DefaultIdleTimeout, "The amount of time client connections can exist without any activity. "+
```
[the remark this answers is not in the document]
nareddyt: Done. Keep in mind that Envoy treats HTTP/1.x as HTTP/2 internally.

src/go/configgenerator/route_generator.go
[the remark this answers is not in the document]
qiwzhang: I see

tests/integration_test/grpc_deadline_test.go
```diff
@@ -190,7 +192,121 @@ plans {
...
+    request {
+      text: "Hello, world!"
+      response_delay: 20
+    }
+    count: 1
+  }
+}`,
+		},
+		{
+			desc: "When deadline is NOT specified for method, a low stream idle timeout specified via flag is not honored and the request succeeds.",
```
TAOXUY, line 270: Be specific: a stream idle timeout lower than default response timeout(15s)?

tests/integration_test/http1_deadline_test.go
```diff
@@ -180,3 +186,78 @@ func TestDeadlinesForLocalBackend(t *testing.T) {
...
+		desc           string
+		confArgs       []string
+		reqDuration    time.Duration
+		deadlineToTest ConfiguredDeadline
+		wantErr        string
+	}{
+		// Please be cautious about adding too many time-based tests here.
+		// This can slow down our CI system if we sleep for too long.
+		{
+			desc: "When deadline is NOT specified, ESPv2 does not honor the idle timeout flag if the value is too low. Request still succeeds.",
```
TAOXUY, line 204: ditto

TAOXUY (approved):

tests/integration_test/grpc_deadline_test.go
```diff
@@ -190,7 +192,121 @@ plans {
...
+	}
+}
+
+func TestIdleTimeoutsForGrpcStreaming(t *testing.T) {
+	t.Parallel()
+
+	testData := []struct {
+		desc        string
+		confArgs    []string
+		addDeadline time.Duration
```
qiwzhang, line 223: s/addDeadline/perRouteDeadline  or methodDeadline

tests/integration_test/grpc_deadline_test.go
```diff
@@ -190,7 +192,121 @@ plans {
...
+      response_delay: 20
+    }
+    count: 1
+  }
+}`,
+		},
+		{
+			desc: "When deadline is NOT specified for method, a low stream idle timeout specified via flag is not honored and the request succeeds.",
+			confArgs: append([]string{
+				"--stream_idle_timeout=3s",
```
qiwzhang, line 272: This is confusing.  Naturally, people will expect this flag to apply.  

How about this:
can we distinguish between flag specified, and not-specified.  if not specified,  we want to be smart, to allow long timeout request.  
if specified,  we should honor it, always apply it

src/go/configinfo/service_info.go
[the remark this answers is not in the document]
qiwzhang: Yes, thanks

tests/integration_test/grpc_deadline_test.go
nareddyt: Done

tests/integration_test/http1_deadline_test.go
nareddyt: Done

tests/integration_test/grpc_deadline_test.go
nareddyt: Done

tests/integration_test/grpc_deadline_test.go
nareddyt: Ok, I added a few more tests and changed the behavior a little. A few points below:

- We are not exposing the flag to the user. It is just for integration tests so we can set the default stream idle timeout. Otherwise Envoy will default to `5m`, that will be hard to integration test.
- This means the only way for the user to modify the deadlines is via service config. This will modify both the response deadline and the stream idle timeout. So let's not worry too much about the flag, only about the `deadline` from service config.

The new behavior is split into 2 cases now:
- For HTTP and unary gRPC: response timeout is set from `deadline`. Stream idle timeout does not matter. It will always be set to something higher than the user specified deadline.
- For non-unary gRPC: response timeout is disabled, it does not matter. Instead, the stream idle timeout will be set equal to the `deadline`. This will allow users to set a large or small deadline and have it override the default 5m stream idle timeout.

examples/grpc_dynamic_routing/envoy_config.json
```diff
@@ -1044,7 +1044,7 @@
                             "route": {
                               "cluster": "backend-cluster-grpc-echo-oxouww7xzq-uc.a.run.app:443",
                               "hostRewriteLiteral": "grpc-echo-oxouww7xzq-uc.a.run.app",
-                              "idleTimeout": "401s",
```
nareddyt: Note this change is only to streaming methods. 

All unary methods still have `400` as response timeout, `401` as stream idle timeout.

tests/integration_test/http1_deadline_test.go
```diff
@@ -180,3 +186,96 @@ func TestDeadlinesForLocalBackend(t *testing.T) {
...
+		desc           string
+		confArgs       []string
+		reqDuration    time.Duration
+		deadlineToTest ConfiguredDeadline
+		wantErr        string
+	}{
+		// Please be cautious about adding too many time-based tests here.
+		// This can slow down our CI system if we sleep for too long.
+		{
+			desc: "When deadline is NOT specified, default deadline (15s) kicks in and the request fails.",
```
qiwzhang, line 204: Please add comment like following for each case
// used route Timeout = 15s (default), IdleTimeout = 25s
// This 504 is caused by `Timeout`, not by `IdleTimeout`.

tests/integration_test/grpc_deadline_test.go
qiwzhang: I see.  Your approach is to use "deadline" from service config for two different timeouts.  For streaming, it is stream_idle_timeout, for unary, it is response_timeout.   By this way, we don't need extra config for streaming_idle_timeout. so we don't expose --stream_idle_timeout flag. It is just for testing.

Hmm, using same "deadline" for two timeout?  My concern is confusing. Maybe we can document it in openapi spec. we should be fine. 

The flag is confusing,   if we can change it to --stream_idle_timeout_for_test,  and add comment, not to expose it, I will fell better.   Its current implementation is confusing to users.

tests/integration_test/http1_deadline_test.go
```diff
@@ -180,3 +186,96 @@ func TestDeadlinesForLocalBackend(t *testing.T) {
...
+			desc: "When a small deadline is specified, it overrides the larger global stream idle timeout specified by flag. The request fails with a 504, not 408.",
+			confArgs: append([]string{
+				"--stream_idle_timeout=15s",
+			}, utils.CommonArgs()...),
+			reqDuration:    time.Second * 8,
+			deadlineToTest: Short,
+			wantErr:        `504 Gateway Timeout, {"code":504,"message":"upstream request timeout"}`,
+		},
+		{
+			desc: "When a small deadline is specified, it overrides the smaller global stream idle timeout specified by flag. The request fails with a 504, not 408.",
```
qiwzhang, line 238: Can we have a test getting 408?

tests/integration_test/http1_deadline_test.go
nareddyt: Done. Looking at them after the weekend, I see why they are confusing. Tried renaming the test desc as well.

tests/integration_test/http1_deadline_test.go
nareddyt: It's not possible for that to occur, the behavior clients see is:
- For unary methods: Always return 504 when timeout occurs. We always set stream idle timeout to be more than the response deadline, so 408 will never occur.
- For streaming methods: Always return 408 when timeout occurs. Response deadlines do not apply to streaming methods, so 504 can never occur.

We have tests for these cases split across the 2 integration test files.

tests/integration_test/grpc_deadline_test.go
nareddyt: Done. Renamed the flag. Use prefix `_test_only` for consistency.

We can document it in the `x-google-backend` docs, but I feel this new behavior is intuitive and is what uses expected from the beginning.
- Before this PR, if users set `backend.rule.deadline` for a streaming method, it would basically be ignored and the stream idle timeout would default to 5m.
- With this PR, `deadline` will work as expected for both unary and streaming methods. The default stream idle timeout of 5m will be overridden correctly in both cases.

qiwzhang (approved):

TAOXUY (approved):

PR #459, opened 2021-01-11
b/177252401: Add integration test for bug in transcoding

nareddyt:
### Overview

Original analysis is here: go/espv2-backend-retry-analysis

Add an integration test to confirm and document the bug. Fix should be in upstream envoy's gRPC transcoder filter (and/or the transcoding library).

### Context

If a HTTP client sends unexpected query params, the transcoder does not handle them correctly. It returns the following error:

```
{"code":503,"message":"upstream connect error or disconnect/reset before headers. reset reason: remote reset"}
```

This is not good, it results in ESPv2 retrying the request and wastes time on the critical path. There is no easy way to configure Envoy to ignore this error specifically. This error is considered a **reset**, which we should retry on in general. 

Ref: [Documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on) on different retry triggers in Envoy.

Instead, the transcoder should return a HTTP 400 Bad Request for the invalid query param. This will prevent the unnecessary retry and be the correct behavior clients expect.

### DIFF
--- a/tests/integration_test/transcoding_errors_test.go
+++ b/tests/integration_test/transcoding_errors_test.go
@@ -30,7 +30,6 @@ type TranscodingTestType struct {
 	clientProtocol     string
 	httpMethod         string
 	method             string
-	noBackend          bool
 	token              string
 	headers            map[string][]string
 	bodyBytes          []byte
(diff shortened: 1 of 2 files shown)


tests/integration_test/transcoding_errors_test.go
```diff
@@ -60,8 +59,7 @@ func TestTranscodingServiceUnavailableError(t *testing.T) {
 		clientProtocol: "http",
 		httpMethod:     "GET",
 		method:         "/v1/shelves/200/books/2001?key=api-key",
-		noBackend:      true,
-		wantErr:        "503 Service Unavailable",
+		wantErr:        `503 Service Unavailable, {"code":503,"message":"upstream connect error or disconnect/reset before headers. reset reason: connection failure"}`,
```
nareddyt, line 62: This is expected, the backend is unavailable. But the error in the next test is not.

qiwzhang (approved):

TAOXUY (approved):
Curious how is it caught? In our e2e test?

nareddyt:
> Curious how is it caught? In our e2e test?

Yup, from e2e test. go/espv2-backend-retry-analysis has the full procedure if you are curious.

PR #463, opened 2021-01-21
b/176538814: Propagate trace ID to correlate access logs and traces

### LINKED ISSUE #431: Trace not added to a log entry
Hello, 
I am using ESPv2 2.21 within a GKE.
A request is coming in a sequence as following:

Http call >> glbc with neg -> service A  ->  service B

1. Service A receives a x-cloud-trace-context header within ESPv2 which is translated to a traceparent header and passed to a backend. A backend makes a call to a service B propagating a traceparent header.
2. Service B receives a traceparent header within ESPv2 which is passed to a backend

In a trace view I can see that a trace is propagated correctly and requests are tied together
image

However, when I try to search the stackdriver logs tied together by traceid, I cannot find espv2 requests, because they are missing trace field.

Here are the requests that are **correctly tied together** by trace id.

image

Here are the ESPv2 requests that are missing the trace field and are **not tied correctly** to a previous search result

image

Am I missing something, or is it just not a supported feature yet ?

Thanks!

nareddyt:
**Description**: Plumbing to take the trace ID provided by Envoy's trace interface and fill the data into SC report. See #431 for more background on the feature request.

**Limitations**: Will only work if ESPv2 is running on GCP. If running with `--non_gcp` (even with `--tracing_project_id` flag), this will be a NOOP and the trace won't be associated with the access log.

**Testing done**:
- Integration tests verify trace ID is propagated correctly
- e2e test to check Cloud Trace can correlate a trace to a request in Cloud Logging

**Examples**

You can now correlate access logs for a single request.

image

Fixes #431

### DIFF
--- a/src/api_proxy/service_control/request_builder.cc
+++ b/src/api_proxy/service_control/request_builder.cc
@@ -814,6 +814,11 @@ void SetOperationCommonFields(const OperationInfo& info,
   *op->mutable_end_time() = current_time;
 }
 
+std::string TraceResourceName(absl::string_view trace_id,
+                              absl::string_view project_id) {
+  return absl::StrCat("projects/", project_id, "/traces/", trace_id);
+}
+
 void FillLogEntry(const ReportRequestInfo& info, const std::string& name,
                   const std::string& config_id, const Timestamp& current_time,
                   LogEntry* log_entry) {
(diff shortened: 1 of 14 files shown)


qiwzhang (approved):

TAOXUY:
Nice job!

TAOXUY (approved):

PR #468, opened 2021-01-27
b/178530647 #1: Respond with HTTP 400 when required headers are omitted in CORS preflight request

### LINKED ISSUE #465: 503 returned for OPTIONS when ESPv2 handles CORS
Hello!

There is a weird behavior with the OPTIONS Request when ESPv2 is handling cors (not forwarded to the backend).

If the request has valid `origin` and `Access-Control-Request-Method` headers then a 200 is returned and the logs look ok.

However if it does not:
- the route returns a 503 with:
```json
{
  "message": "upstream connect error or disconnect/reset before headers. reset reason: connection failure",
  "code": 503
}
```
- the debug logs show that the gateway tried to connect to `127.0.0.1:8082` which failed (hence the message)

I am not a super pro of CORS. Is that expected behavior?

If I enable `allowCors: True`, it works as expected.

Thanks

nareddyt:
**Background:** The Envoy CORS filter is coded to `Continue` the request when a required header is missing or mismatched. See https://github.com/envoyproxy/envoy/issues/14233 for more details. This results in requests with missing `origin` or `access-control-request-method` headers not being handled properly.

**Fix:** Modify the CORS route config to generate two routes:
- First route will match OPTIONS request with _all_ required headers. This will be treated as a CORS request.
- Second round is a fallback. It will match OPTIONS requests with _any_ missing headers. It will direct reply with HTTP 400.

**Testing done:**
- Current CORS integration tests are very confusing. I combined/rewrote the first few to make it clear if they are testing simple vs. preflight request, which CORS config is used, etc. Feel free to ignore these rewrites, the diff is messy unfortunately.
- The bug fix in this PR is tested via `TestProxyHandlesCorsPreflightRequestsBasic`. You can find the `400` and `405` response codes in this test.

**Follow-up:** This PR only handles missing headers. When the `origin` header is specified but mismatched, the CORS filter `continues` the request as described above. It is not easy to handle this unless we do some "hack" in the SC filter. I may look into this later, not a big deal.

Fixes #465
Signed-off-by: Teju Nareddy nareddyt@google.com

### DIFF
--- a/src/go/configgenerator/route_generator.go
+++ b/src/go/configgenerator/route_generator.go
@@ -64,16 +64,18 @@ func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfigu
 	}
 	host.Routes = backendRoutes
 
-	cors, corsRoute, err := makeRouteCors(serviceInfo)
+	cors, corsRoutes, err := makeRouteCors(serviceInfo)
 	if err != nil {
 		return nil, err
 	}
 
 	if cors != nil {
 		host.Cors = cors
-		host.Routes = append(host.Routes, corsRoute)
-		jsonStr, _ := util.ProtoToJson(corsRoute)
-		glog.Infof("adding cors route configuration: %v", jsonStr)
+		host.Routes = append(host.Routes, corsRoutes...)
+		for i, corsRoute := range corsRoutes {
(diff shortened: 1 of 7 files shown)


qiwzhang (approved):
Thanks for cleaning up the CORS integration tests.  They are much cleaner now.

TAOXUY (approved):

PR #469, opened 2021-01-27
b/179288988: #1 refactor filter/perRouteConfig generator

TAOXUY:
- wrap making filter logic into a list of lambda method and move into filter_generator.go
- wrap making PerRouteConfigGen logic into a lambda method and create it along with making filters, which is actually tightly coupled.

**Goals:**
- allow remove/inject filters by processing the FilterGenerators.
- decouple making PerRouteConfig logic with making route

**Flow of Replacing the filter config**:
- call MakeFilterGenerator and get the filter chain(slice)
- replace SC v1 with SC v2 in the filter chain(slice)
- call MakeListeners(serviceInfo, filterGenerators)

**Follow-up**
PerRouteConfig code was unit tested under `route_generator_test` and is removed in this test. Considering this PR is so large, will add them in the next one:
- cover each `filter_gen`.
- cover route_generator will add PerRouteConfig

Signed-off-by: Xuyang Tao <taoxuy@google.com>


src/go/configgenerator/filter_generator.go
```diff
@@ -0,0 +1,758 @@
...
+	emptypb "github.com/golang/protobuf/ptypes/empty"
+	wrapperspb "github.com/golang/protobuf/ptypes/wrappers"
+	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
+	smpb "google.golang.org/genproto/googleapis/api/servicemanagement/v1"
+)
+
+// A wrapper of filter generation logic.
+type FilterGenerator struct {
+	FilterName string
+	FilterGenFun
```
nareddyt, line 52: Can we also associate the per-route config generation here? Each filter can have a function to generate its own per-route config. `PerRouteConfigGenFunc`

TAOXUY: SG

TAOXUY: Just think about it again. The `PerRouteConfigGenFunc` is tightly coupled in `FilterGenFun` while making filter config. Or we can make `PerRouteConfigGenFunc` under `FilterGenFun` after making it as object while I don't see too much improvement. WDYT?

TAOXUY: Also considering not all the filters are using per route config.

nareddyt: For each filter (JWT, SC v2, CORS, etc.), we want all the filter-generation code to be under one object or folder.

You can modify `FilterGenFunc` to return a list of per-route configs as well, that's fine. If the list is empty, then the filter doesn't need per-route config.

TAOXUY: PTAL

TAOXUY:
Please only take a review on the functional code(the unit test change will be large and will be posted later).

src/go/configgenerator/filter_generator.go
```diff
@@ -0,0 +1,777 @@
...
+	"strings"
+
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/options"
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util/httppattern"
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/ptypes"
+	anypb "github.com/golang/protobuf/ptypes/any"
+
+	sc "github.com/GoogleCloudPlatform/esp-v2/src/go/configinfo"
```
nareddyt, line 29: File is too long to review. Recommend splitting these into separate files. For each filter, have a file that has all the filter-specific logic in that.

```
sc_filter_gen.go
jwt_filter_gen.go
```

For example, JWT and SC filters have some helper functions that are only used for that filter. Some filters have helper functions for per-route configs.

src/go/configgenerator/filter_generator.go
```diff
@@ -0,0 +1,777 @@
...
+
+// A wrapper of filter generation logic.
+type FilterGenerator struct {
+	// The filter name.
+	FilterName string
+	// The function to generate filter config and the methods requiring per route configs.
+	FilterGenFunc
+	// The function to generate per route config.
+	// It should be set if the filter needs to set per route config.
+	sc.PerRouteConfigGenFunc
```
nareddyt, line 57: Please rename `sc`. It sounds like service control `scpb`.

src/go/configgenerator/listener_generator.go
```diff
@@ -54,113 +41,54 @@ const (
 
 // MakeListeners provides dynamic listeners for Envoy
 func MakeListeners(serviceInfo *sc.ServiceInfo) ([]*listenerpb.Listener, error) {
-	listener, err := makeListener(serviceInfo)
+	filterGenerators, err := MakeFilterGenerators(serviceInfo)
+	if err != nil {
+		return nil, err
+	}
+
+	listener, err := MakeListener(serviceInfo, filterGenerators)
```
nareddyt, line 49: How this will work in google3? I think our internal code will follow this flow:

1. Call `MakeFilterGenerator`
2. Override some of the key/value pairs in the slice. Remove SC v1, add our own filters.
3. Call `MakeListeners(serviceInfo, filterGenerators)`.

Is this correct?

src/go/configgenerator/filter_generator.go
```diff
@@ -0,0 +1,777 @@
...
+
+// The function type to generate filter config.
+// Return
+//  - the filter config
+//  - the methods needed to add per route config
+//  - the error
+type FilterGenFunc func(sc *sc.ServiceInfo) (*hcmpb.HttpFilter, []*sc.MethodInfo, error)
+
+// MakeFilterGenerators provide of a slice of FilterGenerator in sequence.
+func MakeFilterGenerators(serviceInfo *sc.ServiceInfo) ([]*FilterGenerator, error) {
```
nareddyt, line 68: Instead of a slice, can it return a map of filter name to `FilterGenerator`? Then you don't need to store the filter name in the struct, and lookups will be O(1).

src/go/configgenerator/filter_generator.go
TAOXUY: Good point!

Done.  We have separate files for 
- service control
- jwt authn
- backend auth
- path rewrite

src/go/configgenerator/listener_generator.go
TAOXUY: My thought is the filter_generator is responsible to generate the intended filter chains.

src/go/configgenerator/filter_generator.go
TAOXUY: The lookup overhead is over-concerned since it is less than 10 filters.

TAOXUY:
@qiwzhang Please take a look.

src/go/configgenerator/filter_generator.go
```diff
@@ -0,0 +1,262 @@
...
+		// TODO(b/176432170): Handle errors here, prevent startup.
+		filterGenerators = append(filterGenerators, &FilterGenerator{
+			FilterName:            util.JwtAuthn,
+			FilterGenFunc:         jaFilterGenFunc,
+			PerRouteConfigGenFunc: jaPerRouteFilterConfigGen,
+		})
+	}
+
+	// Add Service Control filter if needed.
+	if !serviceInfo.Options.SkipServiceControlFilter {
```
qiwzhang, line 98: I am wondering if we can move such "SkipServiceControlFilter" check into each filterGen function.  It is filter specific.

If we move this into each filter, then each filter can register a factory (just like envoy factory register). At the high level,  we just loop over registered gen function to create create listener.

By doing this way,  OSS and g3 just need to have its own filter_gen folders.  just like envoy extensions.

qiwzhang (changes requested):

src/go/configgenerator/filter_generator.go
```diff
@@ -0,0 +1,262 @@
...
+	filterGenerators = append(filterGenerators, &FilterGenerator{
+		FilterName: util.Router,
+		FilterGenFunc: func(sc *ci.ServiceInfo) (*hcmpb.HttpFilter, []*ci.MethodInfo, error) {
+			return makeRouterFilter(serviceInfo.Options), nil, nil
+		},
+	})
+	return filterGenerators, nil
+}
+
+func makeTranscoderFilter(serviceInfo *ci.ServiceInfo) *hcmpb.HttpFilter {
```
nareddyt, line 166: can we move these into different files as well, one per filter?

My idea is similar to Wayne's. In g3, we will have an override of `MakeFilterGenerators`. So the `filter_generator.go` file will have different contents in g3 vs OSS. All the other per-filter files can be synced as needed into g3.

OSS:
```
filter_generator.go
sc_v1.go
backend_auth.go
transcoding.go
...
```

g3
```
filter_generator.go (modified contents for g3 filters)
sc_v2.go
transcoding.go
...
```

So we do not want to tie any filter-specific functionality into filter_generator.go.

src/go/configgenerator/filter_generator.go
nareddyt: I think putting this logic here is OK. Basically `MakeFilterGenerators` will decide which filters to create. In OSS, it will decide based on the config generator flags/options. In g3, we will have an override of `MakeFilterGenerators` that will create filters differently.

It would be nice to have `Skip_${FilterName}_Filter` flags per filter, but honestly I don't think we will use them. It seems like an antipattern to have these flags for testing. The service config should decide which filters are generated, not random flags we pass in.

src/go/configgenerator/filter_generator.go
nareddyt: Let me know if you have a different idea for how filters will be generated in google3

qiwzhang: Yes, I like the idea of each filter has its own .go file,  or even its own go package.   There are different between Teju idea and mime.

Mine:  we don't have separate MakeFilterGenerator,  one for OSS and one for g3.  There is only one, a very simple, it just read the registered filter_list.  Similar to Envoy extensions,  there is not MakeExtersions function,   each extension just register itself.  By removing, adding filter_xx.go code, OSS and g3 will have different list.

with this list,  each filter decides if it needs to be added.  that is why moving the "if" checking to the filter config gen code is important. The high level code just blindly loop through each registered filter, to call its config_gen function.  IF return nil,  not to add it.

Teju has MakeFilterGenerator function, two separate one, one for OSS, and one for g3.

src/go/configgenerator/filter_generator.go
qiwzhang: I am not suggesting each filter should have a skip_xyz_filter flag.  I am suggesting "if" checking can be moved into each filter config_gen function.

src/go/configgenerator/filter_generator.go
nareddyt: I see, each filter_xx.go can statically register itself. Makes sense to me.

TAOXUY: My understanding about Wayne's proposal: 
- each filter register itself
- its config generator handles the logic if the filter is needed or not like checking `Skip_${FilterName}_Filter` flags
- OSS and g3 has two separate lists of targeted filter lists

My 2 cents: it would be overkill for our current requirement, swapping SC_V2 filter.

It is  will be confusing have two places to control the filter chain, one in the filter lists provided and another inside the filter config generator. I would prefer having a central sequential `MakeFilterGenerator` to control this,  and we don't necessarily need two `MakeFilterGenerator` but just patch the result it generates to replace SC_v2.

If we plan to have more filters in g3, we can easily refactor based on that.

qiwzhang: My idea is we don't have a high level function MakeFilterGenerator.   We just have a global variable registered_filter list.
By register or not register, we can control the list.  For example:
* we dont move in v1_service_control code in g3, so it will not be registered
* we don't move in backed_auth, path_rewrite into g3,  so they will not be in g3 list.
* v2_sc only in g3, not in OSS.  so their registered list will be different.

Technically, I am not sure if golang can do this easily.  If each filter is a package,  each package can have __init__funciton, that can register.  If each filter is a go file,  not sure if it can have its own __init__ function for each go file.

If it is too much trouble,  Having separate MakeFilterGenerator function is fine for me too

TAOXUY: Oh, I see. Good point of decoupling the config generation code from the whole library in terms of it is introduced in g3 or not! Though the readability of wrapping all the logic of if the filter config is generated into each package doesn't looks good, compared to a literal sequential `MakeFilterGenerator`, I am OK with that. 

Feasibility: yes, I know Golang should be able to do this.

qiwzhang: Not clear which approach you are going to take.   Are you going to add filter self-registration code?
If no,  I will review this code more carefully, and get this one in.

TAOXUY: Sorry for the misleading. I prefer not to add the registry for now.

examples/auth/envoy_config.json
```diff
@@ -338,10 +338,6 @@
                               "timeout": "15s"
                             },
                             "typedPerFilterConfig": {
-                              "com.google.espv2.filters.http.service_control": {
```
qiwzhang, line 341: Why is service_control perRouteConfig removed?

TAOXUY: Because we don't set service control filter in this test and before the refactor, the route perRouteConfig isn't aware of that the service control filter is missing.

https://github.com/GoogleCloudPlatform/esp-v2/blob/1ffe51f3e7526214e050f6eff9084ed3656e1d05/src/go/bootstrap/static/bootstrap_test.go#L61

src/go/configgenerator/filter_gen_path_rewrite.go
```diff
@@ -0,0 +1,63 @@
...
+	ci "github.com/GoogleCloudPlatform/esp-v2/src/go/configinfo"
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util"
+	"github.com/GoogleCloudPlatform/esp-v2/src/go/util/httppattern"
+	hcmpb "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	"github.com/golang/protobuf/ptypes"
+	anypb "github.com/golang/protobuf/ptypes/any"
+)
+
+var prPerRouteFilterConfigGen = func(method *ci.MethodInfo, httpRule *httppattern.Pattern) (*anypb.Any, error) {
+	pr := MakePathRewriteConfig(method, httpRule)
```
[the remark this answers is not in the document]
qiwzhang: Can we move MakePathRewriteConfig function into this file?

qiwzhang (approved):

src/go/configgenerator/filter_gen_path_rewrite.go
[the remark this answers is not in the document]
TAOXUY: Done

TAOXUY:
Also finished the unit test change. PTAL.

qiwzhang (approved):

src/go/configgenerator/filter_generator.go
nareddyt: > I would prefer having a central sequential MakeFilterGenerator to control this, and we don't necessarily need two MakeFilterGenerator but just patch the result it generates to replace SC_v2.

I don't understand how this will work in g3. We need to support:
- One version of `MakeFilterGenerator` that will generate SC v1, backend auth, path rewrite, etc. for API Gateway.
- Another version of `MakeFilterGenerator` that will generate SC v2 for us.

I thought that the caller of config generator would decide the logic. But you're moving the decision to config generator itself. There is only 1 copy of config generator inside g3. So how can we have 2 different filter generation decisions?

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -0,0 +1,262 @@
...
+	filterGenerators = append(filterGenerators, &FilterGenerator{
+		FilterName: util.Router,
+		FilterGenFunc: func(sc *ci.ServiceInfo) (*hcmpb.HttpFilter, []*ci.MethodInfo, error) {
+			return makeRouterFilter(serviceInfo.Options), nil, nil
+		},
+	})
+	return filterGenerators, nil
+}
+
+func makeTranscoderFilter(serviceInfo *ci.ServiceInfo) *hcmpb.HttpFilter {
```
nareddyt, line 166: Move these helper functions into different files.

If we have a different version of MakeFilterGenerator function in g3, we don't want to rely on copybara regex to replace the entire function and add imports.

Instead we should just have a forked copy of this file in g3. It will be easy to do this if these helper functions are in different files, then we only need to worry about diffs in `MakeFilterGenerator`.

src/go/configgenerator/filter_generator.go
TAOXUY: In the import target [doc](https://docs.google.com/document/d/1TDyjFNfkx-s2NaopWEK_-xYi1l1DVEIYYSZBlKYgPng/edit#bookmark=id.dnilp44edtew), I thought we got aligned about getting the filter config chains from ESPv2 lib and replacing the SC v2 filter, because we are essentially applying the GKE sidecar deployment model for CAP. 

Please correct me if I miss sth?

src/go/configgenerator/filterconfig/filter_generator_test.go
```diff
@@ -0,0 +1,954 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filterconfig
+
+import (
```
nareddyt, line 17: What is the plan for this test file in g3? Will any of these fail in g3?

 Just like the filter generator files, it will be easier if we can move each filter-specific test to its own file. Then we copy just the files we need.

src/go/configgenerator/filterconfig/filter_generator.go
TAOXUY: ACK

src/go/configgenerator/filterconfig/filter_generator_test.go
TAOXUY: ACK

Was planning to add PerRouteConfig tests for individual filter_gen along with sort out tests in this test into specific one. Will do this PR.

src/go/configgenerator/filter_generator.go
nareddyt: Discussed offline, I understand the flow now. Thanks!

nareddyt (approved):

src/go/configgenerator/filterconfig/filter_generator_test.go
TAOXUY: Move the tests with PerRouteConfig into their own test files for better readability.

Issue #479, opened 2021-02-09
Feature: Add extra headers sent to the backend


dariuszsq:
Team,
I have multiple ESPv2 with one backend. I realized that ESP is not able to send additional / extra header to the backend which i can use to recognize source from where request is comming.

Kindly, can i ask you to add this feature to next release ?

Thank you
Dariusz Sobiechowski

qiwzhang:
In theory, this feature can be implemented as following:
1) add, append or remove any request headers before sending to the upstream
2) add, append or remove any response headers before sending to the downstream. 
3) it can be per-route. this require OP service config change.

Let us only do 1) for  now.   add some flags to support this.  e.g.

```
--request_headers_add=key=value
--request_headers_append=key=value
--request_headers_remove=key
```

This flag can be set multiple times. e.g
```
--request_headers_add=key1=value1  --request_headers_add=key2=value2
```

dariuszsq:
Sounds like a plan - do you know ETA ?

nareddyt:
Hi @qiwzhang, do you have an implementation plan for this? I'm curious if you will use the envoy router or if you'll introduce custom logic into our SC filter.

I know the router can add headers, but will it remove the header if it's already in the request?

qiwzhang:
I will use VirtualHost or RouteConfig  [requst_headers_to_add](https://github.com/envoyproxy/envoy/blob/eeb7adc3a30456f0d4ac65e5e6c8e88e25481d2a/api/envoy/config/route/v3/route.proto#L59) field.

So it can support custom request/response [headers](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers) 

With this, it support both request and response.  so I am going to add 4 flags
--add_request_header
--append_request_header
--add_response_header
--append_response_header

PR #480, opened 2021-02-13
Add flags to add extra headers

qiwzhang:

Add following flags to add extra headers to the request:
```
--add_request_header
--append_request_header
--add_response_header
--append_response_header
```

https://github.com/GoogleCloudPlatform/esp-v2/issues/479


TAOXUY (approved):

src/go/configgenerator/route_generator.go
```diff
@@ -83,12 +84,79 @@ func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfigu
...
+		key_value := strings.Split(h, "=")
+		if len(key_value) != 2 {
+			return l, fmt.Errorf("invalid header: %v. should be in key=value format.", h)
+		}
+		if key_value[0] == "" {
+			return l, fmt.Errorf("header key should not be empty for: %v.", h)
+		}
+		l = append(l, &corepb.HeaderValueOption{
+			Header: &corepb.HeaderValue{
+				Key:   key_value[0],
```
nareddyt, line 119: Should we have any validation for duplicate keys? What will Envoy do in these two cases?

```
addRequestHeaders: k=a;k=b
```
```
addRequestHeaders: k=a
appendRequestHeaders: k=b
```

If envoy will fail startup, no need for validation on our side. If envoy will randomly honor only one of the rules, I would prefer we validate against it.

tests/integration_test/add_headers_test.go
```diff
@@ -0,0 +1,132 @@
...
+				"key2": "old-value2",
+			},
+			wantRespHeader: map[string]string{
+				"Key1":      "value1",
+				"Echo-Key2": "old-value2;new-value2",
+			},
+		},
+	}
+	for _, tc := range testData {
+		func() {
```
nareddyt, line 107: Nit:
```
t.Run(tc.desc, func() {
  ...
})
```
Makes it easier to debug test failures.

nareddyt (approved):

src/go/configgenerator/route_generator.go
qiwzhang: Check Envoy code, they don't verify.

tests/integration_test/add_headers_test.go
qiwzhang: Done

PR #483, opened 2021-02-23
Move integration tests into separate packages

nareddyt:
Agreed this change is not ideal, but it's needed to run the integration tests in google3. Otherwise the internal version of https://github.com/bazelbuild/bazel-gazelle will generate a single test target for all integration tests. Much easier to test and import if each file is a separate test target.

Files moved using [this script](https://unix.stackexchange.com/a/480808/331278). Manually changed package names.

Confirmed this works in google3.


nareddyt:
This reverts https://github.com/GoogleCloudPlatform/esp-v2/pull/59.
Need to debug test failures. Seems like tests are still running in parallel on CI

qiwzhang (approved):

TAOXUY (approved):

PR #487, opened 2021-03-04
b/181697283: Propagate trace context headers to SC Check

nareddyt:
**Background**: This was an action item from the SRE brainstorming meeting to improve observability of latency issues. We should propagate this header so we can better debug customer issues by correlating the external request with internal ones.

**Implementation**: Very simple, uses the Envoy tracing interface.

**Testing done**: Integration test.

### DIFF
--- a/tests/integration_test/opencensus_tracing_test/opencensus_tracing_test.go
+++ b/tests/integration_test/opencensus_tracing_test/opencensus_tracing_test.go
@@ -622,6 +622,16 @@ func TestTraceContextPropagationHeadersForScCheck(t *testing.T) {
 			createTraceparentContext(traceId, spanId),
 		},
 	}
+	expectedTraceContexts := map[string][]string{
+		// Only the trace id is checked. Span id should be changed.
+		// By default, both trace contexts are generated.
+		"Traceparent": {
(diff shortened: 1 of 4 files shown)


tests/integration_test/opencensus_tracing_test/opencensus_tracing_test.go
```diff
@@ -612,6 +612,92 @@ func TestTraceContextPropagationHeaders(t *testing.T) {
...
+		expectedScReqHeaders map[string][]string
+	}{
+		{
+			desc:              "SC Check receives trace context propagation header.",
+			tracingSampleRate: 1,
+			expectedScReqHeaders: map[string][]string{
+				// Only the trace id is checked. Span id should be changed.
+				"Traceparent": {
+					createTraceparentContextPrefix(traceId),
+				},
```
qiwzhang, line 638: what is the different between this case and case 3?

should we also expect "X-Cloud-Trace-Context" in this case?

nareddyt: Fixed. Merged the 2 cases together.

TAOXUY (approved):

qiwzhang (approved):

PR #497, opened 2021-03-29
b/182396604: Remove additional_bindings in test

nareddyt:
These are not needed for grpc-web support. See discussion in bug. @qiwzhang will also update the docs.


qiwzhang:
I don't remember what exactly the reason why we need to add this additional_bindings for grpc_web now.  Does the proto descriptor file used in the integration_test have such additional_bindings?   If yes, can you update the proto descriptor files after the removal to make sure it will not break the grpc_web?

nareddyt:
Yup, this PR also updates `tests/endpoints/bookstore_grpc/proto/api_descriptor.pb` to remove them. It seems to be working so far.

qiwzhang (approved):
OK, cool.  Thanks for testing it out

PR #501, opened 2021-04-20
b/184142099: upgrade envoy to 18.2

TAOXUY:
Upgrade envoy to 18.2.  Changes:
- several timeout integration tests become flaky between 408(downstream timeout) and 504(upstream timeout) so use a general timeout error to compare for now
- due to https://github.com/envoyproxy/envoy/pull/15585, route.max_stream_timeout(streaming timeout) will pick up `route.timeout`. As we don't have setting on it before, set a static 1-hr streaming timeout for backward compatibility.

### DIFF
--- a/tests/integration_test/grpc_deadline_test/grpc_deadline_test.go
+++ b/tests/integration_test/grpc_deadline_test/grpc_deadline_test.go
@@ -77,22 +77,23 @@ plans {
   }
 }`,
 		},
-		{
-			desc: "Success after 20s because ESPv2 automatically disables response timeouts for streaming RPCs",
-			testPlan: `
-plans {
-  echo_stream {
-    call_config {
-      api_key: "this-is-an-api-key"
-    }
-    request {
-      text: "Hello, world!"
-      response_delay: 20
-    }
-    count: 1
-  }
-}`,
-		},
+		// TODO(b/185919750): Adjust the integration tests due to the introduced newly-involved max_stream_timeout.
+		//		{
+		//			desc: "Success after 20s because ESPv2 automatically disables response timeouts for streaming RPCs",
+		//			testPlan: `
+		//plans {
+		//  echo_stream {
+		//    call_config {
+		//      api_key: "this-is-an-api-key"
+		//    }
+		//    request {
+		//      text: "Hello, world!"
+		//      response_delay: 20
+		//    }
+		//    count: 1
+		//  }
+		//}`,
+		//		},
 	}
 
 	for _, tc := range testData {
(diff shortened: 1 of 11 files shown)


TAOXUY:
Made b/185946962 to track the prow gke issue; This PR should be good as covered as cloud run test case should cover it.

tests/integration_test/grpc_deadline_test/grpc_deadline_test.go
```diff
@@ -77,22 +77,23 @@ plans {
...
-    }
-    request {
-      text: "Hello, world!"
-      response_delay: 20
-    }
-    count: 1
-  }
-}`,
-		},
+		// TODO(b/185919750): Adjust the integration tests due to the introduced newly-involved max_stream_timeout.
```
nareddyt, line 80: Well if these tests caught the error, won't this affect our customers as well? It changes the behavior the customers expect.

What is the failure message for these tests?

ericvn: Our Istio tests noted the change from 504 to 408 and our tests are now failing. See comment https://github.com/istio/istio/issues/31616#issuecomment-824156192 as to if this was expected and documented.

TAOXUY: I think after setting a static large `max_stream_timeout` can workaround. The issue now is the error code is flaky between 408 and 504.

nareddyt: I guess the actual error code does not matter, as we encourage users to not depend on the exact error.

https://cloud.google.com/apis/design/errors

> Error messages are not part of the API surface. They are subject to changes without notice. Application code must not have a hard dependency on error messages.

@TAOXUY can you change the failing integration tests to just check an error occurred, without checking the specific error? Thanks for researching and setting the static `max_stream_timeout`.

TAOXUY:
Our CI is broken as OSS Prow's issue. They are working on it. b/186123000

nareddyt (approved):

tests/integration_test/grpc_deadline_test/grpc_deadline_test.go
TAOXUY: ACK on not depending on the error.

TAOXUY:
Pass the CI. Ready to review.

nareddyt (approved):

TAOXUY:
This [max_stream_route](https://github.com/envoyproxy/envoy/pull/16133) change is reverted and is going to be backported in 18.2.

PR #505, opened 2021-04-27
b/184731851: Fix overhead latency calculation for backend timeout

nareddyt:
When a backend timeout occurs, currently we set backend request time as `0`. Now we add some special handling to handle backend timeout specifically.

Testing:
- Automated via unit tests
- Manual testing by inspecting Report payload in integration tests. Hard to automate with current test framework.


src/envoy/http/service_control/handler_utils.cc
```diff
@@ -177,25 +177,35 @@ void fillLatency(const Envoy::StreamInfo::StreamInfo& stream_info,
   if (stream_info.requestComplete()) {
     latency.request_time_ms =
         convertNsToMs(stream_info.requestComplete().value());
-    filter_stats.filter_.request_time_.recordValue(latency.request_time_ms);
   }
 
   auto start = stream_info.firstUpstreamTxByteSent();
```
nareddyt, line 182: The logic here is messy, I'll try to explain:
- If we know how long the backend took from request/response: overhead = total - backend
- Otherwise, we calculate overhead (less accurately) and calculate: backend = total - overhead

qiwzhang (approved):

src/envoy/http/service_control/handler_utils.cc
```diff
@@ -177,25 +177,35 @@ void fillLatency(const Envoy::StreamInfo::StreamInfo& stream_info,
...
+  if (latency.request_time_ms >= 0) {
+    if (latency.backend_time_ms >= 0) {
+      // Calculation based on backend response timing.
+      latency.overhead_time_ms =
+          latency.request_time_ms - latency.backend_time_ms;
+    } else if (start) {
+      // Less accurate calculation based on overhead timing.
+      latency.overhead_time_ms = convertNsToMs(start.value());
+      latency.backend_time_ms =
+          latency.request_time_ms - latency.overhead_time_ms;
```
TAOXUY, line 202: Sorry could you comment which is the codepath of timeout? Is it same as the rejected request in line 187?

nareddyt: When timeout occurs, the following should happen:
1. `start` has value, but `end` has no value. `if/else` block on line 184 is skipped
2. Default value of `latency.backend_time_ms` is -1, so it goes into `if` on line 198

nareddyt:
Do not submit yet. Seems there is a bug in envoy where the `firstUpstreamTxByteSent` is not always set. Debugging it.
We have no other way to determine the latency during timeout if this field is missing.

nareddyt:
PTAL again. I added a hacky solution to handle this envoy bug: https://github.com/envoyproxy/envoy/issues/16684

It will under-report overhead latency when a retry occurs, but it should be OK.

qiwzhang (approved):

TAOXUY (approved):

Issue #503, opened 2021-04-27
Cloud Run ESPv2 timed out even though Cloud function returned with 200 response


mejuhi:
We have Architecture where we have configured Cloud Run and Cloud Functions to connect using Cloud Endpoint. We have used following [documentation](https://cloud.google.com/endpoints/docs/openapi/get-started-cloud-functions). In which the ESP image is running on Cloud run

Base image used by cloud run is gcr.io/endpoints-release/endpoints-runtime-serverless:2.24.0 which gets build using the following [script](https://github.com/GoogleCloudPlatform/esp-v2/blob/40caa16ea9b4231ab9272bce2c9dfcec95dee3e5/docker/serverless/gcloud_build_image)

Cloud run (ESPv2) is running with following configuration:

CPU allocated: 2
Memory allocated: 2048
Concurrency: 80
Request timeout: 800 seconds
Auto-scaling: Min instances: 5
Auto-scaling: Max instances: 1,000

Additionally we have added following arguments in the base docker image:
ENV ESPv2_ARGS ^++^--cors_preset=basic++--cors_allow_method="GET,PUT,POST"++--backend_retry_num=2

The Cloud Function is API Business Layer. Cloud function was able to execute successfully and returned response with 200 HTTP status code, but somehow Cloud Run ESP timed out and then Cloud endpoint timed out with HTTP status code 504. 

Attaching screenshot:
![image](https://user-images.githubusercontent.com/45816852/116202142-e4f22580-a757-11eb-8bde-f33a0327eb91.png)

qiwzhang:
Looking at the log from your screenshot.  It seems that 504 is from time-out.  The last two lines have  800s and 3600s respectively.   

A couple questions:
* Does 504 happens to all requests or only to some?
* If only for some, do these requests have long latency to response?

mejuhi:
1. It has happened only few times
2. No, as you can see the third last line in the screenshot, cloud function was able to complete its execution in 1043 ms with 200, but still cloud run running ESPv2 timed out for some reason.

qiwzhang:
backend default timeout is 15s.  It seems that your backend responded within a couple seconds so it should not be the default timeout issue.
How about the data size,  your screen shot blocked out the data size,   what are the request and response data size?

mejuhi:
Http request size i can see from the log of Cloud Endpoint is around 2846 B (attached screenshot below), and response size should be around 1274 B (from the cloud function)

[...]

PR #509, opened 2021-05-06
b/187411694: add perTryTimeout for doing retry when the upstream times out

TAOXUY:
When the upstream times out, ESPv2 won't do retry currently as the route timeout has been exhausted. Add `per_try_timeout_sec` flag to allow doing retry in this case. Though this flag is related to the route timeout(set in the service config), there is no need to have it per API method/route and it is also more consistent to be together with other retry setting in deployment flags. 

**Context**: it is to fix https://github.com/GoogleCloudPlatform/esp-v2/issues/503, the response code detail of which is `upstream_response_timeout`.

### DIFF
--- a/src/go/configmanager/flags/flags.go
+++ b/src/go/configmanager/flags/flags.go
@@ -160,8 +160,12 @@ var (
         x-envoy-retry-grpc-on(https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on).`)
 	BackendRetryNum = flag.Uint("backend_retry_num", 1,
 		`The allowed number of retries. Must be >= 0 and defaults to 1. This retry
-	setting will be applied to all the backends if you have multiple ones.`)
-	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0, "The upstream timeout per retry attempt in second. By default it is set with 0 and a request that times out won't be retried.")
+setting will be applied to all the backends if you have multiple ones.`)
+	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0,
+		`The backend timeout per retry attempt in second. A default value 0
+means it is unspecified and ESPv2 will use the "deadline" in the "x-google-backend"
+extension. Consequently, a request that times out will not be retried as the total
+timeout budget would have been exhausted.`)
 )
 
 func EnvoyConfigOptionsFromFlags() options.ConfigGeneratorOptions {
--- a/src/go/configmanager/flags/flags.go
+++ b/src/go/configmanager/flags/flags.go
@@ -160,7 +160,7 @@ var (
         x-envoy-retry-grpc-on(https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on).`)
 	BackendRetryNum = flag.Uint("backend_retry_num", 1,
 		`The allowed number of retries. Must be >= 0 and defaults to 1. This retry
-setting will be applied to all the backends if you have multiple ones.`)
+        setting will be applied to all the backends if you have multiple ones.`)
 	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0,
 		`The backend timeout per retry attempt in second. Please note the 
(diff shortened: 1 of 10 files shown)


src/go/configmanager/flags/flags.go
```diff
@@ -161,6 +161,7 @@ var (
 	BackendRetryNum = flag.Uint("backend_retry_num", 1,
 		`The allowed number of retries. Must be >= 0 and defaults to 1. This retry
 	setting will be applied to all the backends if you have multiple ones.`)
+	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0, "The upstream timeout per retry attempt in second. If unspecified, a request that times out won't be retried.")
```
qiwzhang, line 164: what does 0 mean?  should it be "None"?

TAOXUY: golang is strong typing so it won't accept nil for primitive uint type. 0 here means the default value.

Not sure if I get your point.

qiwzhang: maybe state that 0 means "nospecified"?

TAOXUY: Done.

src/go/configmanager/flags/flags.go
```diff
@@ -161,7 +161,7 @@ var (
 	BackendRetryNum = flag.Uint("backend_retry_num", 1,
 		`The allowed number of retries. Must be >= 0 and defaults to 1. This retry
 	setting will be applied to all the backends if you have multiple ones.`)
-	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0, "The upstream timeout per retry attempt in second. If unspecified, a request that times out won't be retried.")
+	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0, "The upstream timeout per retry attempt in second. By default it is set with 0 and a request that times out won't be retried.")
```
qiwzhang, line 164: How about,  a default value 0 means ....

BTW, I am still not clear the "deadline" from x-google-backend and this one. Could you help to explain it?

TAOXUY: Done PTAL.

qiwzhang (approved):

src/go/configmanager/flags/flags.go
```diff
@@ -160,7 +160,14 @@ var (
         x-envoy-retry-grpc-on(https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on).`)
 	BackendRetryNum = flag.Uint("backend_retry_num", 1,
 		`The allowed number of retries. Must be >= 0 and defaults to 1. This retry
-	setting will be applied to all the backends if you have multiple ones.`)
+        setting will be applied to all the backends if you have multiple ones.`)
+	BackendPerTryTimeoutSec = flag.Uint("backend_per_try_timeout_sec", 0,
```
nareddyt, line 164: Nit: change to flag.Duration and have python pass as string?

nareddyt (commented):
Why not just use the `deadline` from `x-google-backend` as per-try timeout if user configures retry policy? This is a lot more intuitive and is what the user expected when they specified deadline. Users did not expect `deadline` to be the full deadline across retries.

My main concern is adding this flag makes the design inflexible. And then it will be a breaking change when if we want to support per-route per-retry timeout in the future.

src/go/configmanager/flags/flags.go
TAOXUY: We have retry setting by default. To the contrary, users will be surprised when the real deadline=deadline * (1 + retry_num) if we use per_try_timeout=deadline without knowing retry setting. Potentially, it is a breaking change. If we set `per_try_timeout` here, it would be transparent to users. If users would like to set the retry in the flags, they can see the description about the relationship between x-google-backend.deadline and per_try_timeout. 

I regard it as part of retry settings so it is more natural to put them together in deployment flags.
 
As for the per-route retry, actually we have same discussion when we were adding the retry. We doubted per route retry would be really needed for our customers, so we added other retry setting as flag.

nareddyt (approved):
SG. Then my only comment is about the flag

TAOXUY:
> SG. Then my only comment is about the flag

done.

PR #510, opened 2021-05-07
b/187501317: Test gRPC streaming on Cloud Run

nareddyt:
Cloud Run supports bidirectional streaming since earlier this year. Let's test it in our pre-existing e2e tests.


nareddyt:
Fixed by #537

PR #511, opened 2021-05-12
Add path normalization options for CVE-2021-29492

nareddyt:
Changes:
- Update to Envoy 1.18.3.
- Enable configuration of `disallow_escaped_slashes_in_path`.
- Flip `normalize_path` and `merge_slashes_in_path` to true by default.

Testing done:
- Unit tests
- Integration tests


docker/generic/start_proxy.py
```diff
@@ -361,9 +361,9 @@ def make_argparser():
         help='''Allow headers contain underscores to pass through. By default
         ESPv2 rejects requests that have headers with underscores.''')
 
-    parser.add_argument('--normalize_path', action='store_true',
-        help='''Enable normalization of the `path` HTTP header according to
-        RFC 3986. It is recommended to turn on this option if your backend
+    parser.add_argument('--no_normalize_path', action='store_true',
```
qiwzhang, line 364: disable_path_normalization ?

docker/generic/start_proxy.py
```diff
@@ -377,11 +377,15 @@ def make_argparser():
...
+        By default, ESPv2 will normalize paths.
+        Disable the feature only if your traffic is affected by the behavior.
+        
+        For more details, see:
+        https://cloud.google.com/api-gateway/docs/path-templating''')
 
-    parser.add_argument('--merge_slashes_in_path', action='store_true',
-        help='''Enable merging of adjacent slashes in the `path` HTTP header.
-        It is recommended to turn on this option if your backend
+    parser.add_argument('--no_merge_slashes_in_path', action='store_true',
```
qiwzhang, line 386: disable....?

docker/generic/start_proxy.py
```diff
@@ -394,7 +398,34 @@ def make_argparser():
...
+        help='''Enables escaping of the following characters in the `path`
+        HTTP header:
+        - %%2F or %%2f -> /
+        - %%5C or %%5c -> \
+        
+        When enabled, the behavior depends on the protocol:
+        - For OpenAPI backends, request paths with escaped slashes will be
+          automatically unescaped via a redirect.
+        - For gRPC backends, request paths with escaped slashes will be
+          rejected.
```
qiwzhang, line 417: not clear to me. 

if true,   %2f will be still %2f when received by the backend, is that correct?  if so,  should this be called `keep_escaped_slash_in_path`

TAOXUY (approved):

docker/generic/start_proxy.py
nareddyt: No, if this flag is set, the %2f will be converted to a `/` character. I tried to clarify.

docker/generic/start_proxy.py
nareddyt: Done

docker/generic/start_proxy.py
nareddyt: Done

docker/generic/start_proxy.py
```diff
@@ -394,7 +398,34 @@ def make_argparser():
...
+        help='''Enables escaping of the following characters in the `path`
+        HTTP header before the request is sent to the backend:
+        - %%2F or %%2f is converted to /
+        - %%5C or %%5c is converted to \
+        
+        When enabled, the behavior depends on the protocol:
+        - For OpenAPI backends, request paths with unescaped percent-encoded
+          slashes will be automatically escaped via a redirect.
+        - For gRPC backends, request paths with unescaped percent-encoded 
+          slashes will be rejected.
```
shuoyang2016, line 417: qq: does it mean grpc backend will not be vulnerable?

nareddyt: gRPC doesn't support redirects, so we just reject the request. That will prevent against the vulnerability occurring if the request is rejected.

docker/generic/start_proxy.py
qiwzhang: I see.  then how about `--unescape_escaped_slashes_in_path` with help as
If there are any escaped slashes in path, e.g. %%2f,  they will be converted to slashes when doing routing.

shuoyang2016 (approved):

docker/generic/start_proxy.py
nareddyt: How about `--disallow_escaped_slashes_in_path`? Because gRPC does not support redirects, this flag will rejects gRPC requests that have escaped characters. Only non-gRPC traffic will be automatically unescaped.

qiwzhang: Cool.  I like it.  Thanks

qiwzhang (approved):

PR #514, opened 2021-05-18
GCS Runner: Allow unknown static fields

nareddyt:
This allows new fields to be consumed in accordance with standard SemVer
procedure. Otherwise, new features can become breaking changes.

### DIFF
--- a/src/go/gcsrunner/start_envoy.go
+++ b/src/go/gcsrunner/start_envoy.go
@@ -51,6 +51,7 @@ func StartEnvoyAndWait(signalChan chan os.Signal, opts StartEnvoyOptions) error 
 		"--log-path", opts.LogPath,
 		"--log-format", "%L%m%d %T.%e %t envoy] [%t][%n]%v",
 		"--log-format-escaped",
+		"--allow-unknown-static-fields",
 	}
 	if opts.ComponentLogLevel != "" {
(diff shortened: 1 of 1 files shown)


todd-beckman (approved):

toddbeckman-g (approved):

src/go/gcsrunner/start_envoy.go
```diff
@@ -51,6 +51,7 @@ func StartEnvoyAndWait(signalChan chan os.Signal, opts StartEnvoyOptions) error
 		"--log-path", opts.LogPath,
 		"--log-format", "%L%m%d %T.%e %t envoy] [%t][%n]%v",
 		"--log-format-escaped",
+		"--allow-unknown-static-fields",
```
qiwzhang, line 54: It is kind of dangerous to turn this on.  Default is to make sure envoy config can be understood by the envoy.  

I prefer not to enable this

qiwzhang (changes requested):

src/go/gcsrunner/start_envoy.go
qiwzhang: I guess, our config is always generated by config mgr,  it should not have typo.  
It is ok to turn this on then

qiwzhang (approved):

PR #519, opened 2021-05-21
b/188808823: Simplify CORS regex and fix brittle tests

nareddyt:
go/go-brittle-errors. Checking the exact size breaks with this code change or a new go version.

### DIFF
--- a/src/go/configgenerator/route_generator_test.go
+++ b/src/go/configgenerator/route_generator_test.go
@@ -3487,7 +3487,7 @@ func getOverSizeRegexForTest() string {
 	overSizeRegex := ""
 	for i := 0; i < 333; i += 1 {
 		// Form regex in a way that it cannot be simplified.
-		overSizeRegex += "/**/a"
+		overSizeRegex += "[abc]+123"
 	}
 	return overSizeRegex
 }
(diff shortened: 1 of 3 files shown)


src/go/configgenerator/route_generator_test.go
```diff
@@ -3487,7 +3487,7 @@ func getOverSizeRegexForTest() string {
 	overSizeRegex := ""
 	for i := 0; i < 333; i += 1 {
 		// Use "/**" as it is a replacement token for wildcard uri template.
-		overSizeRegex += "/**"
+		overSizeRegex += "/**/a"
```
TAOXUY, line 3490: Why we add an `a` in the end? Could you give explanation? Also it doesn't match the comment?

src/go/util/regex_test.go
```diff
@@ -30,7 +31,7 @@ func TestValidateRegexProgramSize(t *testing.T) {
 			desc:        "oversize regex",
 			regex:       "/**",
 			programSize: 1,
-			wantError:   fmt.Errorf("regex program size(5) is larger than the max expected(1): /**"),
+			wantError:   fmt.Errorf("regex program size"),
```
TAOXUY, line 34: why it couldn't do exact match?

nareddyt: go/go-brittle-errors. Checking the exact size breaks with this code change or a new go version.

src/go/configgenerator/route_generator_test.go
nareddyt: Fixed comment. Changed regex.

Previous regex could be simplified (coalesce `*` characters). New regex cannot be simplified easily, so it will execute the correct branch.

TAOXUY (approved):

PR #532, opened 2021-06-10
Update envoy to 06/09/2021

qiwzhang:
Envoy updated to latest protobuf repo.  It has changed util::Status to match with abls::Status. 

Changed service_control_client_cxx to use the new Status
Changed ESPv2 code to use the new Status.

Also have to fix build breaking [changes](https://groups.google.com/g/envoy-announce/c/MZU4vp3nf5U)
* BUILD files remove @envoy/include with @envoy/
* #include to prefix with "source/"


WORKSPACE
```diff
@@ -39,9 +39,9 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 # 3) Check if envoy_build_config/extensions_build_config.bzl is up-to-date.
 # Try to match it with the one in source/extensions and comment out unneeded extensions.
 
-ENVOY_SHA1 = "98c1c9e9a40804b93b074badad1cdf284b47d58b"  # 2021-05-11: v1.18.3
+ENVOY_SHA1 = "9ebffc61cfd2ea11807757b12a4ae7b92ab0dc8c"  # latest envoy commit June9th
```
shuoyang2016, line 42: Is this a permanent strategy change that we'll always sync to envoy head? Or it's a one-time thing to fulfill the customer's request? I am not sure the the difference between latest envoy vs released envoy version. Any risks for reliability etc?

qiwzhang: This is just one-time thing.  The next envoy release is at the end of month.  This customer could not wait for that long.

In general,  we only sync to the released envoy.

nareddyt (approved):

WORKSPACE
nareddyt: We had similar discussions in the past about syncing to head vs released envoy. In general we should prefer released envoy because security issues will be easier to patch. Other than security patches, we did not find any other risks in reliability.

nareddyt (commented):
I assume the padding issue will be fixed in a follow-up CL?

TAOXUY:
> I assume the padding issue will be fixed in a follow-up CL?

Yes, the change hasn't be submitted in upstream

tests/endpoints/echo/server/app.go
```diff
@@ -350,7 +350,7 @@ func authInfoHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	b, err := base64.StdEncoding.WithPadding(base64.StdPadding).DecodeString(encodedInfo)
```
nareddyt, line 353: Isn't this a breaking change? Can you explain why this is changed here?

nareddyt (changes requested):

tests/endpoints/echo/server/app.go
qiwzhang: Jason reverted padding change in upstream envoy.  this envoy got that change.  So we just revert the behaviors back.  Jason still works on the pr to add a flag to add padding in upstream envoy.  Once that change is in, we can add flag to enable it.

TAOXUY (approved):

nareddyt (approved):

tests/endpoints/echo/server/app.go
nareddyt: Got it. Well I guess reverting the breaking change is OK

varungbt:
@qiwzhang How can I configure this [option ](https://github.com/qiwzhang/envoy/blob/a7f423fe5b7d24c39d7eff7827e8a1f1af18992b/api/envoy/extensions/filters/http/jwt_authn/v3/config.proto#L251)? I do not see any changes in the configmanager flags to make it configurable?

PR #534, opened 2021-06-11
Enable jwks async fetch by default

qiwzhang:

Enable remote jwks async fetch by default and add a flag `disable_jwks_async_fetch` to allow disabling it.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -496,6 +496,16 @@ environment variable or by passing "-k" flag to this script.
         otherwise, it will be rejected. Default is `open`.
         ''')
     parser.add_argument(
+        '--enable_jwks_async_fetch',
+        action='store_true',
+        default=False,
+        help='''
+        Enable JWKS fetching for JWT authentication to be done before processing any requests.
+        If disabled, JWKS fetching is done when processing the requests. Recommended to
+        enable it. The default is disabled to use the old behaviors for not to break any
+        production systems.'''
+    )
+    parser.add_argument(
         '--jwks_cache_duration_in_s',
         default=None,
         help='''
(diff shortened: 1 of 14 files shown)


docker/generic/start_proxy.py
```diff
@@ -495,6 +495,16 @@ def make_argparser():
...
         ''')
+    parser.add_argument(
+        '--enable_jwks_async_fetch',
+        action='store_true',
+        default=False,
+        help='''
+        Enable JWKS fetching for JWT authentication to be done before processing any requests.
+        If disabled, JWKS fetching is done when processing the requests. Recommended to
+        enable it. The default is disabled to use the old behaviors for not to break any
+        production systems.'''
```
nareddyt, line 506: What do you think will break if we enable by default? I do not see much risk.

nareddyt (approved):

docker/generic/start_proxy.py
qiwzhang: My original concern is this case:  If the service config has a jwt_provider with unreachable remote_jwks_uri, but it is not used.    the old way is fine since it is not used, we will never try to fetch it.  In the new way,  we will try to fetch it,  but it fails
My concern is it will block listener to be active.   But I just checked the code,  we still activate listener after it fails.

So the only concern is delayed listener port binding since envoy has to fetch all jwks first before  binding the port. 
Some deployment may care.

qiwzhang: On the second though, if that is the only concern,  I will default this flag to be on

TAOXUY (approved):

qiwzhang:
Changed to enable this feature by default.  

PTAL

TAOXUY (approved):

nareddyt (approved):

PR #538, opened 2021-06-14
b/189965239: Allow using another service account for GCS fetches.

toddbeckman-g:
- Allow choosing between default credentials or a provided service account.
- Use IAM Credentials API to impersonate the provided service account.

Some additional refactoring:
- Streamline client interfaces for easier testing.
- Cover some testing edge cases which were missed.

### DIFF
--- a/src/go/gcsrunner/fetch_config.go
+++ b/src/go/gcsrunner/fetch_config.go
@@ -64,31 +64,28 @@ type FetchConfigOptions struct {
 	FetchGCSObjectTimeout         time.Duration
 }
 
-var (
-	// can be overwritten in unit tests
-	newGCSClient = func(ctx context.Context, sa string) (gcsReader, error) {
-		ts, err := TokenSource(ctx, sa)
-		if err != nil {
-			return nil, err
-		}
-		client, err := storage.NewClient(ctx, option.WithTokenSource(ts))
-		if err != nil {
-			return nil, err
-		}
-		return &gcsClient{client}, nil
+func newGCSClient(ctx context.Context, sa string) (gcsReader, error) {
+	ts, err := TokenSource(ctx, sa)
+	if err != nil {
+		return nil, err
 	}
-	newDefaultCredsGCSClient = func(ctx context.Context) (gcsReader, error) {
-		creds, err := google.FindDefaultCredentials(ctx)
-		if err != nil {
-			return nil, err
-		}
-		c, err := storage.NewClient(ctx, option.WithCredentials(creds))
-		if err != nil {
-			return nil, err
-		}
-		return &gcsClient{c}, nil
+	client, err := storage.NewClient(ctx, option.WithTokenSource(ts))
+	if err != nil {
+		return nil, err
 	}
-)
+	return &gcsClient{client}, nil
+}
+func newDefaultCredsGCSClient(ctx context.Context) (gcsReader, error) {
+	creds, err := google.FindDefaultCredentials(ctx)
+	if err != nil {
+		return nil, err
+	}
+	c, err := storage.NewClient(ctx, option.WithCredentials(creds))
+	if err != nil {
+		return nil, err
+	}
+	return &gcsClient{c}, nil
+}
 
 // FetchConfigFromGCS handles fetching a config from GCS, applying any transformation,
 // and writing it to file.
--- a/src/go/gcsrunner/fetch_config.go
+++ b/src/go/gcsrunner/fetch_config.go
@@ -27,13 +27,36 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/cenkalti/backoff"
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


src/go/gcsrunner/serviceaccount.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+)
+
+// staticTokenSource implements oath2.TokenSource
+type staticTokenSource struct {
+	accessToken string    `json:"access_token"`
+	tokenType   string    `json:"token_type,omitempty"`
```
KyleWiese, line 20: Do we use `tokenType` anywhere?

src/go/gcsrunner/serviceaccount.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (s staticTokenSource) Token() (*oauth2.Token, error) {
+	return &oauth2.Token{
+		AccessToken: s.accessToken,
+		Expiry:      s.expiry,
+	}, nil
+}
+
+type iamcredentialsResponse struct {
```
KyleWiese, line 31: Optional: we could also use https://pkg.go.dev/google.golang.org/api/iamcredentials/v1#GenerateAccessTokenResponse here instead of rolling our own response struct.

src/go/gcsrunner/serviceaccount.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return nil, err
+	}
+	token, err := c.TokenSource.Token()
+	if err != nil {
+		return nil, err
+	}
+	token.SetAuthHeader(req)
+
+	resp, err := http.DefaultClient.Do(req)
```
KyleWiese, line 63: Should we specify an HTTP client with a timeout? e.g 5 or 10 secs

src/go/gcsrunner/serviceaccount.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed iamcredentials request: %v", resp.Status)
```
KyleWiese, line 68: Might be helpful to also print the error body here (if present)

src/go/gcsrunner/main/runner.go
```diff
@@ -52,6 +55,7 @@ var (
 	envoyLogPath = flag.String("envoy_log_path", "",
 		"Envoy application logging path. Default is to write to stderr.")
 	envoyComponentLogLevel = flag.String("envoy_component_log_level", "", "Mapping for Envoy log level by component.")
+	sa                     = flag.String("run_as_service_account", "", "If provided, use this account when fetching the config from GCS.")
```
KyleWiese, line 58: nit: "If provided, use this account when fetching the config from GCS. If not specified, the instance's default credentials will be used. "

src/go/gcsrunner/fetch_config_test.go
```diff
@@ -15,147 +15,193 @@
...
 		},
 		{
-			name:                    "success with retries",
-			wantBytes:               output,
-			wantMultipleReaderCalls: true,
-			passMultipleReaderCalls: true,
-			newReaderReturns:        bytes.NewReader(output),
-			newReaderErr:            fmt.Errorf("temporary reader error should resolve"),
-			credsFunc:               goodFindDefaultCredentials,
+			name:           "timeout retrying newGSC",
```
KyleWiese, line 160: s/newGSC/newGCS

src/go/gcsrunner/fetch_config.go
```diff
@@ -27,52 +27,67 @@ import (
...
+	ConfigFileName                string
+	WriteFilePath                 string
+	FetchGCSObjectInitialInterval time.Duration
+	FetchGCSObjectTimeout         time.Duration
 }
 
 var (
-	osCreate = func(p string) (file, error) { return os.Create(p) }
+	// can be overwritten in unit tests
+	newGCSClient = func(ctx context.Context, sa string) (gcsReader, error) {
```
KyleWiese, line 69: optional: Instead of making these function vars, they could just be functions e.g

```
var (
 gcsCient = newGCSClient
)
func newGCSClient(...)...
```

src/go/gcsrunner/serviceaccount.go
toddbeckman-g: The struct here versus in the iamcredentials API is functionally identical anyway, so I don't think we gain much by including this dependency. The context here is that my attempts to use the iamcredentials API resulted in odd errors I couldn't debug, and when I stepped away from using the API at all, everything just worked™.

KyleWiese (approved):

TAOXUY (approved):

PR #539, opened 2021-06-15
Update version number to 2.28.0.

TAOXUY:
2.27.0 -> 2.28.0


nareddyt (changes requested):
Can we change the release process so we update the version number right before a release, instead of after?

The reason we did not update it last time is because it won't work for patch release. For example, you updated to v2.29.0 today. If we need to make v2.28.1 next week, the version will be wrong. It will send the wrong report to SC v2 report.

nareddyt:
So for now, let's just update to 2.28.0 and use this version number of the release

qiwzhang:
Then the step will be:
1) change the version
2) make a build so the binary will have the correct version. 
3) use that build for the release.

SGTM

nareddyt:
> Then the step will be:
> 
> 1. change the version
> 2. make a build so the binary will have the correct version.
> 3. use that build for the release.
> 
> SGTM

I did not think about "make a build so the binary will have the correct version". That will add a delay of 1 day to the release for all release qualification tests to run.

WDYT @TAOXUY

nareddyt (approved):

qiwzhang (approved):

TAOXUY:
> > Then the step will be:
> > 
> > 1. change the version
> > 2. make a build so the binary will have the correct version.
> > 3. use that build for the release.
> > 
> > SGTM
> 
> I did not think about "make a build so the binary will have the correct version". That will add a delay of 1 day to the release for all release qualification tests to run.
> 
> WDYT @TAOXUY? I guess even today, you cannot make a release until the version number is correct (you need a build with version number 2.28.0).

I see the point

I would suggest we just use the old commit to run the qualification tests and release the new one with version update change. The oncaller still need to wait for the new one check in but not its finishing qulification test. Other than this step, the rest should be not effected.

qiwzhang:
I don't think it will take a day to build a release.  building itself should take <30minutes

It is the qualification process that is taking long time.  We can 
1) qualify a SHA, top of master tree
2) then make a version change, submit the change, use prow to build it
3) use that for the release. 

Just need to make sure:
 released SHA =  qualified SHA + version_only_change

nareddyt:
> I would suggest we just use the old commit to run the qualification test verification and release the new one with version update change. The oncaller still need to wait for the new one check in but not its finishing qulification test(1day). Other than this step, the rest should be not effected.

SGTM. The only other thing to be aware of is the release branch name. https://github.com/GoogleCloudPlatform/esp-v2/blob/e23a5d27c19f9c487a66e1a5e3d1cb644d998980/scripts/release/release-new-branch.sh#L52

Issue #545, opened 2021-06-16
ESPv2.26+ disregards grpcs backend scheme to IAP protected backends


someone1:
Upgrading from 2.24 to 2.28, services that point to a `grpcs` scheme backend protected by IAP are trying to do so via HTTP and get a 302 response to redirect to HTTPS.

Debug log of request from ESP:
```
D0616 21:35:58.758 27 envoy] [27][router][C6][S15140320129877370655] router decoding headers:
':method', 'POST'
':scheme', 'http'
':authority', '<REMOVED>'
':path', '/<REMOVED>'
'user-agent', 'curl/7.64.0'
'accept', '*/*'
'x-forwarded-proto', 'https'
'x-forwarded-for', '<REMOVED>'
'forwarded', 'for=\"<REMOVED>\";proto=https'
'authorization', 'Bearer <REMOVED>'
'x-request-id', 'a3f25a21-46c3-4645-b84a-89e01136b675'
'x-endpoint-api-userinfo', '<REMOVED>'
'content-type', 'application/grpc'
'x-envoy-original-path', '/<REMOVED>'
'x-envoy-original-method', 'POST'
'te', 'trailers'
'x-forwarded-authorization', 'Bearer <REMOVED>'
'x-cloud-trace-context', 'c29c31729541ce2bca41c1bc51b79409/1024743661290867553;o=0' 
D0616 21:35:58.758 27 envoy] [27][pool][C8] using existing connection
D0616 21:35:58.758 27 envoy] [27][pool][C8] creating stream
D0616 21:35:58.758 27 envoy] [27][router][C6][S15140320129877370655] pool ready
D0616 21:35:58.768 27 envoy] [27][router][C6][S15140320129877370655] upstream headers complete: end_stream=false
D0616 21:35:58.768 27 envoy] [27][filter]Filter::encodeHeaders is called.
D0616 21:35:58.768 27 envoy] [27][http][C6][S15140320129877370655] encoding headers via codec (end_stream=false):
':status', '302'
'location', '<REMOVED>'
'x-goog-iap-generated-response', 'true'
'grpc-message', 'Redirecting to HTTPS.'
'grpc-status', '16'
'content-length', '21'
'content-type', 'text/html; charset=UTF-8'
'date', 'Wed, 16 Jun 2021 21:35:58 GMT'
'x-envoy-upstream-service-time', '9'
'server', 'envoy'
'x-envoy-decorator-operation', 'ingress <REMOVED>' 
[...]

PR #546, opened 2021-06-18
Disable preserve downstream scheme

qiwzhang:
This is to fix https://github.com/GoogleCloudPlatform/esp-v2/issues/545

Envoy new way is "preserve_downstream_scheme".  This is not working for ESPv2 deployed in Cloud-Run.  Especially if its backend is using IAP.   IAP requires ":scheme" to be https.   But "preserve_downstream_scheme" will change it to http since the connection between Cloud-Run app-sever and ESPv2 is http.

### DIFF
--- a/examples/auth/envoy_config.json
+++ b/examples/auth/envoy_config.json
@@ -5,7 +5,8 @@
       {
         "name": "static-runtime",
         "staticLayer": {
-          "envoy.reloadable_features.preserve_downstream_scheme": false
+          "envoy.reloadable_features.preserve_downstream_scheme": false,
+          "re2.max_program_size.error_level": 1000
         }
       }
     ]
--- a/examples/auth/envoy_config.json
+++ b/examples/auth/envoy_config.json
@@ -3,9 +3,9 @@
   "layeredRuntime": {
     "layers": [
       {
-        "name": "deprecation",
+        "name": "static-runtime",
         "staticLayer": {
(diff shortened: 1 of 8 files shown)


qiwzhang:
E2E test passed.   Even downstream `:scheme` is http, but upstream `:scheme` is https.  

So  it works

PTAL

TAOXUY:
> E2E test passed. Even downstream `:scheme` is http, but upstream `:scheme` is https.
> 
> So it works
> 
> PTAL

What do you mean by e2e test passed? I thought we didn't have tests covering this case. If not, is it easy to add one?

qiwzhang:
I manually tested the E2E test with grpc-echo deployed in Cloud-Run,  I checked the envoy debug log,  scheme set to "https" properly.

It is not easy to add such E2E test,   we need a gRPC E2E test with IAP enabled to verify IAP still work.  In server-less platform only app-engine supports gRPC, but it doesn't support gRPC.

qiwzhang:
Or like  https://github.com/GoogleCloudPlatform/esp-v2/issues/545,  we need to setup backend as GCE Instance Group, and set it up behind GCLB, and setup IAP in the GCLB.  gRPC is supported in such setup. 

File b/191431028 to such task

TAOXUY (approved):

nareddyt (approved):

PR #549, opened 2021-06-25
Rename a few methods visibility in configgenerator.

shuoyang2016:
These methods will be later used by other pkgs.

### DIFF
--- a/src/go/configgenerator/route_generator.go
+++ b/src/go/configgenerator/route_generator.go
@@ -38,7 +38,7 @@ const (
 	virtualHostName = "backend"
 )
 
-func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfiguration, error) {
+func makeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfiguration, error) {
 	var virtualHosts []*routepb.VirtualHost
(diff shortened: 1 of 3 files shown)


nareddyt (approved):

src/go/configgenerator/route_generator.go
```diff
@@ -38,7 +38,7 @@ const (
 	virtualHostName = "backend"
 )
 
-func MakeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfiguration, error) {
+func makeRouteConfig(serviceInfo *configinfo.ServiceInfo) (*routepb.RouteConfiguration, error) {
```
nareddyt, line 41: Is changing to private really needed? Are you sure you will never call this directly?

TAOXUY (approved):

src/go/configgenerator/route_generator.go
shuoyang2016: Good question. I think changing this to private is only for best practice. Functionality wise it's same. IMO it's better to keep these private until we really need it. Like MakeRouteTable we just made it public. I was curious why this name is public and the rest are private but in fact this is never used outside of the package :)

qiwzhang (approved):

PR #552, opened 2021-06-29
Add operation name as route name for backend route.

shuoyang2016:
Also refactor the listener_genarator code to expose the AddPerRouteConfigGenToMethods.


src/go/configgenerator/listener_generator.go
```diff
@@ -83,9 +86,18 @@ func MakeListener(serviceInfo *sc.ServiceInfo, filterGenerators []*filterconfig.
 					return nil, err
 				}
 			}
-
 		}
 	}
+	return nil, nil
```
nareddyt, line 91: Return httpFilters?

src/go/configgenerator/route_generator.go
```diff
@@ -403,6 +403,7 @@ func makeRoute(routeMatcher *routepb.RouteMatch, method *configinfo.MethodInfo)
 	}
 
 	return &routepb.Route{
```
nareddyt, line 405: There are many other places where a route is constructed in this file. CORS, etc. Do control-f for `&routepb.Route{` just to see a few.

shuoyang2016: Looks like the cors route doesn't associate with an method info? Is that so? In this case, what should be the route name?

shuoyang2016:
Finally fixed all testdata. I didn't anticipate there are so many testdata which took me hours to complete.

src/go/configgenerator/listener_generator.go
shuoyang2016: Thanks! Done.

shuoyang2016:
PTAL

qiwzhang (approved):

nareddyt:
> Finally fixed all testdata. I didn't anticipate there are so many testdata which took me hours to complete.

This is exactly why I didn't add unit tests to our config generator in google3 beforehand :) I want to find a better way to do testing internally. Something like this: https://ieftimov.com/post/testing-in-go-golden-files/#keeping-our-golden-files-up-to-date

src/go/configgenerator/route_generator.go
nareddyt: Ok, that's fine then.

src/go/configgenerator/route_generator_test.go
```diff
@@ -71,121 +71,123 @@ func TestMakeRouteConfig(t *testing.T) {
 			},
 			wantRouteConfig: `
 {
-    "name": "local_route",
-    "virtualHosts": [
+  "name": "local_route",
```
nareddyt, line 74: nit: did you change indentation? diff is big

nareddyt (approved):

src/go/configmanager/testdata/test_fixed_mode_dynamic_routing.go
```diff
@@ -599,6 +608,7 @@ var (
                         ],
                         "path": "/search"
                       },
+											"name": "1.echo_api_endpoints_cloudesf_testing_cloud_goog.dynamic_routing_Search",
```
TAOXUY, line 611: NIT, indent

src/go/configmanager/testdata/test_fetch_listeners.go
```diff
@@ -2457,6 +2486,7 @@ var (
                         ],
                         "path": "/simplegetcors"
                       },
+											"name": "1.echo_api_endpoints_cloudesf_testing_cloud_goog.Simplegetcors",
```
TAOXUY, line 2489: ditto and for the rest

TAOXUY (approved):

src/go/configgenerator/route_generator_test.go
shuoyang2016: Yeah, I used go/jsonformatter to pretty print the json which uses two space for indent I think.

src/go/configmanager/testdata/test_fixed_mode_dynamic_routing.go
shuoyang2016: Done. I didn't notice this. My intellij's indent showed two spaces.

src/go/configmanager/testdata/test_fetch_listeners.go
shuoyang2016: Done.

shuoyang2016:
> This is exactly why I didn't add unit tests to our config generator in google3 beforehand :) I want to find a better way to do testing internally. Something like this: https://ieftimov.com/post/testing-in-go-golden-files/#keeping-our-golden-files-up-to-date

That makes senses. Another difficult I had is there's no per line diff. I have to copy the entire json and pretty print. If we can use cmp.diff, that would make things a bit better too.

PR #554, opened 2021-06-30
b/192247795: support backendRetryOnStatusCodes

TAOXUY:
Support backend retry on status codes.

**implementation details**:
- add `--backend_retry_on_status_codes` to [config.route.v3.RetryPolicy.retriable_status_codes](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-retrypolicy)
- add `retriabel-status-code` into [config.route.v3.RetryPolicy.retry_on](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-retrypolicy) if it is not set.

### DIFF
--- a/src/go/configinfo/service_info.go
+++ b/src/go/configinfo/service_info.go
@@ -17,6 +17,7 @@ package configinfo
 import (
 	"fmt"
 	"math"
+	"strconv"
 	"strings"
 	"time"
 
@@ -148,6 +149,9 @@ func NewServiceInfoFromServiceConfig(serviceConfig *confpb.Service, id string, o
 		return nil, err
 	}
 	if err := serviceInfo.processLocalBackendOperations(); err != nil {
+		return nil, err
+	}
+	if err := serviceInfo.processAllBackends(); err != nil {
 		return nil, err
 	}
 	if err := serviceInfo.processAuthRequirement(); err != nil {
@@ -594,9 +598,6 @@ func (s *ServiceInfo) addBackendInfoToMethod(r *confpb.BackendRule, scheme strin
 		TranslationType: r.PathTranslation,
 		Deadline:        deadline,
 		IdleTimeout:     idleTimeout,
-		RetryOns:        s.Options.BackendRetryOns,
-		RetryNum:        s.Options.BackendRetryNum,
-		PerTryTimeout:   s.Options.BackendPerTryTimeout,
 	}
 
 	jwtAud := s.determineBackendAuthJwtAud(r, scheme, hostname)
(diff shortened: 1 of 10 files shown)


src/go/configinfo/service_info.go
```diff
@@ -891,3 +922,16 @@ func calculateStreamIdleTimeout(operationDeadline time.Duration, opts options.Co
 	operationIdleTimeout := operationDeadline + time.Second
 	return util.MaxDuration(operationIdleTimeout, opts.StreamIdleTimeout)
 }
+
+func parseStatusCodes(statusCodes string) ([]uint32, error) {
+	codeList := strings.Split(statusCodes, ",")
+	var codes []uint32
+	for _, codeStr := range codeList {
+		if code, err := strconv.Atoi(codeStr); err != nil {
```
qiwzhang, line 930: should we verify status code is in legal range: "> 300 and < 600"?

TAOXUY: Done. Made valid http status code check [100, 600).

src/go/configinfo/service_info.go
```diff
@@ -630,6 +631,39 @@ func (s *ServiceInfo) determineBackendAuthJwtAud(r *confpb.BackendRule, scheme s
...
+// Apply global setting to all the backends.
+func (s *ServiceInfo) processAllBackends() error {
+	for _, method := range s.Methods {
+		backendInfo := method.BackendInfo
+		if backendInfo == nil {
+			return fmt.Errorf("all the methods should have an un-empty BackendInfo")
+
+		}
+
+		backendInfo.RetryOns = s.Options.BackendRetryOns
```
qiwzhang, line 643: It seems all these "retry" fields are global,  not need to be per "backend" or per "method".   Can we define them in s.Options?

nareddyt: I guess the future design is that retry will be configurable in the service config instead of flag, so then it will be per backend. It's good to keep this design.

src/go/configinfo/service_info.go
```diff
@@ -891,3 +922,16 @@ func calculateStreamIdleTimeout(operationDeadline time.Duration, opts options.Co
 	operationIdleTimeout := operationDeadline + time.Second
 	return util.MaxDuration(operationIdleTimeout, opts.StreamIdleTimeout)
 }
+
+func parseRetriableStatusCodes(statusCodes string) ([]uint32, error) {
+	codeList := strings.Split(statusCodes, ",")
+	var codes []uint32
+	for _, codeStr := range codeList {
+		if code, err := strconv.Atoi(codeStr); err != nil || code < 100 || code >= 600 {
+			return nil, fmt.Errorf("invalid http status codes: %v, the valid one should be a number in [100, 600)", code)
```
nareddyt, line 931: nit: put `err` in the message

nareddyt (approved):

src/go/configinfo/service_info.go
TAOXUY: This error `strconv.Atoi: parsing "invalid-status-code": invalid syntax` is not that informative   and is covered in `the valid one should be a number in [100, 600)`. So prefer not.

src/go/configinfo/service_info.go
TAOXUY: Yes, we would like to be open for the per backend retry policy.

qiwzhang (approved):

shuoyang2016 (approved):

PR #560, opened 2021-07-12
b/182218056: support downstream mTLS

TAOXUY:
**context:** current we already have supported downstream TLS, upstream mTLS and sidestream TLS.

**content:**
- support downstream client cert-based auth to make downstream mTLS
- add unit tests and integration test
- fix our cert generating make target


docker/generic/start_proxy.py
```diff
@@ -182,6 +182,13 @@ def make_argparser():
         Cipher suites to use for downstream connections as a comma-separated list.
         Please refer to https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/common.proto#auth-tlsparameters''')
 
+    parser.add_argument('--ssl_server_root_cert_path', default=None, help='''
+         The file path of root certificates that ESPv2 uses to do downstream client certificate-based authentication.
+        If not specified, ESPv2 doesn't it by default.
```
qiwzhang, line 187: s/ESPv2 doesn't it by default/ESPv2 doens't verify client certificates/?

docker/generic/start_proxy.py
```diff
@@ -182,6 +182,13 @@ def make_argparser():
         Cipher suites to use for downstream connections as a comma-separated list.
         Please refer to https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/common.proto#auth-tlsparameters''')
 
+    parser.add_argument('--ssl_server_root_cert_path', default=None, help='''
```
qiwzhang, line 185: mTLS means:
* verify client certificates
* specify which clients are allowed.  Do we want to support that? or it is in a separate pr?

src/go/configmanager/flags/flags.go
```diff
@@ -60,6 +60,7 @@ var (
 
 	SslServerCertPath                = flag.String("ssl_server_cert_path", "", "Path to the certificate and key that ESPv2 uses to act as a HTTPS server")
 	SslServerCipherSuites            = flag.String("ssl_server_cipher_suites", "", "Cipher suites to use for downstream connections as a comma-separated list.")
+	SslServerRootCertsPath           = flag.String("ssl_server_root_cert_path", "", "The file path of root certificates that ESPv2 uses to verify downstream client certificate. If not specified, ESPv2 doesn't the validation by default.")
```
qiwzhang, line 63: ESPv2 doens't verify client certificate.

docker/generic/start_proxy.py
timothyleung1: Not sure if this is better: The file path of root certificates that ESPv2 uses to verify downstream client certificate.

src/go/util/transport_socket_test.go
```diff
@@ -188,6 +189,42 @@ func TestCreateDownstreamTransportSocket(t *testing.T) {
...
 			} `,
 		},
+		{
+			desc:               "Downstream Transport Socket for mTLS",
+			sslPath:            "/etc/ssl/endpoints/",
+			sslRootCertPath:    "/etc/ssl/endpoints/root.crt",
+			sslMinimumProtocol: "TLSv1.1",
+			wantTransportSocket: `{
+				"name": "envoy.transport_sockets.tls",
+				"typedConfig": {
```
timothyleung1, line 199: Im not 100% familar with ESPv2 yet, but should there be a `require_client_certificate` flag in the config? https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/securing

timothyleung1 (commented):
Thanks for working on this!

src/go/util/transport_socket_test.go
timothyleung1: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-certificatevalidationcontext And are we going to support all kind of client auth or just trustedCA?

docker/generic/start_proxy.py
TAOXUY: Done

src/go/configmanager/flags/flags.go
TAOXUY: Done.

docker/generic/start_proxy.py
```diff
@@ -182,6 +182,13 @@ def make_argparser():
...
+         The file path of root certificates that ESPv2 uses to do downstream client certificate-based authentication.
+        If not specified, ESPv2 doesn't it by default. 
+        
+        Before using this feature, please verify if your compute environment is compatible with it.
```
qiwzhang, line 189: what computer environment?   how to verify?, verify what?

src/go/configmanager/flags/flags.go
qiwzhang: I still see this "ESPv2 doesn't the validation by default".  This is not right grammar

docker/generic/start_proxy.py
TAOXUY: You are right. This PR, like our existing TLS connections, only do the basic server certs validation/authn. As for tls access control, envoy does have a bunch of other [options](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-certificatevalidationcontext) to do access control or extra validation. I am thinking we make another ones if necessary after asking them what they need. We can even encourage to contribute based on this PR.

src/go/util/transport_socket_test.go
TAOXUY: `require_client_certifacte` added. It is interesting that seems this field doesn't work and  client certificate  is enforced. As I didn't set it before, my test would raise error if the client doesn't provide certs, which fits https://github.com/envoyproxy/envoy/issues/17224.

Yes, this PR only provides basic client cert authn for this PR and for those [options](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-certificatevalidationcontext), what do you need?

tests/integration_test/transport_security_test/transport_security_test.go
```diff
@@ -162,73 +163,182 @@ func TestServiceControlWithTLS(t *testing.T) {
...
 		},
 		{
-			desc:         "Succcess for HTTP2 client with TLS",
-			httpsVersion: 2,
+			desc:                  "failure(client authn),  client cert is unknown to proxy's root cert",
+			scheme:                "http",
+			httpVersions:          []int{1},
+			sslServerRootCertPath: platform.GetFilePath(platform.DownstreamClientCert),
+			getClientCred: func(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+				c, err := tls.LoadX509KeyPair(platform.GetFilePath(platform.MismatchCert), platform.GetFilePath(platform.MismatchKey))
```
qiwzhang, line 198: It seems the code for getClientCred are all the same,  just different pairs of client_cert files.  Maybe moving this code into client.DoHttpsGet

TAOXUY: Done.

docker/generic/start_proxy.py
TAOXUY: Just reworded. I wanna warn users that many GCP platforms will terminate TSL/mTSL in ingress.

src/go/util/transport_socket_test.go
timothyleung1: im having a meeting with partner team on Wednesday and ill raise that point. but i guess this CL is a good start already, it should be easy to add new options in the future.

qiwzhang (approved):

src/go/util/transport_socket_test.go
TAOXUY: SG

docker/generic/start_proxy.py
```diff
@@ -176,12 +176,28 @@ def make_argparser():
     parser.add_argument('--ssl_server_cert_path', default=None, help='''
         Proxy's server cert path. When configured, ESPv2 only accepts HTTP/1.x and
         HTTP/2 secure connections on listener_port. Requires the certificate and
-        key files "server.crt" and "server.key" within this path.''')
+        key files "server.crt" and "server.key" within this path.
+        
+        Before using this feature, please make sure TLS isn't terminated before ESPv2
+        in your deployment model. In general, Cloud Run, GKE(GCLB enforced in ingress)
```
nareddyt, line 182: GKE with container-native load balancing

tests/env/platform/ports.go
```diff
@@ -32,12 +32,12 @@ const (
 	TestAddHeaders
```
nareddyt, line 32: Can you add a comment on how you are sorting these automatically?

tests/env/testdata/mismatch.crt
```diff
@@ -0,0 +1,23 @@
+-----BEGIN CERTIFICATE-----
```
nareddyt, line 1: Need some information on how these mismatch certs were generated. I cannot tell what the "mismatch" is looking at this file

tests/integration_test/transport_security_test/transport_security_test.go
```diff
@@ -162,73 +162,152 @@ func TestServiceControlWithTLS(t *testing.T) {
...
-			t.Errorf("Test (%s): failed\nexpected: %v\ngot: %v", tc.desc, tc.wantError, err)
-		}
+
+			s := env.NewTestEnv(platform.TestDownstreamMTLS, platform.EchoSidecar)
+			defer s.TearDown(t)
+			if err := s.Setup(args); err != nil {
+				t.Fatalf("fail to setup test env, %v", err)
+			}
+
+			// FIXME: Use of localhost. Difficult to generate certs with ip addresses.
```
nareddyt, line 288: Yeah this will not work in google3 easily. That's fine for now.

nareddyt (approved):

docker/generic/start_proxy.py
TAOXUY: Done.

tests/env/platform/ports.go
TAOXUY: I wish I could but it is done by IDE.

tests/env/testdata/mismatch.crt
TAOXUY: Mismatched crt is used in the case when the server crt is unknown to the client root ca. Before, we used proxy.crt to replace server.crt or vice versa to make this case but it is confusing.

PR #575, opened 2021-07-31
b/194945474: Unify route match policy with CORS filter

nareddyt:
Add stronger guarantees on the `Origin` header match for a CORS preflight request. With this change, the route table will reject requests that have a mismatching `Origin`.


qiwzhang (approved):

TAOXUY (approved):

nareddyt:
Ignoring `ESPv2-API-regression-test` because error messages are not part of the API surface

PR #586, opened 2021-08-17
b/195940890: Improve gcsrunner remote dependency handling

syangabq:
-Moved IAM Creds fetching into the backoff/retry loop
-Added logging for the duration of each individual RPC in addition to the overall end-to-end latency.
-Added logging for every external dependency error

### DIFF
--- a/src/go/gcsrunner/fetch_config.go
+++ b/src/go/gcsrunner/fetch_config.go
@@ -91,10 +91,10 @@ func newDefaultCredsGCSClient(ctx context.Context) (gcsReader, error) {
 	start = time.Now()
 	c, err := storage.NewClient(ctx, option.WithCredentials(creds))
 	if err != nil {
-		glog.Errorf("error creating new GCS client with credentials: %v", err)
+		glog.Errorf("error creating new GCS client with default credentials: %v", err)
 		return nil, err
 	}
-	glog.Infof("created new GCS client with credentials in %s", time.Since(start))
+	glog.Infof("created new GCS client with default credentials in %s", time.Since(start))
 	return &gcsClient{c}, nil
 }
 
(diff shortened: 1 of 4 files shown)


src/go/gcsrunner/fetch_config.go
```diff
@@ -109,31 +120,38 @@ func readBytes(opts FetchConfigOptions) ([]byte, error) {
...
+				glog.Errorf("error getting default creds GCS client (retrying): %v", err)
+				return err
+			}
+		} else {
+			client, err = newGCS(ctx, opts.ServiceAccount)
+			if err != nil {
+				glog.Errorf("error getting GCS client using service account (retrying): %v", err)
+				return err
+			}
+		}
```
emschwar, line 145: Won't this create a new client every time we retry? It might be preferred to move the `var client` outside the op, and that that it's non-nil before trying to create a new client each time.

syangabq: Added commit to do this

src/go/gcsrunner/fetch_config.go
```diff
@@ -110,30 +121,40 @@ func readBytes(opts FetchConfigOptions) ([]byte, error) {
...
-	}
+
 	ebo := backoff.NewExponentialBackOff()
 	ebo.InitialInterval = opts.FetchGCSObjectInitialInterval
 	var out []byte
 	op := func() error {
 		if err := ctx.Err(); err != nil {
 			return backoff.Permanent(err)
 		}
+
```
nareddyt, line 132: nit: consider refactoring into separate function. Lots of nesting makes it hard to review all code paths and ensure `client != nil` by line 151.

nareddyt (approved):

src/go/gcsrunner/fetch_config.go
syangabq: Done.

qiwzhang:
@TAOXUY  could you help to review it? and help to merge it?  Thanks

nareddyt (approved):

TAOXUY (approved):

PR #596, opened 2021-09-08
b/195940890: Log all levels to stderr in gcsrunner

syangabq:


nareddyt (approved):
Just FYI, for esp-v2 we pass the flag `--alsologtostderr` so all log levels (INFO - ERROR) go to Cloud Logging.

nareddyt:
cc @emschwar @KyleWiese

src/go/gcsrunner/main/runner.go
```diff
@@ -59,7 +60,19 @@ var (
 )
 
 func main() {
+	err := setGcsrunnerLogLevel(gcsrunnerDefaultLogLevel); if err != nil {
+		glog.Errorf("Failed to override flag --stderrthreshold to default log level %s: %v", gcsrunnerDefaultLogLevel, err)
+	}
+
 	flag.Parse()
+	
+	gcsrunnerLogLevel := os.Getenv("GCSRUNNER_LOG_LEVEL")
```
emschwar, line 69: Instead of calling `setGcsrunnerLogLevel` twice, consider teaching it to handle an empty value, and then call it once with the output of `os.Getenv` here. Something like:

```go
if len(logLevel) == 0 {
  logLevel = gcsrunnerDefaultLogLevel
}
return flag.Set(...)
```

KyleWiese (approved):
LGTM

nareddyt (approved):

Issue #600, opened 2021-09-13
A lot of deprecation warnings on startup about using `envoy.config.route.v3.HeaderMatcher.exact_match`


daaain:
I'm in the process of migrating from the ESP v1 image with GKE OpenAPI Cloud Endpoints and managed to get v2 working, but getting about a 100 or so of these deprecation warnings on startup.

Just wanted to check if this is something you're aware of? Is it because of some configuration I need to update? If not, is it something that needs to be fixed internally in this image?

Thanks in advance!

```
W0913 16:48:08.571 17 envoy] [17][misc]Deprecated field: type envoy.config.route.v3.HeaderMatcher Using deprecated option \'envoy.config.route.v3.HeaderMatcher.exact_match\' from file route_components.proto. This configuration will be removed from Envoy soon. Please see https://www.envoyproxy.io/docs/envoy/latest/version_history/version_history for details. If continued use of this field is absolutely necessary, see https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime#using-runtime-overrides-for-deprecated-features for how to apply a temporary and highly discouraged override.
```

qiwzhang:
Yes,  we will fix these warning.  Thanks for reporting.  

-Wayne

PR #601, opened 2021-09-14
Replace deprecated HeaderMatch_ExactMatch

qiwzhang:

To gix https://github.com/GoogleCloudPlatform/esp-v2/issues/600

Envoy this [PR](https://github.com/envoyproxy/envoy/pull/17119)  added StringMatch in router.HeaderMatch and deprecated ExactMatch.

Also updated to the latest envoyproxy/go-control-plane.  It has some changes in "cache"

### DIFF
--- a/src/go/configmanager/config_manager.go
+++ b/src/go/configmanager/config_manager.go
@@ -16,6 +16,7 @@ package configmanager
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"flag"
@@ -36,6 +37,7 @@ import (
 	gen "github.com/GoogleCloudPlatform/esp-v2/src/go/configgenerator"
 	sc "github.com/GoogleCloudPlatform/esp-v2/src/go/serviceconfig"
 	corepb "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	rsrc "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
 	confpb "google.golang.org/genproto/googleapis/api/serviceconfig"
 )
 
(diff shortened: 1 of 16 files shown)


src/go/configmanager/config_manager.go
```diff
@@ -249,13 +251,13 @@ func (m *ConfigManager) applyServiceConfig(serviceConfig *confpb.Service) error
...
 	}
-	return m.cache.SetSnapshot(m.envoyConfigOptions.Node, *snapshot)
+	return m.cache.SetSnapshot(context.Background(), m.envoyConfigOptions.Node, *snapshot)
 }
 
 func (m *ConfigManager) makeSnapshot() (*cache.Snapshot, error) {
 	m.Infof("making configuration for api: %v", m.serviceInfo.Name)
 
-	var clusterResources, endpoints, secrets, runtimes, routes, listenerResources []types.Resource
+	var clusterResources, listenerResources []types.Resource
```
nareddyt, line 260: Where are the other resources, like routes?

nareddyt (approved):

src/go/configmanager/config_manager.go
qiwzhang: They are empty,  we don't generate other resources beside listeners and clousters.   I believe we don't need to set empty resources to NewSnapshot.

TAOXUY (approved):

nareddyt:
Hi @qiwzhang, can you double check the flaky test? Is it not flaky on our periodic jobs: https://testgrid.k8s.io/googleoss-esp-v2-periodic#presubmit. Though this PR should not cause the test failure.

PR #611, opened 2021-09-24
b/185372123: Use alpine as base image

qiwzhang:

This reduces the docker image size from 1.5GB to 500MB.  
Most importantly,  it eliminates all security vulnerabilities.  Here is the trivy scan result

```
gcr.io/cloudesf-testing/apiproxy-serverless:698f6b245df43f1fa40f8969dc461a07cf6a4937 (alpine 3.14.0)
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
```

### DIFF
--- a/docker/Dockerfile-envoy
+++ b/docker/Dockerfile-envoy
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.6-buster
+FROM frolvlad/alpine-glibc:alpine-3.14_glibc-2.33
 
 # Set the default root CA for gRpc.
 # opensensus lib is using gRpc to call Stackdriver
--- a/docker/Dockerfile-envoy
+++ b/docker/Dockerfile-envoy
@@ -14,6 +14,9 @@
 
 FROM frolvlad/alpine-glibc:alpine-3.14_glibc-2.33
 
+# Update to the latest
+RUN apk update && apk upgrade
+
 # Set the default root CA for gRpc.
 # opensensus lib is using gRpc to call Stackdriver
 ENV GRPC_DEFAULT_SSL_ROOTS_FILE_PATH /etc/ssl/certs/ca-certificates.crt
--- a/docker/Dockerfile-proxy.tmpl
+++ b/docker/Dockerfile-proxy.tmpl
@@ -14,6 +14,12 @@
 
 FROM _ENVOY_IMAGE_SHA_NAME
 
+# Install python3
+ENV PYTHONUNBUFFERED=1
+RUN apk add --update --no-cache python3 && ln -sf python3 /usr/bin/python
+RUN python3 -m ensurepip
+RUN pip3 install --no-cache --upgrade pip setuptools
+
 ENV PATH /bin:$PATH
 
 ADD docker/generic/* /apiproxy/
@@ -25,5 +31,4 @@ RUN groupadd -g 999 envoy && useradd -r -u 999 -g envoy envoy
 # run as user envoy
 USER envoy
 
-
-ENTRYPOINT ["python", "/apiproxy/start_proxy.py"]
+ENTRYPOINT ["python3", "/apiproxy/start_proxy.py"]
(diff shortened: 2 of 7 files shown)


qiwzhang:
PTAL

TAOXUY (approved):

docker/Dockerfile-envoy
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.6-buster
+FROM frolvlad/alpine-glibc:alpine-3.14_glibc-2.33
```
shuoyang2016, line 15: Which team owns and maintains this image?

nareddyt: +1, are you ok introducing a dependency on a community maintained image?

FYI license for image is OK, it is MIT: https://github.com/Docker-Hub-frolvlad/docker-alpine-glibc/blob/master/LICENSE

But please check go/cloudimage. I am not sure if this falls under the pre-approved list. Might be worth reaching out to OSPO.

Another approach is to consider forking https://github.com/Docker-Hub-frolvlad/docker-alpine-glibc/blob/master/Dockerfile and building the image ourself, with the proper attribution to the source license and author. Please check with OSPO if this approach is allowed.

nareddyt (changes requested):

docker/Dockerfile-envoy
qiwzhang: I decided to just use the one build by frolvlad/alpine-glibc:  it has a community maintaining this, to upgrade and fix some bugs. it is better than we fork it

I added its LICENSE file in. It is MIT license.

shuoyang2016: Depending on community image is OK as long as we have a maintainability story moving forward. I guess we'll just update this image from time to time after the community lands a new version?

docker/Dockerfile-envoy
```diff
@@ -12,7 +12,10 @@
...
+
+# Update to the latest
+RUN apk update && apk upgrade
```
shuoyang2016, line 18: Will this cause the build to be non-deterministic?

qiwzhang: Yes, but only at docker image build time.  
I think it is an acceptable practice to do "RUN apt -y update && apt -y install xyz" to get the latest packages.

shuoyang2016: I think it's generally better practice to make the build deterministic so we can find out if one issue appears one build. If the community continuously qualify new images, we can just update our images to the new stable ones from time to time. What's the benefit we get the latest (may not be qualified?) packages ourselves?

nareddyt: +1 to @qiwzhang's comment, updating to latest packages at build time is acceptable for docker images. It is better we keep them up to date to receive security patches. If issue appears, e2e test will fail, and we will know.

nareddyt (approved):

docker/Dockerfile-envoy
shuoyang2016: I was imagining we just need to update the stable alpine version from time to time. Like https://github.com/envoyproxy/envoy/commit/ed56db4e23e2cce943252b63c2e68be6abe5215e. 

If you guys think this is OK, I am also fine with it. Since this looks low risk to me. It's just the non-repeatable build part worries me in terms of reproducing the same artifacts and signature. 

Is this equivalent to use frolvlad/alpine-glibc:latest?

qiwzhang: It is similar to "frolvlad/alpine-glibc:alpine-3.14_glibc-2.33" the latest.   But we need specific glibc > 2.28 (required by envoy).  they don't have "latest" tag for a specific one.

shuoyang2016: SG. Thanks for the explain. That's why I normally don't like depend on these community build images:)

shuoyang2016 (approved):

PR #630, opened 2021-10-23
b/186465840: Support unescape plus in grpc transcoding

qiwzhang:

To support [HTML 2.0](https://tools.ietf.org/html/rfc1866#section-8.2.1) which escape space as + in the query parameter.

This is behavior changes, in order not to break existing customers, 
add a flag `--transcoding_query_parameters_unescape_plus` to enable the new behavior.


TheSpy:
Just for information, default behavior breaks base64 encoded strings in URL as base64 allows + characters

qiwzhang:
Thanks for the info.  I think query parameter should use Base64URL_encode,  not Base64_encode.  Base64URL should not allow +.

PR #639, opened 2021-12-03
b/208891954: disallow colon in url wildcard path segment for route match

TAOXUY:
**Context:** according [Google http template spec](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L246-L250), the colon or the custom verb shouldn't be part of path segment. However, our current route match does, so there is a discrepancy between our json grpc transcoding and route match.

Though Google http spec doesn't allow colon in path segment, IETF [URL spec](https://screenshot.googleplex.com/AaYEhyGKoPC3mfm) does. We may have users uses such urls.

**Note**: IMO, we should fully follow Google's spec and it is also consistent with our transcoder behavior.  This PR will disallow column as the wildcard match so that the trailing custom verb won't be matched into wildcard segment. By default, it is not enabled.

Signed-off-by: Xuyang Tao <taoxuy@google.com>


docker/generic/start_proxy.py
```diff
@@ -869,6 +869,15 @@ def make_argparser():
         in grpc-json transcoding. This is to support HTML 2.0<https://tools.ietf.org/html/rfc1866#section-8.2.1>.
         Set this flag to true to disable this feature.
         ''')
+    parser.add_argument(
+        '--exclude_colon_in_url_wildcard_path_segment', action='store_true',
```
qiwzhang, line 873: maybe just call "--exclude_colon_in_wildcard_path_match"

qiwzhang (commented):
Two small nits

docker/generic/start_proxy.py
```diff
@@ -869,6 +869,15 @@ def make_argparser():
...
+        help='''
```
nareddyt, line 874: Should we default this to true and let users set false manually? I know it is a potential breaking change, but it can lead mismatched route (and potential auth bypass) if we keep it disabled.

src/go/configgenerator/route_generator.go
```diff
@@ -494,7 +494,7 @@ func makeHttpExactPathRouteMatcher(path string) *routepb.RouteMatch {
 	}
 }
 
-func makeHttpRouteMatchers(httpRule *httppattern.Pattern, seenUriTemplatesInRoute map[string]bool) ([]*routepb.RouteMatch, []*routepb.RouteMatch, error) {
+func makeHttpRouteMatchers(httpRule *httppattern.Pattern, seenUriTemplatesInRoute map[string]bool, ExcludeColonInUrlWildcardPathSegment bool) ([]*routepb.RouteMatch, []*routepb.RouteMatch, error) {
```
nareddyt, line 497: nit: lowercase e

nareddyt: Please fix across all functions

tests/integration_test/transcoding_bindings_test/transcoding_bindings_test.go
```diff
@@ -147,72 +149,115 @@ func TestTranscodingBindings(t *testing.T) {
...
-		wantResp           string
-		wantErr            string
-		wantGRPCWebTrailer client.GRPCWebTrailer
+		desc                                  string
+		clientProtocol                        string
+		httpMethod                            string
+		method                                string
+		noBackend                             bool
+		token                                 string
+		ExcludeColumnInUrlPathWildcardSegment bool
```
nareddyt, line 176: ditto

docker/generic/start_proxy.py
nareddyt: @qiwzhang

docker/generic/start_proxy.py
nareddyt: How about we make the flag name more high level so it is in terms of the service config or end user behavior? Users don't know the details on path matching, colon, wildcards, etc.

Also it should be prefixed with transcoding, as users will only see behavior change in the transcoder with this flag.

`--transcoding_disallow_unregistered_custom_verbs`? Similar to https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/pull/66

tests/endpoints/bookstore_grpc/proto/v1/bookstore.proto
```diff
@@ -86,6 +86,40 @@ service Bookstore {
...
+      post: "/v1/shelves/{shelf}/double/**"
+      body: "book"
+    };
+  }
+
+  // Creates a new book with http template containing custom verb.
+  rpc CreateBookWithCustomVerb(CreateBookRequest) returns (Book) {
+    // Client example - create a new book in the first shelf:
+    //   curl -d '{"author":"foo","title":"bar"}'
+    //   http://DOMAIN_NAME/v1/shelves/1/books
```
shuoyang2016, line 116: Please update the comment

TAOXUY: Done.

docker/generic/start_proxy.py
```diff
@@ -869,6 +869,15 @@ def make_argparser():
...
+        Whether exclude colon in the url wildcard path segment for route match. According
+        to Google http url template spec[1], the literal colon cannot be used in
+        wildcard match.
+        
+        [1]https://github.com/googleapis/googleapis/blob/165280d3deea4d225a079eb5c34717b214a5b732/google/api/http.proto#L226-L252
```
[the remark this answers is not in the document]
TAOXUY: If we use master, we cannot point to certain lines which make reader inconvenience to find certain lines. This file should be stable so there shouldn't be much difference between master and certain SHA.

docker/generic/start_proxy.py
TAOXUY: + 1 on this idea. Done.

src/go/configgenerator/route_generator.go
TAOXUY: Done.

tests/integration_test/transcoding_bindings_test/transcoding_bindings_test.go
TAOXUY: Done.

docker/generic/start_proxy.py
nareddyt: @TAOXUY

TAOXUY: Prefer `--allow_colon_in_wildcard_path_match` as it actually unrelated to the gRPC transcoding but for the route match only.

docker/generic/start_proxy.py
```diff
@@ -869,6 +869,15 @@ def make_argparser():
         in grpc-json transcoding. This is to support HTML 2.0<https://tools.ietf.org/html/rfc1866#section-8.2.1>.
         Set this flag to true to disable this feature.
         ''')
+    parser.add_argument(
+        '--allow_colon_in_wildcard_path_segment', action='store_true',
```
qiwzhang, line 873: What is the default value?  If false, you are changing the default behaviors.  It may break some users for ESPv2.

TAOXUY: @nareddyt and I both agree it is better to enable this change by default, as:
- we should follow Google spec and the misroute is a bug we wanna fix for all users
- it also aligned with our transcoder behavior

qiwzhang: For  a /foo/{x=*},    /foo/bar1:bar2 used to work with x=bar1:bar2  after the change, it will be rejected.   This is a breaking change.  Prefer not to have  breaking change.

docker/generic/start_proxy.py
nareddyt: Hmm, I agree implementation wise it is a change in path matching.

But these routes are only used when gRPC transcoding is configured. gRPC pass through does not use custom verbs. OpenAPI has no concept of custom verbs. This is only used by google.api.http annotations, which is for transcoding.

So I would prefer to keep this flag as `--transcoding_...`. It will make it very clear to customers. Customers who don't use transcoding know that they don't need to worry about the flag

TAOXUY: We don't know if/how http backend users use the `:` in their path segment. For those users, they need to flip this flag for backward compatibility and `--transcoding_...` doesn't make sense to them.

For transcoding users, they should understand this flag as they defined the custom verb in their httpRule.

TAOXUY:
Changed the fix disabled by default. 

PTAL.

shuoyang2016 (approved):

qiwzhang (approved):

nareddyt (approved):

PR #640, opened 2021-12-08
Support http.rules in the service config for gRPC transcoding

qiwzhang:

To support specifying custom http rules in the service config.  Normally you specify "google.api.http" annotation in the service proto files to map HTTP path to gRPC /service/method, and extract some field info from the path segments.

But for some shared gRPC service, such as HealthService or LongRunningOperation,  you may want to add your own http.rules.  With this feature, you can add them in the service config.  It will overwrite the ones from the proto files.

For example, in your proto, you may have
```
  // Returns a specific bookstore shelf.
  rpc GetShelf(GetShelfRequest) returns (Shelf) {
    option (google.api.http) = {
      get: "/v1/shelves/{shelf}"
    };
  }
```

If you want to use http `get "/vlalpha/custom_shelves/{shelf}"` to call this method, you can add following http.rules in the service config
```
http
   rules
   - selector: "PackageName.BookStore.GetShelf"
     get "/vlalpha/custom_shelves/{shelf}"
```

The whole prototype is in the cl/415315843.  This PR is ESPV2 part of changes


src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -165,6 +169,60 @@ func MakeFilterGenerators(serviceInfo *ci.ServiceInfo) ([]*FilterGenerator, erro
...
+					proto.SetExtension(method.GetOptions(), ahttp.E_Http, rule)
+				}
+			}
+		}
+	}
+
+	newData, err := proto.Marshal(fds)
+	if err != nil {
+		glog.Error("failed to marshal protodescriptor, error: %v", err)
+		return descriptorBytes
```
shuoyang2016, line 221: Return the error. This looks a critical failure.

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -165,6 +169,60 @@ func MakeFilterGenerators(serviceInfo *ci.ServiceInfo) ([]*FilterGenerator, erro
 	return filterGenerators, nil
 }
 
+func updateProtoDescriptor(serviceInfo *ci.ServiceInfo, descriptorBytes []byte) []byte {
+	// To support specifying custom http rules in service config.
+	// api-compiler (or CLI: gcloud endpoints service deploy) has following behaviours:
+	// * If a "google.api.http" annotation is specified in a method in the proto,
+	//   service config yaml did not specify one, api-compiler will copy it out
+	//   to the normalized service config.
+	// * If a htpp.rule is specified in the service config, it will override
```
shuoyang2016, line 178: nit: typo htpp.rule

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -165,6 +169,60 @@ func MakeFilterGenerators(serviceInfo *ci.ServiceInfo) ([]*FilterGenerator, erro
...
+		ruleMap[rule.GetSelector()] = rule
+	}
+	apiMap := make(map[string]bool)
+	for _, apiName := range serviceInfo.ApiNames {
+		apiMap[apiName] = true
+	}
+
+	fds := &descpb.FileDescriptorSet{}
+	if err := proto.Unmarshal(descriptorBytes, fds); err != nil {
+		glog.Error("failed to unmarshal protodescriptor, error: %v", err)
```
shuoyang2016, line 194: Ditto: return error. This seems critical.

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -165,6 +169,60 @@ func MakeFilterGenerators(serviceInfo *ci.ServiceInfo) ([]*FilterGenerator, erro
 	return filterGenerators, nil
 }
 
+func updateProtoDescriptor(serviceInfo *ci.ServiceInfo, descriptorBytes []byte) []byte {
```
shuoyang2016, line 172: Please add unit test.

shuoyang2016 (commented):
Is this only used by ESPv2/Cloud ESF? Adding it in API compiler may benefit other usages too? Is it more complicated to touch api compiler for this use case?

qiwzhang:
api-compiler is used for many google teams.  But Only envoy is using proto descriptor for grpc_json transcoder.   So only CloudESF/ESPv2 is using this feature. Not benefits of add this logic in api-compler

shuoyang2016:
> api-compiler is used for many google teams. But Only envoy is using proto descriptor for grpc_json transcoder. So only CloudESF/ESPv2 is using this feature. Not benefits of add this logic in api-compler

Thanks.

I just realized that this won't work for ASM because the original proto descriptor is read from file (init container) http://shortn/_T3sIJAVFat. Perhaps we should consider adding the http rule in the filter config.

qiwzhang:
So ASM is using filePath?  The only place generating grpc_json_transcoder is here.   It requires service_config to have descriptor. Otherwise  the filter is not created.

Did you hack the ASM code, to use this code to generate grpc_json filter, and later modify it to use "file_path" to load descriptor?

src/go/configgenerator/filterconfig/filter_generator.go
qiwzhang: Done

src/go/configgenerator/filterconfig/filter_generator.go
qiwzhang: Done

src/go/configgenerator/filterconfig/filter_generator.go
qiwzhang: Done

qiwzhang:
PTAL

src/go/configgenerator/filterconfig/update_proto_descriptor_test.go
```diff
@@ -0,0 +1,301 @@
...
+file: {
+  name: "proto_file_path"
+  package: "package.name"
+  service: {
+    name: "Service"
+    method: {
+      name: "Method"
+      options: {
+        [google.api.http]: {
+	  post: "/v1/abc/{name=*}"
```
nareddyt, line 192: Instead of overwrite, should we place the copied http rule into `additional_bindings`?

nareddyt (approved):
Test cases are very good and exhaustive, thanks for adding them. One small comment

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -194,10 +273,18 @@ func makeTranscoderFilter(serviceInfo *ci.ServiceInfo) *hcmpb.HttpFilter {
...
 				},
 			}
-			if serviceInfo.Options.TranscodingFilePath != "" {
+			if serviceInfo.Options.TranscodingFilePathHost != "" {
 				transcodeConfig.DescriptorSet = &transcoderpb.GrpcJsonTranscoder_ProtoDescriptor{
-					ProtoDescriptor: serviceInfo.Options.TranscodingFilePath,
+					ProtoDescriptor: serviceInfo.Options.TranscodingFilePathHost,
+				}
+			}
+			if serviceInfo.Options.TranscodingFilePathOutput != "" {
```
shuoyang2016, line 281: This is only needed by asm gen? Can we do it there in google3? I feel this side effect is too subtle for makeTranscoderFilter which is a bit hard to find.

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -165,7 +175,71 @@ func MakeFilterGenerators(serviceInfo *ci.ServiceInfo) ([]*FilterGenerator, erro
 	return filterGenerators, nil
 }
 
-func makeTranscoderFilter(serviceInfo *ci.ServiceInfo) *hcmpb.HttpFilter {
+func updateProtoDescriptor(service *confpb.Service, apiNames []string, descriptorBytes []byte) ([]byte, error) {
```
shuoyang2016, line 178: Can we make this a public method and be called by asmgen?

src/go/configgenerator/filterconfig/filter_generator.go
qiwzhang: I will take a look.  The idea is after TranscoderFilter is generated,  ASM code needs to Unmarshal it and modify it according to its special need.   I will see how much is the code changes,   if it is too big,  will do it in the separate PR.

shuoyang2016 (approved):

PR #646, opened 2021-12-16
Remove codes to use filePath in GrpcJsonTranscoder

qiwzhang:


nareddyt (approved):
What about the flag definitions? Are they still set in ESPv2 codebase even though they are used for ASM?

TAOXUY (approved):

PR #647, opened 2021-12-22
Update janitor to delete orphan LBs

nareddyt:
Find and delete all orphan target pools in `us-central1`

### DIFF
--- a/prow/janitor.sh
+++ b/prow/janitor.sh
@@ -1,10 +1,6 @@
 #!/bin/bash
 
-# Janitor script triggered by prow to cleanup ESPv2 on serverless resources.
-# This includes:
-# - Cloud run services
-# - Cloud functions
-# - Endpoints services
+# Janitor script triggered by prow to cleanup orphan resources used for ESPv2 testing.
 
 # Fail on any error.
 set -eo pipefail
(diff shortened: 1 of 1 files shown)


prow/janitor.sh
```diff
@@ -112,3 +108,33 @@ for service in $ENDPOINTS_SERVICES ; do
 
 done
 echo "Done cleaning up Endpoints Services"
+
+# Load balancer target pools
+# Source: https://gist.github.com/prasvats/b2a4e33ad12b40191dd9d7e222d1abde/
```
shuoyang2016, line 113: Just want to make sure the license etc is fine?

shuoyang2016 (approved):
Great. Thanks, Teju!

Anyway to cleanup backend service, forwarding rules etc?

prow/janitor.sh
nareddyt: Unfortunately it is not licensed, maybe we cannot use it then.

nareddyt:
> Great. Thanks, Teju!
> 
> Anyway to cleanup backend service, forwarding rules etc?

I could not find an easy way to do this, requires writing a more complex script. We'll see how often we run into the quota issue

PR #654, opened 2022-01-21
b/215322819: Support url templates with variables without wildcard

qiwzhang:

To support following url template for grpc transcoding

```
/v1/{name=operations}
```

Supposedly,  a variable should have at least one wildcard * or double wildcard ** segment.  But above url template has a variable doesn't have any wildcard.  It is legal from the [syntax](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L224-L232), hence add this pr to support it.

The details:
* url_template_parser is working fine,  added two cases to verifiy it in the url_template_parser_test.go
* The problem is in "ExactMatchString" function, when it see a variable it automaticadlly append `/{var=path}` which is wrong.  If there is not wildcard in the variable, just append `/path`


TAOXUY:
Wayne, do you check if the transcoder works for this syntax?

TAOXUY (approved):

PR #659, opened 2022-01-31
b/208258342: add transcoding_match_unregistered_custom_verb

TAOXUY:
Adopt [`match_unregistered_custom_verb`](https://github.com/envoyproxy/envoy/pull/19203) from upstream envoy.


src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -271,6 +271,7 @@ func makeTranscoderFilter(serviceInfo *ci.ServiceInfo) (*hcmpb.HttpFilter, error
 					AlwaysPrintEnumsAsInts:     serviceInfo.Options.TranscodingAlwaysPrintEnumsAsInts,
 					PreserveProtoFieldNames:    serviceInfo.Options.TranscodingPreserveProtoFieldNames,
 				},
+				MatchUnregisteredCustomVerb: serviceInfo.Options.TranscodingMatchUnregisteredCustomVerb,
```
qiwzhang, line 274: The correct behaviors should be "on",  is that correct?
Should we just turn it on by default?  
is this a breaking change?

docker/generic/start_proxy.py
```diff
@@ -877,6 +877,20 @@ def make_argparser():
         
         [1]https://github.com/googleapis/googleapis/blob/165280d3deea4d225a079eb5c34717b214a5b732/google/api/http.proto#L226-L252
         ''')
+    parser.add_argument(
+        '--transcoding_match_unregistered_custom_verb', action='store_true',
```
qiwzhang, line 881: Even I don't know how to use this flag,  I am not sure if our users know.

src/go/configgenerator/filterconfig/filter_generator.go
nareddyt: +1, how likely is it to break users? Silently turning it on seems fine to me

TAOXUY: Yes, it could a breaking change so I prefer to disable it. Internally, we can enable it by default.

As it is the issue effecting internal users, we just don't  expose to ESPv2 customers for now(most likely, ESPv2 users don't use custom verbs).

nareddyt: Ok, keep it disabled externally and enable internally SG

nareddyt (approved):

qiwzhang (approved):

PR #662, opened 2022-02-06
Expose the `--config-yaml` envoy flag 

Maor-RZ:
Allows the flag to be set by the user.
Enables passing a config YAML string to Envoy, for better controlling the bootstrap configuration as described [here](https://www.envoyproxy.io/docs/envoy/latest/operations/cli).

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -887,6 +887,17 @@ environment variable or by passing "-k" flag to this script.
         envoy. Only change if running multiple ESPv2 instances on the same host.
         '''
     )
+    parser.add_argument(
+        '--config_yaml',
+        default=None,
+        help='''
+            Specify an additional bootstrap configuration for Envoy in the form of a YAML string.
+            Values will override and merge with any bootstrap configuration loaded with '--config-path'.
+            This will not override any value which has been set dynamically.
+
+            For more details see:
+            https://www.envoyproxy.io/docs/envoy/latest/operations/cli
+            ''')
 
     # Start Deprecated Flags Section
 
@@ -1417,6 +1428,9 @@ def gen_envoy_args(args):
         # Enable debug logging, but not for everything... too noisy otherwise.
         cmd.append("-l debug")
         cmd.append("--component-log-level upstream:info,main:info")
+
+    if args.config_yaml:
+        cmd.append("--config-yaml {}".format(args.config_yaml))
 
     return cmd
 


TAOXUY:
Thanks for the contribution. Please add unit test in tests/start_proxy/start_proxy_test.py.

docker/generic/start_proxy.py
```diff
@@ -887,6 +887,17 @@ def make_argparser():
         envoy. Only change if running multiple ESPv2 instances on the same host.
         '''
     )
+    parser.add_argument(
+        '--config_yaml',
```
qiwzhang, line 891: maybe call it --envoy-extra-config-yaml?   

It is the extra boostrap config in additional to ESPv2 clusters/listeners.  The cluster names and listener name should not step over the ones generated by ESPv2, otherwise it will break ESPv2. 

I am really hesitate to add this flag, if it is not used carefully, we could shot yourself in the foot.   Could you provide some config examples that you like to use this flag for?

Maor-RZ: Hi, 
The main goal is to enable all Envoy's abilities - Gzip configuration in my case - independently (and, in addition, to allow dynamic management to every aspect of the bootstrap config).

We can also name it with a more "alarming" name such as `--modify-bootstrap-conig-yaml` and highlight the "overriding" and "merging" aspects of it in the help text - although I am pretty sure that users will mostly use it to manage static resources that managing them isn't being supported yet through ESP (like in my case)

qiwzhang: I prefer to call it `--envoy-extra-config-yaml` and ok to add this flag.

tests/start_proxy/start_proxy_test.py
```diff
@@ -17,10 +17,10 @@
 
 import os, inspect
 
-currentdir = os.path.dirname(
-    os.path.abspath(inspect.getfile(inspect.currentframe())))
-sys.path.insert(0, currentdir + "/../../docker/generic")
-from start_proxy import gen_bootstrap_conf, make_argparser, gen_proxy_config, gen_envoy_args
+# currentdir = os.path.dirname(
+#     os.path.abspath(inspect.getfile(inspect.currentframe())))
+# sys.path.insert(0, currentdir + "/../../docker/generic")
```
qiwzhang, line 22: if importing "docker.generic.start_proxy" works, then remove above 3 lines then.

Maor-RZ: Actually, I mistakenly left it like this due to problems with my environment, I will just change it back - I don't want to perform any unnecessary changes.

qiwzhang (approved):

PR #663, opened 2022-02-08
Increase health check time

nareddyt:
`TestRetryCallServiceManagement` fails because health checks exit too early


TAOXUY (approved):

qiwzhang:
ESPv2-presubmit  still fail!.

qiwzhang (approved):

PR #667, opened 2022-02-24
b/221072669: Remove envoy runtime flag preserve_downstream_scheme

qiwzhang:

Envoy coming release 1.22 will drop this runtime flag  `preserve_downstream_scheme`.  Currently, we are using this flag to work around a Cloud-Run bug: it sets scheme to "http" if the flag `--use_http2`  is used in Cloud-Run deployment.  It will cause ESPv2 to set scheme as "http" too even the backend connection is using TLS.  This will break the backend using IAP which requires scheme to be "https".

The changes: 
1)  remove the runtime flag, 
2) use SchemeHeaderOverride in the http_connection_manager to set scheme as "https" under following conditions:
* Deployed in server_less mode, 
* backend is grpc and uses TLS

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -54,6 +54,7 @@ DEFAULT_ROLLOUT_STRATEGY = "fixed"
 GOOGLE_CREDS_KEY = "GOOGLE_APPLICATION_CREDENTIALS"
 
 # Flag defaults when running on serverless.
+# SERVERLESS_PLATFORM has to match the one in src/go/util/util.go
 SERVERLESS_PLATFORM = "Cloud Run(ESPv2)"
 SERVERLESS_XFF_NUM_TRUSTED_HOPS = 0
 
@@ -1356,7 +1357,6 @@ def gen_proxy_config(args):
     if args.on_serverless:
         proxy_conf.extend([
             "--compute_platform_override", SERVERLESS_PLATFORM])
-        proxy_conf.append("--server_less")
 
     if args.backend_dns_lookup_family:
         proxy_conf.extend(
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -1356,6 +1356,7 @@ def gen_proxy_config(args):
     if args.on_serverless:
         proxy_conf.extend([
             "--compute_platform_override", SERVERLESS_PLATFORM])
+        proxy_conf.append("--server_less")
 
     if args.backend_dns_lookup_family:
         proxy_conf.extend(
--- a/src/go/configgenerator/listener_generator.go
+++ b/src/go/configgenerator/listener_generator.go
@@ -107,6 +107,7 @@ func MakeListener(serviceInfo *sc.ServiceInfo, filterGenerators []*filterconfig.
 	if err != nil {
 		return nil, fmt.Errorf("makeHttpConnectionManager got err: %s", err)
 	}
+	httpConMgr.SchemeHeaderTransformation = makeSchemeHeaderOverride(serviceInfo)
 
 	jsonStr, _ := util.ProtoToJson(httpConMgr)
 	glog.Infof("adding Http Connection Manager config: %v", jsonStr)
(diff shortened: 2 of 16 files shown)


docker/generic/start_proxy.py
```diff
@@ -1356,6 +1356,7 @@ def gen_proxy_config(args):
     if args.on_serverless:
         proxy_conf.extend([
             "--compute_platform_override", SERVERLESS_PLATFORM])
+        proxy_conf.append("--server_less")
```
shuoyang2016, line 1359: Just use flag compute_platform_override?

src/go/configgenerator/listener_generator.go
```diff
@@ -165,6 +166,33 @@ func MakeListener(serviceInfo *sc.ServiceInfo, filterGenerators []*filterconfig.
 	return listener, nil
 }
 
+// To fix b/221072669: a hack to work around b/221308324 where
+// Cloud Run always set :scheme header to http when using http2 protocol for grpc.
+// Override scheme header to https when following conditions meet:
+// * Deployed in serverless platform.
+// * Backend uses grpc
+// * All remote backends are using TLS
+func makeSchemeHeaderOverride(serviceInfo *sc.ServiceInfo) *corepb.SchemeHeaderTransformation {
```
shuoyang2016, line 175: Add a unit test?

qiwzhang: Done

docker/generic/start_proxy.py

src/go/configgenerator/listener_generator.go
```diff
@@ -165,6 +166,33 @@ func MakeListener(serviceInfo *sc.ServiceInfo, filterGenerators []*filterconfig.
...
+	if serviceInfo.Options.ComputePlatformOverride != util.ServerlessPlatform || !serviceInfo.GrpcSupportRequired {
+		return nil
+	}
+	allTLS := true
+	for _, v := range serviceInfo.RemoteBackendClusters {
+		if !v.UseTLS {
+			allTLS = false
```
shuoyang2016, line 182: nit: break

qiwzhang:
PTAL

src/go/configgenerator/listener_generator.go
```diff
@@ -165,6 +166,33 @@ func MakeListener(serviceInfo *sc.ServiceInfo, filterGenerators []*filterconfig.
...
+// Override scheme header to https when following conditions meet:
+// * Deployed in serverless platform.
+// * Backend uses grpc
+// * All remote backends are using TLS
+func makeSchemeHeaderOverride(serviceInfo *sc.ServiceInfo) *corepb.SchemeHeaderTransformation {
+	if serviceInfo.Options.ComputePlatformOverride != util.ServerlessPlatform || !serviceInfo.GrpcSupportRequired {
+		return nil
+	}
+	allTLS := true
+	for _, v := range serviceInfo.RemoteBackendClusters {
```
shuoyang2016, line 180: What will happen if some backends are TLS but some aren't?

qiwzhang: We could not use scheme_override.  For now, envoy scheme_override applies to all backends.  if some backend is not ssl,  the override will be wrong. 

I think this is ok.  All backends for serverless are using TLS.

shuoyang2016: > All backends for serverless are using TLS.

If this is the case, I think we should just return error here so that it simply fails at startup?

qiwzhang: That will be too brutal. In the case of not all backends use TLS,  my pr just don't do "override_scheme=https" ,  your proposal is, not to support it at all.  

I am afraid that there are some users using ESPv2 in such way.  We don't want to break them.  They may not need correct "scheme" header.   Only IAP backends require correct "scheme" headers.

TAOXUY (approved):

src/go/configgenerator/listener_generator.go
shuoyang2016: I see. Where I come from is that it looks to me there are some combinations that we just cannot support. If there are some backends require https like IAP and some backends require http? Anyway we'll have to break one of them?

TAOXUY: Just discussed with Wayne about this. Since Envoy doesn't allow per backend scheme override, it would be a breaking change for certain use cases. Now, we only override scheme when makeSchemeHeaderOverride **all** backends are https which breaks the case Shuo mentioned where some backends are using IAP and some backends are HTTP.
. For fixing that, probably we can change to override scheme when **any** backend is https. Most Cloud Run backend should be https and the http backend should be fine with this wrong scheme header.

shuoyang2016: >Most Cloud Run backend should be https and the http backend should be fine with this wrong scheme header
Thanks. This sounds good.

shuoyang2016 (approved):

src/go/configgenerator/listener_generator.go
qiwzhang: Code is updated to:  if ANY of backends is using SSL,  we will use scheme_override.  This is to supported multiple backends with mixed SSL and non-SSL.

nareddyt (approved):
I guess this is too much of an edge case to have an integration test for, right?

Can we somehow use HTTP echo server (with dynamic routing) to check it receives the correct `:scheme` header? 

https://github.com/GoogleCloudPlatform/esp-v2/blob/85785b79c3b719145e0b1779147ae4cca3c2aede/tests/endpoints/echo/server/app.go#L68

Maybe not worth it as long as Cloud Run e2e tests catch it

PR #715, opened 2022-07-12
Hard code health check to skip service control.

dafudeng:


src/go/configinfo/service_info.go
```diff
@@ -719,6 +719,17 @@ func (s *ServiceInfo) processLocalBackendOperations() error {
 }
 
 func (s *ServiceInfo) processUsageRule() error {
+
+	// TODO(dafudeng) b/238652680
```
nareddyt, line 723: Remove TODO?

nareddyt (commented):
Ideally we need an integration test. How do you know that setting `SkipServiceControl` will actually result in Envoy allowing the request through?

The bookstore gRPC server already registers the gRPC health check service: https://github.com/GoogleCloudPlatform/esp-v2/blob/b2d2351c48b59cf1f3e090e9dbdbeb5c271fc1ab/tests/endpoints/bookstore_grpc/server/server.go#L170

So try adding these health check API to the fake service config (https://github.com/GoogleCloudPlatform/esp-v2/blob/b2d2351c48b59cf1f3e090e9dbdbeb5c271fc1ab/tests/env/testdata/fake_grpc_bookstore_service_config.go#L29) and check if you can successfully call the health check service without providing an API key (so no chemist check occurs).

src/go/configinfo/service_info.go
dafudeng: Yes, planned to add integration test and will do it. 
I put TODO here for two reasons: 1. it does not look elegant 2. if user wants to define their own health check mechanism, then we may need to implement a more generic way to allow the health check to bypass service control, i.e. use flag --healthcheck=grpc.health.v1.Health. If a more generic implementation is available, this hardcoded logic will be replaced.

Let me know if the above make sense and if they are real concerns, especially the second one.

qiwzhang: I have two concerns
1) chnages here also impact ESPv2.  It may break existing customers,  should we do it here?
2) Users may want to "skip_service_control" their custom health service, do we want to support it.

nareddyt: > 1. chnages here also impact ESPv2. It may break existing customers, should we do it here?

This is not rejecting the config for ESPv2 customers, it is just adding in the skip_service_control config. It shouldn't break any customers, unless some customer defined gRPC health check protocol and doesn't want to skip service control for it. I doubt that is the case.

> 2. if user wants to define their own health check mechanism

I don't think 2 is a concern. If customer is using gRPC backend, they should be using gRPC health check protocol.

For ESPv2, we still allow customers to define their own mechanism. We don't reject any configs with skip_service_control. So there is no issue.

We can enforce that Cloud ESF users follow gRPC health check protocol.

qiwzhang: SG

src/go/configinfo/service_info.go
```diff
@@ -719,6 +719,17 @@ func (s *ServiceInfo) processLocalBackendOperations() error {
...
+	// hard code to allow health check, if defined, to bypass service control,
+	// need to refactor this.
+	if method, err := s.getMethod("grpc.health.v1.Health.Check"); err == nil {
+		method.SkipServiceControl = true
+	}
+	if method, err := s.getMethod("grpc.health.v1.Health.Watch"); err == nil {
```
qiwzhang, line 729: Coding wise, maybe to define a HardCodedSkipServiceControlMethodList  in "util", it is more flexible.

src/go/configinfo/service_info.go
dafudeng: TODO removed.

src/go/configinfo/service_info.go
dafudeng: Agree. Changed.

tests/integration_test/service_control_skip_test/service_control_skip_test.go
```diff
@@ -90,3 +92,68 @@ func TestServiceControlSkipUsage(t *testing.T) {
...
+		{
+			desc:                  "health check should bypass service control by default",
+			enforceServiceControl: false,
+			wantResp:              "SERVING",
+			wantScRequestCount:    0,
+		},
+		{
+			desc:                  "service control on health check can be enforced setting SkipServiceControl to be false",
+			enforceServiceControl: true,
+			wantResp:              "UNAUTHENTICATED",
```
qiwzhang, line 113: Could you add comment on why the result is "UNAUTHENTICATED"?

src/go/configinfo/service_info.go
```diff
@@ -719,6 +719,13 @@ func (s *ServiceInfo) processLocalBackendOperations() error {
 }
 
 func (s *ServiceInfo) processUsageRule() error {
+	// hard code to allow health check, if defined, to bypass service control,
+	for _, methodName := range util.HardCodedSkipServiceControlMethods {
```
nareddyt, line 723: please add comment: If user desires to enable service control for one of these allowlisted methods, they can specify `skip_service_control: false` in their service config and ESPv2 honors it.

src/go/configinfo/service_info_test.go
```diff
@@ -3776,6 +3776,117 @@ func TestProcessAccessToken(t *testing.T) {
...
+							HttpMethod:  util.POST,
+							UriTemplate: parseUriTemplate("/grpc.health.v1.Health/Watch"),
+						},
+					},
+					SkipServiceControl: true,
+				},
+			},
+		},
+		{
+			desc: "Attach usage rules to methods",
```
nareddyt, line 3828: User overrides hardcoded skip service control with usage rule

src/go/util/util.go
```diff
@@ -155,3 +155,8 @@ func MaybeTruncateSpanName(spanName string) string {
 	newSpanName := spanName[:SpanNameMaxByteNum-3] + "..."
 	return newSpanName
 }
+
+var HardCodedSkipServiceControlMethods = []string{
+	"grpc.health.v1.Health.Check",
+	"grpc.health.v1.Health.Watch",
+}
```
nareddyt, line 162: nit: add newline at end of file

src/go/util/util.go
```diff
@@ -155,3 +155,8 @@ func MaybeTruncateSpanName(spanName string) string {
 	newSpanName := spanName[:SpanNameMaxByteNum-3] + "..."
 	return newSpanName
 }
+
+var HardCodedSkipServiceControlMethods = []string{
```
nareddyt, line 159: nit: need to add comment for exported types

// HardCodedSkipServiceControlMethods is a list of methods that ...

https://go.dev/doc/comment

tests/endpoints/bookstore_grpc/server/server.go
```diff
@@ -34,7 +34,6 @@ import (
 
 	bspbv1 "github.com/GoogleCloudPlatform/esp-v2/tests/endpoints/bookstore_grpc/proto/v1"
 	bspbv2 "github.com/GoogleCloudPlatform/esp-v2/tests/endpoints/bookstore_grpc/proto/v2"
-	healthgrpc "google.golang.org/grpc/health/grpc_health_v1"
```
nareddyt, line 37: Please revert. This is needed for copybara into google3. gRPC imports need to be different from proto imports

nareddyt (approved):

qiwzhang (approved):

PR #716, opened 2022-07-14
b/235088060: Add option to enable verbose tracing (not enabled, not exposed)

nareddyt:
Unfortunately it is not looking very helpful from ESPv2 e2e tests. But I would like to expose this setting so we can test with streaming calls in Cloud ESF.

It is **disabled by default** and not exposed to ESPv2 users

Example:

image

Signed-off-by: nareddyt <tejunareddy@gmail.com>

### DIFF
--- a/src/go/commonflags/flags.go
+++ b/src/go/commonflags/flags.go
@@ -27,24 +27,24 @@ var (
 	// Any flags in this file are used by both the ADS Bootstrapper (startup) and Config Generation via the static bootstrapper or config manager.
 	defaults = options.DefaultCommonOptions()
 
-	AdminAddress               = flag.String("admin_address", defaults.AdminAddress, "Address that envoy should serve the admin page on. Supports both ipv4 and ipv6 addresses.")
-	AdsNamedPipe               = flag.String("ads_named_pipe", defaults.AdsNamedPipe, "Unix domain socket to use internally for xDs between config manager and envoy.")
-	DisableTracing             = flag.Bool("disable_tracing", defaults.DisableTracing, `Disable stackdriver tracing`)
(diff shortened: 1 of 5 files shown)


src/go/commonflags/flags.go
```diff
@@ -44,6 +44,7 @@ var (
 	TracingMaxNumAnnotations   = flag.Int64("tracing_max_num_annotations", defaults.TracingMaxNumAnnotations, "Sets the maximum number of annotations that each span can contain. Defaults to the maximum allowed by Stackdriver. In practice, the number of annotations published will be much less.")
 	TracingMaxNumMessageEvents = flag.Int64("tracing_max_num_message_events", defaults.TracingMaxNumMessageEvents, "Sets the maximum number of message events that each span can contain. Defaults to the maximum allowed by Stackdriver. In practice, the number of message events published will be much less.")
 	TracingMaxNumLinks         = flag.Int64("tracing_max_num_links", defaults.TracingMaxNumLinks, "Sets the maximum number of links that each span can contain. Defaults to the maximum allowed by Stackdriver. In practice, the number of links published will be much less.")
+	TracingEnableVerboseTiming = flag.Bool("tracing_enable_verbose_timing", defaults.TracingEnableVerboseTiming, "If enabled, spans are annotated with timing events on when the request/response started/ended")
```
qiwzhang, line 47: I feel, adding "_timing" suffix is confusing.   maybe remove it?  or "tracing_add_more_event_timing"

nareddyt: How about now? `--tracing_enable_verbose_annotations`, the flag adds more verbose annotations to trace span

nareddyt:
Hi @qiwzhang, PTAL. I updated the description to disable by default

shuoyang2016 (approved):

src/go/commonflags/flags.go
qiwzhang: SG

qiwzhang (approved):

PR #725, opened 2022-08-16
Enforce the default http rule

dafudeng:
Add the default http rule of a rpc endpoint to the additional_bindings field of any new http rules of that rpc endpoint. Otherwise, once a new http rule is set, the default http rule will no longer work, which is not expected.


src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -252,6 +252,15 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
 					if method.GetOptions() == nil {
 						method.Options = &descpb.MethodOptions{}
 					}
+
+					// If an http rule is specified for a rpc endpoint then the rpc's default http path will be
+					// disabled according to the logic in the envoy's json transcoder filter. To still enable
+					// the default http path, which is the designed behavior, the default http path needs to be
+					// added to the http rule's additional bindings.
+					defaultPath := fmt.Sprintf("/%s/%s", apiName, method.GetName())
+					defaultRule := &ahpb.HttpRule{Pattern: &ahpb.HttpRule_Put{defaultPath}, Body: "*"}
```
TAOXUY, line 261: POST

src/go/configgenerator/filterconfig/update_proto_descriptor_test.go
```diff
@@ -250,8 +262,8 @@ file: {
       name: "Method"
       options: {
         [google.api.http]: {
-	  post: "/v1/abc/{name=*}"
-	}
+	        post: "/v1/abc/{name=*}"
+	      }
```
TAOXUY, line 266: Could you please add an integration test also?

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -252,6 +252,15 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
 					if method.GetOptions() == nil {
```
TAOXUY, line 252: `if rule, ok := ruleMap[sel]; ok {` is when http.rules in service config is defined but we want the case http.rules in proto is defined, which should check `method.GetOptions()`

When adding in additional_bindings, check if it is duplicate?

TAOXUY (commented):
Thanks Dafu. Look we are pretty close!

src/go/configgenerator/filterconfig/filter_generator.go
dafudeng: Done

src/go/configgenerator/filterconfig/filter_generator.go
dafudeng: Logic changed.

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -266,6 +275,19 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
 	return newData, nil
 }
 
+func preserveDefaultHttpBinding(httpRule *ahpb.HttpRule, defaultPath string) {
+	defaultBindingExisted := false
+	defaultBinding := &ahpb.HttpRule{Pattern: &ahpb.HttpRule_Post{defaultPath}, Body: "*"}
+	for _, addtionalBinding := range httpRule.AdditionalBindings {
```
TAOXUY, line 281: also compare with the primary binding other than additional bindings also

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -266,6 +275,19 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
...
+		if addtionalBinding == defaultBinding {
+			defaultBindingExisted = true
```
TAOXUY, line 283: This is address comparison not the deep comparison ?

Maybe add unit tests:
- with primary binding of default HTTP Pattern
- with additional bindings of default HTTP Pattern

dafudeng: Can you clarify the first case - without httprule?

src/go/configgenerator/filterconfig/filter_generator.go
dafudeng: Fixed.

src/go/configgenerator/filterconfig/filter_generator.go
TAOXUY: If no httpRule annotation, no need to add the default HTTP Pattern as the transcoder will use it  automatically.

TAOXUY (commented):
LGTM overall before adding the integration test

TAOXUY (approved):
@nareddyt  PTAL

src/go/configgenerator/filterconfig/filter_generator.go
dafudeng: The no httpRule annotation test case is added in UpdateProtoDescriptorTest.

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -266,6 +275,32 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
...
 
+func preserveDefaultHttpBinding(httpRule *ahpb.HttpRule, defaultPath string) {
+	defaultBinding := &ahpb.HttpRule{Pattern: &ahpb.HttpRule_Post{defaultPath}, Body: "*"}
+
+	// Check existence of the default binding in httpRule's additional_bindings to avoid duplication.
+	defaultBindingExisted := false
+	for _, addtionalBinding := range httpRule.AdditionalBindings {
+		if proto.Equal(addtionalBinding, defaultBinding) {
+			defaultBindingExisted = true
+			break
```
nareddyt, line 286: `return`

No need for variable `defaultBindingExisted`. Just return early if you find it, no need to run the rest of the function.

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -266,6 +275,32 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
...
+		}
+	}
+	// check if httpRule is the same as default binding, ignore the difference in fields selector and
+	// additional_bindings.
+	defaultBinding.Selector = httpRule.GetSelector()
+	defaultBinding.AdditionalBindings = httpRule.GetAdditionalBindings()
+	if proto.Equal(httpRule, defaultBinding) {
+		defaultBindingExisted = true
```
nareddyt, line 294: ditto, `return`

src/go/configgenerator/filterconfig/preserve_default_http_binding_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/GoogleCloudPlatform/esp-v2/tests/utils"
+	ahpb "google.golang.org/genproto/googleapis/api/annotations"
+	"google.golang.org/protobuf/encoding/prototext"
+)
+
+func TestPreserveDefaultHttpBinding(t *testing.T) {
+	testData := []struct {
+		desc         string
+		originalDesc string
```
nareddyt, line 29: nit: originalHttpRule, wantHttpRule

src/go/configgenerator/filterconfig/filter_generator.go
```diff
@@ -266,6 +275,32 @@ func updateProtoDescriptor(service *confpb.Service, apiNames []string, descripto
...
+	defaultBindingExisted := false
+	for _, addtionalBinding := range httpRule.AdditionalBindings {
+		if proto.Equal(addtionalBinding, defaultBinding) {
+			defaultBindingExisted = true
+			break
+		}
+	}
+	// check if httpRule is the same as default binding, ignore the difference in fields selector and
+	// additional_bindings.
+	defaultBinding.Selector = httpRule.GetSelector()
```
nareddyt, line 291: How do you guarantee that original http rule has a selector?
Similarly, how do you know the additional bindings above had no selector?

My concern is that this will break in the future. What if we start adding selector to additional binding? Then the loop above will never work

dafudeng: From what I know, if the original http rule doesn't have a selector, an error will be thrown before the execution of this part of the code.
There's no guarantee about the selectors, that's why the difference of selectors are ignored. If we start adding selector to additional binding, then the default binding will always be added.

src/go/configgenerator/filterconfig/preserve_default_http_binding_test.go
dafudeng: done

src/go/configgenerator/filterconfig/filter_generator.go
dafudeng: done

nareddyt (approved):

PR #729, opened 2022-08-29
Update prow docker image to use python3

nareddyt:


qiwzhang (approved):

TAOXUY (approved):

TAOXUY (approved):

qiwzhang:
It seems that there is one integration tests failed. 

--- FAIL: TestAccessLog (13.15s)

And it is:

    access_log_test.go:112: expect access log: "GET /noexistpath?key=test-api-key HTTP/1.1"404 - 0 73"-" "Go-http-client/1.1" "\u003cUnknown Operation Name\u003e" "test-api-key"
        , get acccess log: 

get empty acces_log,  could it be flaky?

nareddyt:
Hmm, so I thought I was just updating the tests to run with python3. But seems like I accidently updated the ESPv2 Envoy image to also run with it. Currently our Envoy crashes with the following exception:

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/apiproxy/start_proxy.py", line 1456, in output_reader
    sys.stdout.write(line.decode())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x92 in position 182: invalid start byte
```

Which option should I continue trying to fix? @qiwzhang @TAOXUY 
1) Only update prow scripts and e2e tests to run with python3, keep ESPv2 built with python2.
2) Update entire repo (including ESPv2 runtime `start_proxy.go`) to run with python2.

We can do 1 easily, but 2 is the correct long-term solution given that python2 was deprecated in 2020. But 2 is a lot of work.

qiwzhang:
Is it typo, your 2) is to run with python3.   Only python code we use for ESPv2 run-time is start_proxy.py,  should it be easy to also update it and its start_proxy_test.py for python-3?

Ideally, we want everything to use p3

nareddyt:
PTAL. Not sure why some e2e tests are still failing. Maybe flaky.

```
403
FAILED: assertEqual(got=403, want=200)
FAILED: assertEqual(got=[], want=[{'name': 'shelves/1/books/3', 'author': 'Neal Stephenson', 'title': 'REAMDE'}])
```

> It seems that there is one integration tests failed.
> 
> --- FAIL: TestAccessLog (13.15s)
> 
> And it is:
> 
> ```
> access_log_test.go:112: expect access log: "GET /noexistpath?key=test-api-key HTTP/1.1"404 - 0 73"-" "Go-http-client/1.1" "\u003cUnknown Operation Name\u003e" "test-api-key"
>     , get acccess log: 
> ```
> 
> get empty acces_log, could it be flaky?

Seems very flaky, unrelated to this change.

qiwzhang:
How about `docker/serverless/env_start_proxy.py`?  it seems it is still using "python" [here](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/docker/serverless/env_start_proxy.py#L148)

nareddyt:
> How about `docker/serverless/env_start_proxy.py`? it seems it is still using "python" [here](https://github.com/GoogleCloudPlatform/esp-v2/blob/master/docker/serverless/env_start_proxy.py#L148)

Done.

qiwzhang (approved):

shuoyang2016 (approved):

Issue #712, opened 2022-07-07
gcloud_build_image exposes access token locally


dseomn:
https://github.com/GoogleCloudPlatform/esp-v2/blob/948edb9658114b99d95452d40e43dd4a67b4f544/docker/serverless/gcloud_build_image#L98 includes the output of `gcloud auth print-access-token` in the command line arguments to `curl`, and command line arguments are generally world-readable in linux. https://curl.se/docs/manpage.html#-H mentions an `@file` syntax which I'm guessing would avoid this problem.

qiwzhang:
Not clear why it matters.  Will the stdout of the script be piped to a public site?  If so, can that individual user change the script to use @file.

dseomn:
> Not clear why it matters.

It matters if anybody runs the script on a shared workstation or server. Probably not the most common use case, but the comment above the line I linked to seems wrong at the moment.

> Will the stdout of the script be piped to a public site?

I don't think anything is leaked on stdout, just in the command line arguments.

dseomn:
Looking at this more, there's another place where this is not secure in a multi-user environment. This `cd` command looks like it's meant to ensure that the `service.json` and `Dockerfile` files are in a secure directory owned by the current user:

https://github.com/GoogleCloudPlatform/esp-v2/blob/948edb9658114b99d95452d40e43dd4a67b4f544/docker/serverless/gcloud_build_image#L95

But even with `set -eo pipefail`, if `mktemp` fails, the cd command won't do anything. E.g., running `bash -c 'set -eo pipefail; cd "$(false)"; pwd'` prints out the current directory with no error.

PR #730, opened 2022-09-02
Update gcloud_build_image script to be more safe.

qiwzhang:
To fix https://github.com/GoogleCloudPlatform/esp-v2/issues/712

To fix two issues:

1) includes the output of `gcloud auth print-access-token` on the command line 

2)  does not check the return value of `mktemp -d /tmp/docker.XXXX`,

### DIFF
--- a/docker/serverless/gcloud_build_image
+++ b/docker/serverless/gcloud_build_image
@@ -92,10 +92,12 @@ if [ -z "${ESP_FULL_VERSION}" ]; then
 fi
 echo "Building image for ESP version: ${ESP_FULL_VERSION}"
 
-cd "$(mktemp -d /tmp/docker.XXXX)"
+tempdir="$(mktemp -d /tmp/docker.XXXX)"
+cd "$tempdir"
 
 # Be careful about exposing the access token.
-curl --fail -o "service.json" -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+access_token="$(gcloud auth print-access-token)"
+curl --fail -o "service.json" -H "Authorization: Bearer ${access_token}" \
   "https://servicemanagement.googleapis.com/v1/services/${SERVICE}/configs/${CONFIG_ID}?view=FULL" \
   || error_exit "Failed to download service config"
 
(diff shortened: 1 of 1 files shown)


docker/serverless/gcloud_build_image
```diff
@@ -92,10 +92,12 @@ if [ -z "${ESP_FULL_VERSION}" ]; then
 fi
 echo "Building image for ESP version: ${ESP_FULL_VERSION}"
 
-cd "$(mktemp -d /tmp/docker.XXXX)"
+tempdir="$(mktemp -d /tmp/docker.XXXX)"
+cd "${tempdir}"
 
 # Be careful about exposing the access token.
-curl --fail -o "service.json" -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+access_token="$(gcloud auth print-access-token)"
```
shuoyang2016, line 99: Is this token not exposed in this case? Can we add --scopes=...?

shuoyang2016 (commented):
Please let the filer review this change too. Thanks.

docker/serverless/gcloud_build_image
dseomn: Yeah, this doesn't solve the problem at all because `${access_token}` is still in the command line below. It needs to be in standard input or somewhere else that's private. For standard input:
```sh
curl ... -H @- ... <<EOF
Authorization: Bearer ${access_token}
EOF
```

qiwzhang: Done

docker/serverless/gcloud_build_image
```diff
@@ -92,12 +92,15 @@ if [ -z "${ESP_FULL_VERSION}" ]; then
 fi
 echo "Building image for ESP version: ${ESP_FULL_VERSION}"
 
-cd "$(mktemp -d /tmp/docker.XXXX)"
+tempdir="$(mktemp -d /tmp/docker.XXXX)"
```
TAOXUY, line 95: No need to change?

TAOXUY (approved):

docker/serverless/gcloud_build_image
qiwzhang: The fix is to check the status of "mktemp".  Before the fix, it is not checked.

dseomn (approved):

PR #737, opened 2022-10-03
fix build_envoy_image failure

qiwzhang:

It failed with error:
```
Step #0 - "build-envoy-image": ERROR: alpine-baselayout-data-3.2.0-r23: trying to overwrite etc/nsswitch.conf owned by glibc-2.35-r0.
```

Not to upgrade to glic2.35,  our envoy binary doesn't compatible with it, it failed to load with errors:

```
Error relocating /lib/ld-linux-x86-64.so.2: unsupported relocation type 37
```

### DIFF
--- a/docker/Dockerfile-envoy
+++ b/docker/Dockerfile-envoy
@@ -21,7 +21,7 @@ ENV LANG=C.UTF-8
 # Here we install GNU libc (aka glibc) and set C.UTF-8 locale as default.
 
 RUN ALPINE_GLIBC_BASE_URL="https://github.com/sgerrand/alpine-pkg-glibc/releases/download" && \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile-envoy
```diff
@@ -61,7 +61,8 @@ RUN ALPINE_GLIBC_BASE_URL="https://github.com/sgerrand/alpine-pkg-glibc/releases
 
 
 # Update to the latest
```
TAOXUY, line 63: NIT:typo pin

TAOXUY:
Could you link the bug or github issue to provide more context?

docker/Dockerfile-envoy
qiwzhang: Done

nareddyt (approved):

PR #753, opened 2022-11-17
Add two jwt_authn flags: "jwt_cache_size" and "jwks_async_fetch_fast_listener"

qiwzhang:

The flag `--jwt_cache_size` can be used to enable jwt cache

The flag `--jwks_async_fetch_fast_listener` is useful in some rare cases where async_fetch first fetch is slow or it always fails and retrying.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -581,7 +581,7 @@ environment variable or by passing "-k" flag to this script.
         action='store_true',
         default=False,
         help='''
-        Only apply when jwks_async_fetch is enabled. This flag determines if the envoy will wait
+        Only apply when --disable_jwks_async_fetch flag is not set. This flag determines if the envoy will wait
         for jwks_async_fetch to complete before binding the listener port. If false, it will wait.
         Default is false.'''
     )
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -577,6 +577,22 @@ environment variable or by passing "-k" flag to this script.
         to the request processing latency. Default is enabled.'''
     )
     parser.add_argument(
+        '--jwks_async_fetch_fast_listener',
+        action='store_true',
+        default=False,
+        help='''
+        Only apply when jwks_async_fetch is enabled. This flag determines if the envoy will wait
+        for jwks_async_fetch to complete before binding the listener port. If false, it will wait.
+        Default is false.'''
+    )
+    parser.add_argument(
+        '--jwt_cache_size',
+        default=None,
+        help='''
+        Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores
+        verified good tokens. If 0, JWT cache is disabled. The default is 0.'''
+    )
+    parser.add_argument(
         '--jwks_cache_duration_in_s',
         default=None,
         help='''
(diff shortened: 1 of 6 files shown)


nareddyt (approved):

docker/generic/start_proxy.py
```diff
@@ -576,6 +576,22 @@ def make_argparser():
         If disabled, JWKS fetching is done when authenticating the JWT, the fetching will add
         to the request processing latency. Default is enabled.'''
     )
+    parser.add_argument(
+        '--jwks_async_fetch_fast_listener',
+        action='store_true',
+        default=False,
+        help='''
+        Only apply when jwks_async_fetch is enabled. This flag determines if the envoy will wait
```
TAOXUY, line 584: when `--disable_jwks_async_fetch` isn't set

src/go/configmanager/flags/flags.go
```diff
@@ -135,14 +135,16 @@ var (
...
 			If not provided, Envoy will decide the default value.`)
 
-	DisableJwksAsyncFetch = flag.Bool("disable_jwks_async_fetch", defaults.DisableJwksAsyncFetch, `When the feature is enabled, JWKS is fetched before processing any requests. When disabled, JWKS is fetched on-demand when processing the requests.`)
-	JwksCacheDurationInS  = flag.Int("jwks_cache_duration_in_s", defaults.JwksCacheDurationInS, "Specify JWT public key cache duration in seconds. The default is 5 minutes.")
+	DisableJwksAsyncFetch      = flag.Bool("disable_jwks_async_fetch", defaults.DisableJwksAsyncFetch, `When the feature is enabled, JWKS is fetched before processing any requests. When disabled, JWKS is fetched on-demand when processing the requests.`)
+	JwksAsyncFetchFastListener = flag.Bool("jwks_async_fetch_fast_listener", defaults.JwksAsyncFetchFastListener, `Only apply when jwks_async_fetch is enabled. This flag determines if the envoy will wait for jwks_async_fetch to complete before binding the listener port. If false, it will wait. Default is false.`)
```
TAOXUY, line 139: ditto

docker/generic/start_proxy.py
```diff
@@ -576,6 +576,22 @@ def make_argparser():
...
+        for jwks_async_fetch to complete before binding the listener port. If false, it will wait.
+        Default is false.'''
+    )
+    parser.add_argument(
```
TAOXUY, line 588: Why not enable them by default? 

If not, each JWT needs verification computation?

docker/generic/start_proxy.py

src/go/configmanager/flags/flags.go
qiwzhang: Done

docker/generic/start_proxy.py
qiwzhang: Agreed that it should be safe to enable it by default.  But it is a behavior change, you never know what will happen to an existing product.  Prefer to enable it by users.

TAOXUY (approved):

docker/generic/start_proxy.py
qiwzhang: Well,  I may use a separate pr to enable it.  Since it needs to modify many tests too and in case, it needs to rollback,  just rollback that pr.

Issue #462, opened 2021-01-14
Re-enable Ubsan in CI


nareddyt:
When updating envoy in #445, the Undefined behavior CI build keeps failing with:

```
ld.lld-10: error: undefined symbol: __muloti4
>>> referenced by int128_have_intrinsic.inc:251 (external/com_google_absl/absl/numeric/int128_have_intrinsic.inc:251)
>>>               bazel-out/k8-dbg/bin/external/com_google_absl/absl/strings/_objs/strings/numbers.pic.o:(absl::operator*(absl::int128, absl::int128))
```

This looks like a common bug mentioned in https://github.com/envoyproxy/envoy/issues/13973. I spent some time investigating but I cannot find a good solution.

Let's disable Ubsan for now to unblock development. Nighthawk also followed the same approach - https://github.com/envoyproxy/nighthawk/issues/593

PR #755, opened 2022-11-28
Update to Envoy 1.24.0

nareddyt:
Closes https://github.com/GoogleCloudPlatform/esp-v2/issues/462

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -64,13 +64,8 @@ build:asan --config=debug
 build:asan --define signal_trace=disabled
 build:asan --define ENVOY_CONFIG_ASAN=1
 
-# FIXME(nareddyt): Enable ubsan builds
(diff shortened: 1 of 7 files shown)


qiwzhang (approved):

TAOXUY (approved):

nareddyt:
Two problems caught by CI tests

1. ASAN build fails, logs are not helpful. Probably need to change some `cxxopt`
2. Quota tests are returning codes other than 200 and 429. I think I see a 403

nareddyt:
e2e test failed with the following error

```
D1129 03:40:58.098 26 src/envoy/http/service_control/http_call.cc:109] [26][filter]http call response status code: 403, body: \010\007\022_Permission \'servicemanagement.services.check\' denied on service \'<redacted_3rd_party_service>\'.
```

I think this is a flake unrelated to Envoy update. Let's see if it passes this time

Issue #743, opened 2022-10-12
ESPv2 not respecting the terminate signal


davidgwps:
Hey, 

How do I gracefully shutdown a ESPv2 container? It doesn't seem that SIGTERM is being listened to.

I have just migrated from using ESP to ESPv2, and running into a problem whilst doing a rolling upgrade of a pod where the terminating pod stays stuck in a terminating state until the graceful shutdown period elapses and kubernetes kills it a few minutes later.

I've run the pod with and without the ESPv2 container, and this only happens with ESPv2 enabled and running locally in minikube or in GKE.

```yaml
        - name: esp
          image: gcr.io/endpoints-release/endpoints-runtime:2
          imagePullPolicy: Always
          args: [
            "--listener_port=8083",
            "--backend=127.0.0.1:8080",
            {{- if eq .Values.k8s.minikube.enabled false }}
            "--service={{ .Values.cloudEndpoints.hostName }}",
            "--rollout_strategy=managed",
            {{- else }}
            "--service_json_path=/etc/espv2_config/service_config.json",
            "--non_gcp",
            "--enable_debug",
            {{- end }}
          ]
          ports:
            - containerPort: 8083
          {{- if .Values.k8s.minikube.enabled }}
          volumeMounts:
            - mountPath: /etc/espv2_config
              name: service-config-volume
          {{- end }}
          readinessProbe:
            httpGet:
              path: /healthz
              port: 8083
          livenessProbe:
            httpGet:
              path: /healthz
              port: 8083
```

The debug logs in the ESPv2 container keep on churning out constantly, also indicating it is still running

[scratch.txt](https://github.com/GoogleCloudPlatform/esp-v2/files/9767240/scratch.txt)

kubes pod events showing that the container is attempted to be killed

```
│   Normal   Created    27m                kubelet            Created container esp                                                                                                                                                                        │
│   Normal   Started    27m                kubelet            Started container esp 
│   Normal   Created    27m                kubelet            Created container my_container                                                                                                                                                                   │
│   Normal   Started    27m                kubelet            Started container my_container                                                                                                                                                                   │
│   Warning  Unhealthy  27m (x2 over 27m)  kubelet            Readiness probe failed: HTTP probe failed with statuscode: 503                                                                                                                               │
│   Normal   Killing    3s                 kubelet            Stopping container esp                                                                                                                                                                       │
│   Normal   Killing    3s                 kubelet            Stopping container my_container 
```

Any help appreciated,
David

[...]

PR #756, opened 2022-11-29
start_proxy to pass SIGTERM signal to envoy and config_mgr 

qiwzhang:

To fix:  https://github.com/GoogleCloudPlatform/esp-v2/issues/743

start_proxy.py is the parent process starting envoy and config_mgr processes.   But it did not handle SIGTERM  signal, it cause the "kubectrl delete pod" needs to wait for `grace-peirod`.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -1520,7 +1520,14 @@ def start_envoy(args):
     return proc
 
 def sigterm_handler(signum, frame):
-    """ Handler for SIGTERM, pass the signal to all child processes. """
+    """ Handler for SIGTERM and SIGINT, pass the SIGTERM to all child processes. """
+    signame = signal.Signals(signum).name
+    logging.warning("{}: got signal: {}".format(datetime.utcnow().isoformat(timespec='microseconds'), signame))
+
+    shutdown()
+
+def siginfo_handler(signum, frame):
+    """ Handler for other signals, just log the signals."""
     signame = signal.Signals(signum).name
     logging.warning("{}: got signal: {}".format(datetime.utcnow().isoformat(timespec='microseconds'), signame))
 
@@ -1535,6 +1542,9 @@ def shutdown():
         except OSError:
             logging.error("error sending TERM to PID={} continuing".format(pid))
 
+    logging.info("exiting...")
+    sys.exit(0)
+
 if __name__ == '__main__':
     logging.basicConfig(format='%(levelname)s: %(message)s', level=logging.INFO)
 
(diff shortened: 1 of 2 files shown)


docker/generic/start_proxy.py
```diff
@@ -1525,6 +1544,11 @@ def start_envoy(args):
     cm_proc = start_config_manager(gen_proxy_config(args))
     envoy_proc = start_envoy(args)
 
+    pid_list.append(cm_proc.pid)
+    pid_list.append(envoy_proc.pid)
+    signal.signal(signal.SIGTERM, sigterm_handler)
+    signal.signal(signal.SIGINT, sigterm_handler)
```
nareddyt, line 1550: Why not handle all signals that envoy handles?

https://github.com/envoyproxy/envoy/blob/3752119b94ae4a5ff337364f585f54305406dc7f/source/server/server.cc#L849-L873

docker/generic/start_proxy.py
```diff
@@ -1515,6 +1519,21 @@ def start_envoy(args):
     t.start()
     return proc
 
+def sigterm_handler(signum, frame):
+    """ Handler for SIGTERM and SIGINT, pass the SIGTERM to all child processes. """
+    signame = signal.Signals(signum).name
+    logging.warning("{}: got signal: {}".format(datetime.utcnow().isoformat(timespec='microseconds'), signame))
```
nareddyt, line 1525: nit: change loging.basicConfig on line 1539 to log timestamps

docker/generic/start_proxy.py
```diff
@@ -1515,6 +1519,21 @@ def start_envoy(args):
...
+    global pid_list
+    for pid in pid_list:
+        logging.info("sending TERM to PID={}".format(pid))
+        try:
+            os.kill(pid, signal.SIGTERM)
+        except OSError:
+            logging.error("error sending TERM to PID={} continuing".format(pid))
+
+    logging.info("exiting...")
+    sys.exit(0)
```
nareddyt, line 1536: no need to exit, just let the health check on line 1553 handle this

nareddyt: Don't you need to give Envoy time to clean up when it receives SIGINT? Looking through Envoy's code, it does some items like flushing stats, finishing up any pending requests in the filter chain, etc. This may take a few seconds to complete. I imagine max 5 seconds.

You are exiting directly after SIGTERM, that is not enough time.

Even if you change to use health checker, you are only giving it 2 seconds with `HEALTH_CHECK_PERIOD` change. Maybe change it to 5 seconds.

docker/generic/start_proxy.py
nareddyt: Nevermind, we don't have hot restart

docker/generic/start_proxy.py
TAOXUY: You may follow http://shortn/_LbIgHpcwUK

docker/generic/start_proxy.py

docker/generic/start_proxy.py

nareddyt (approved):

PR #757, opened 2022-11-30
Increase ID token refetch buffer

qiwzhang:

The problem:
The current refetch buffer is 5s,  we refetch a ID token 5s before it is expired.  But in some networks,  it takes more than 5s to send the token to the backend Cloud Function.   When it is used, it is expired.

Here we increase the overall token refetch buffer to 60s.   Token usually expired in 1 hour, refetch them 60s before 1 hour expiration should be ok.  It should not increase too much load to the token server.

This increase is for all token fetch,   access token and ID token.


nareddyt (approved):

qiwzhang:
Yes, I did an internal search.  We only use Google access tokens,  they expire in 1 hour too.

PR #759, opened 2022-12-01
Enable jwt cache by default for jwt authentication.

qiwzhang:
Benefits JWT cache are
*  Save a lot of CPS consumption as JWT signature verification uses a lot of CPU. 
*  Reduce remote jwks fetch.   jwks cache is 5m,  but jwt lifetime is 1 hour so it can be cached for 1 hour. During this 1 hour, jwks is not needed.  Sometime jwks fetch may fail, it will cause the requests been rejected.

It should be safe to to enable it as I don't see any draw back.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -589,8 +589,9 @@ environment variable or by passing "-k" flag to this script.
         '--jwt_cache_size',
         default=None,
         help='''
-        Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores
-        verified good tokens. If 0, JWT cache is disabled. The default is 0.'''
+        Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified
+        good tokens. If 0, JWT cache is disabled. It limits the memory usage. The cache used memory
+        is roughly (token size + 64 bytes) per token. If not specified, the default is 100000.'''
     )
     parser.add_argument(
         '--jwks_cache_duration_in_s',
--- a/src/go/configmanager/flags/flags.go
+++ b/src/go/configmanager/flags/flags.go
@@ -144,7 +144,7 @@ var (
 	JwksFetchRetryBackOffMaxIntervalMs  = flag.Int("jwks_fetch_retry_back_off_max_interval_ms", int(defaults.JwksFetchRetryBackOffMaxInterval.Milliseconds()), `Specify JWKS fetch retry exponential back off maximum interval in milliseconds. The default is 32 seconds.`)
 	JwtPatForwardPayloadHeader          = flag.Bool("jwt_pad_forward_payload_header", defaults.JwtPadForwardPayloadHeader, `For the JWT in request, the JWT payload is forwarded to backend in the "X-Endpoint-API-UserInfo"" header by default. 
 Normally JWT based64 encode doesn’t add padding. If this flag is true, the header will be padded.`)
-	JwtCacheSize = flag.Uint("jwt_cache_size", defaults.JwtCacheSize, `Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified good tokens. If 0, JWT cache is disabled. The default is 1000.`)
+	JwtCacheSize = flag.Uint("jwt_cache_size", defaults.JwtCacheSize, `Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified good tokens. If 0, JWT cache is disabled. It limits the memory usage. The cache used memory is roughly (token size + 64 bytes) per token. If not specified, the default is 100000.`)
 
 	ScCheckTimeoutMs  = flag.Int("service_control_check_timeout_ms", defaults.ScCheckTimeoutMs, `Set the timeout in millisecond for service control Check request. Must be > 0 and the default is 1000 if not set.`)
 	ScQuotaTimeoutMs  = flag.Int("service_control_quota_timeout_ms", defaults.ScQuotaTimeoutMs, `Set the timeout in millisecond for service control Quota request. Must be > 0 and the default is 1000 if not set.`)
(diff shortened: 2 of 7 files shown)


src/go/configmanager/flags/flags.go
```diff
@@ -144,7 +144,7 @@ var (
 	JwksFetchRetryBackOffMaxIntervalMs  = flag.Int("jwks_fetch_retry_back_off_max_interval_ms", int(defaults.JwksFetchRetryBackOffMaxInterval.Milliseconds()), `Specify JWKS fetch retry exponential back off maximum interval in milliseconds. The default is 32 seconds.`)
 	JwtPatForwardPayloadHeader          = flag.Bool("jwt_pad_forward_payload_header", defaults.JwtPadForwardPayloadHeader, `For the JWT in request, the JWT payload is forwarded to backend in the "X-Endpoint-API-UserInfo"" header by default. 
 Normally JWT based64 encode doesn’t add padding. If this flag is true, the header will be padded.`)
-	JwtCacheSize = flag.Uint("jwt_cache_size", defaults.JwtCacheSize, `Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified good tokens. If 0, JWT cache is disabled. The default is 0.`)
+	JwtCacheSize = flag.Uint("jwt_cache_size", defaults.JwtCacheSize, `Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified good tokens. If 0, JWT cache is disabled. The default is 1000.`)
```
nareddyt, line 147: Will it be user configurable? No change in `start_proxy.go`?

nareddyt, line 147: Curious why you choose 1000. This seems pretty small for 1 whole hour TBH.

qiwzhang: it is the number of unique tokens in 1 hour.      My gut feeling is it should be small, e.g. SA tokens

For end user tokens,  it is too small.  

It limits the memory we allocated for the cache.   Normally 

a token is  100B,  or 0.1K,  
1k tokens is 0.1M.   
100k tokens? which is 10M

I guess we can set the cache size as 1m,  which is only 100M memory.

What do you think?

qiwzhang: Yes, I forget,  we need to expose this in start_proxy.py

nareddyt: Let's do 100k tokens or 10MB.

Based on Cloud ESF benchmark report, baseline memory usage is 200 MB. We can add an extra 5% for cache. 100M extra is 50%, too much.

(I am assuming ESPv2 has similar performance)

nareddyt: Also I don't know how the cache is implemented internally, but it may have extra memory usage per item for LRU or tracking. With this assumption 100k tokens or 10MB makes sense still

qiwzhang: It is already exposed.  Just need to update the help text with new default

src/go/configmanager/flags/flags.go
```diff
@@ -144,7 +144,7 @@ var (
 	JwksFetchRetryBackOffMaxIntervalMs  = flag.Int("jwks_fetch_retry_back_off_max_interval_ms", int(defaults.JwksFetchRetryBackOffMaxInterval.Milliseconds()), `Specify JWKS fetch retry exponential back off maximum interval in milliseconds. The default is 32 seconds.`)
 	JwtPatForwardPayloadHeader          = flag.Bool("jwt_pad_forward_payload_header", defaults.JwtPadForwardPayloadHeader, `For the JWT in request, the JWT payload is forwarded to backend in the "X-Endpoint-API-UserInfo"" header by default. 
 Normally JWT based64 encode doesn’t add padding. If this flag is true, the header will be padded.`)
-	JwtCacheSize = flag.Uint("jwt_cache_size", defaults.JwtCacheSize, `Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified good tokens. If 0, JWT cache is disabled. The default is 0.`)
+	JwtCacheSize = flag.Uint("jwt_cache_size", defaults.JwtCacheSize, `Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified good tokens. If 0, JWT cache is disabled. It limits the memory usage. The cache used memory is roughly (token size + 64 bytes) per token. If not specified, the default is 100000.`)
```
TAOXUY, line 147: NIT

`If 0, JWT cache is disabled. It limits the memory usage. The cache used memory is roughly (token size + 64 bytes) per token. If not specified, the default is 100000.`

The max cache memory overhead is roughly (token_size + 64 bytes) * jwt_cache_size. If the flag is set with 0, JWT cache is disabled and no extra memory usage is introduced. If not specified, the flag is set with 100000 by default.

docker/generic/start_proxy.py
```diff
@@ -589,8 +589,9 @@ def make_argparser():
         '--jwt_cache_size',
         default=None,
         help='''
-        Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores
-        verified good tokens. If 0, JWT cache is disabled. The default is 0.'''
+        Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified
+        good tokens. If 0, JWT cache is disabled. It limits the memory usage. The cache used memory
```
TAOXUY, line 593: ditto

TAOXUY (approved):

src/go/configmanager/flags/flags.go
qiwzhang: also no need to mention, not extra memory usage for disabled cache

src/go/configmanager/flags/flags.go
mfiedorowicz: @nareddyt @qiwzhang https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/jwt_authn/jwt_cache.cc#L25-L27 shows that config with `jwt_cache_size` set to `0` is not disabled (as flag description states here) but uses envoy's default cache size `100`

PR #767, opened 2022-12-13
Disable tracing in e2e tests

nareddyt:
ESPv2 is stable and we don't verify traces published via e2e. Disable it to allow better visualization of Cloud ESF traces. Integration tests in ESPv2 still verify tracing behavior.


qiwzhang:
Cloud ESF and ESPv2 has different trace space.   I still prefer to keep trace on for its e2e tests.   I remember I still used them.

qiwzhang (changes requested):
Please don't disable it

PR #769, opened 2022-12-16
 b/262774083: Fix trace id in the endpoint log

qiwzhang:

If the flag `--tracing_project_id` is used,  `trace_id` in the endpoint log is wrong.  It always uses the GCP deploy project. 

If trace is disabled,  we also write trace_id in the log,   We should not write it.


nareddyt (approved):
Worth having integration test where you specify flag?

PR #771, opened 2022-12-20
Update janitor script for cloud-api-proxy-testing

nareddyt:
- Run on other Cloud ESF project
- Remove filter timestamp for target pools. It is OK to delete them early, as there should never be any unused target pools.

### DIFF
--- a/prow/janitor.sh
+++ b/prow/janitor.sh
@@ -10,163 +10,170 @@
 set -eo pipefail
 
 # Set the project id
-PROJECT="cloudesf-testing"
-REGION="us-central1"
-gcloud config set project ${PROJECT}
-gcloud config set run/region "${REGION}"
+PROJECT_IDS=("cloudesf-testing" "cloud-api-proxy-testing")
+REGIONS=("us-central1" "us-east1")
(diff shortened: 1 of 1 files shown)


prow/janitor.sh
```diff
@@ -9,164 +9,171 @@
...
-  gcloud app services delete "${ap}" \
-    --quiet
-done
-echo "Done cleaning up App Engines"
+# Set the project id
+PROJECT_IDS=("cloudesf-testing" "cloud-api-proxy-testing")
+REGIONS=("us-central1" "us-east1")
+for PROJECT in ${PROJECT_IDS[@]}; do
+  for REGION in ${REGIONS[@]}; do
+    echo "cleaning up resources in ${REGION} for ${PROJECT}"
```
qiwzhang, line 21: will we bail out the whole script if there is any CLI failed?

since cloud-esf-testing, only use us-cdentral1

cloud-api-proxy-testing only use us-east1

the other combinations may fail some CLI.

qiwzhang: Cool

qiwzhang (approved):

PR #777, opened 2023-01-13
Add HTTP backend info for each method in config generator and a test-only HTTP backend address

shuoyang2016:


src/go/configinfo/service_info.go
```diff
@@ -401,8 +415,9 @@ func addHttpRule(method *MethodInfo, r *annotationspb.HttpRule, addedRouteMatchW
 	}
 
 	httpRule := &httppattern.Pattern{
-		HttpMethod:  httpMethod,
-		UriTemplate: uriTemplate,
+		HttpMethod:     httpMethod,
+		UriTemplate:    uriTemplate,
+		UseHTTPBackend: !IsDiscoveryAPI(method.Operation()) && s.Options.TestOnlyHTTPBackendAddress != "",
```
qiwzhang, line 420: I think we should not use "useHTTPBackend" in httpRule.Pattern.  It is overloaded.

Can we just have a "fromHTTPRule" to indicate it is generated from http.rule,  not from RPC methods?

Which backend it should use can be dealt at routeGenerate time?  I think we should have enough info for that.

maybe we don't even need this

 we can say:  `POST /service_name/method_name` is a grpc method,  the others are HTTP.

src/go/configinfo/service_info.go
```diff
@@ -940,15 +992,19 @@ func (s *ServiceInfo) LocalBackendClusterName() string {
 	return util.BackendClusterName(fmt.Sprintf("%s_local", s.Name))
 }
 
-func (s *ServiceInfo) isDiscoveryAPI(operation string) bool {
-	return strings.HasPrefix(operation, "google.discovery") && !s.Options.AllowDiscoveryAPIs
+func (s *ServiceInfo) LocalTestOnlyHttpBackendClusterName() string {
+	return util.BackendClusterName(fmt.Sprintf("%s_local_http", s.Name))
+}
+
+func (s *ServiceInfo) isDiscoveryAPIbutAPIDiscoveryNotEnabled(operation string) bool {
```
qiwzhang, line 999: maybe called "shouldSkipDiscoveryAPI"?

src/go/configinfo/service_info.go
```diff
@@ -161,20 +168,20 @@ func NewServiceInfoFromServiceConfig(serviceConfig *confpb.Service, id string, o
 	return serviceInfo, nil
 }
 
-func (s *ServiceInfo) buildLocalBackend() error {
-
-	scheme, hostname, port, _, err := util.ParseURI(s.Options.BackendAddress)
+func (s *ServiceInfo) buildBackend(address string, isTestOnlyHTTPBackend bool) error {
```
qiwzhang, line 171: keep it as `buildLocalBackend`.    We have local and remote,   remote are from `Backend.address`,  local is from the flag.    Maybe `buildBackendFromFlag`?

src/go/configinfo/service_info.go
shuoyang2016: This is a good idea. Indeed I found we should be able to use path to know it's GRPC. And we don't need to add field in httpRule.Pattern. PTAL.

src/go/configinfo/service_info.go
shuoyang2016: Done.

src/go/configinfo/service_info.go
shuoyang2016: Combining your comment, I changed it to buildLocalBackendFromAddress.

src/go/configgenerator/route_generator.go
```diff
@@ -356,17 +356,22 @@ func MakeRouteTable(serviceInfo *configinfo.ServiceInfo) ([]*routepb.Route, []*r
 		}
 
 		for _, routeMatcher := range routeMatchers {
-			r := makeRoute(routeMatcher, method)
+			bi := method.BackendInfo
+			useLocalHTTPBackend := !method.IsGRPCPath(routeMatcher.GetPath()) && !configinfo.IsDiscoveryAPI(method.Operation()) && method.HttpBackendInfo != nil
+			if useLocalHTTPBackend {
+				bi = method.HttpBackendInfo
+			}
+			r := makeRoute(routeMatcher, method, useLocalHTTPBackend)
```
qiwzhang, line 364: maybe pass in whole "bi"  instead of bool

qiwzhang (approved):
Just a small NIT.

src/go/configgenerator/route_generator.go
shuoyang2016: That was my idea too. But since backendInfo is a private field, I used this way so to avoid expose that to public and change other interfaces. Since this is nit, I will just use the current implementation.

Issue #706, opened 2022-06-08
Add ability to disable JWT audience checking in ESPv2


hravnaas:
We have the need to at least temporarily disable the JWT "aud" claim check in ESPv2. My organization does not currently include the "aud" claim and it will be a slow process to get it added. Also, it'd just be a static string used in all tokens so it would provide very little value if any all all.

It seems that the JWT RFC calls out "aud" as optional. I've tried removing aud from the token and x-google-audiences from the OpenAPI doc, but I'm getting the "Audience not allowed" error still. Same error when using x-google-audiences: "", so setting it to an empty string. I can however get an empty string to work if I use x-google-audiences: "," but that still requires the aud claim to be present in the token, so "aud: ""

The JWT is only used internally in our org. It will not be used externally.

As mentioned above, with the RFC labeling "aud" as optional, it seems appropriate that ESPv2 would allow it to truly be optional. See statement [here](https://2rfc.net/7519#4.1.3:~:text=generally%20application%20specific.-,Use%20of%20this%20claim%20is%20OPTIONAL.,-4.1.4.%C2%A0%C2%A0%22exp).

Works:
`securityDefinitions:
  jwt_company:
    authorizationUrl: "https://www.company.com"
    flow: "implicit"
    type: "oauth2"
    x-google-issuer: "company.com"
    x-google-jwks_uri: "https://storage.googleapis.com/somefile.json"
    x-google-audiences: "webserver.company.com"`

Fails (no x-google-audiences)
`securityDefinitions:
  jwt_company:
    authorizationUrl: "https://www.company.com"
    flow: "implicit"
    type: "oauth2"
    x-google-issuer: "company.com"
    x-google-jwks_uri: "https://storage.googleapis.com/somefile.json"`

If aud is missing, the default service name seems to be added in today: https://github.com/GoogleCloudPlatform/esp-v2/blob/master/src/go/configgenerator/filterconfig/filter_gen_jwt_authn.go#L118

Perhaps a solution could be to allow both an empty string or no aud claim at all if this was provided: x-google-audiences: ""

eduardchai:
Hi, I encounter the same issue. Is there any workaround for this so far?

eduardchai:
Thank you @qiwzhang for the fix. Any ETA for the release? Thank you.

eduardchai:
Hi @qiwzhang, is there any update on this? Thank you.

PR #779, opened 2023-01-18
Add flag --disable_jwt_audience_service_name_check

qiwzhang:
To fix: https://github.com/GoogleCloudPlatform/esp-v2/issues/706

Normally JWT "aud" field is checked against audiences specified in OpenAPI `x-google-audiences` field.
This flag changes the behavior when the `x-google-audiences` is not specified.
 When the `x-google-audiences` is not specified, normally the service name is used to check the JWT `aud` field.
 If this flag is true, the service name is not used, JWT `aud` field will not be checked.

### DIFF
--- a/src/go/configgenerator/filterconfig/filter_gen_jwt_authn.go
+++ b/src/go/configgenerator/filterconfig/filter_gen_jwt_authn.go
@@ -108,16 +108,18 @@ var jaFilterGenFunc = func(serviceInfo *ci.ServiceInfo) (*hcmpb.HttpFilter, []*c
 			PadForwardPayloadHeader: serviceInfo.Options.JwtPadForwardPayloadHeader,
 		}
 
-		if len(provider.GetAudiences()) != 0 {
-			for _, a := range strings.Split(provider.GetAudiences(), ",") {
-				jp.Audiences = append(jp.Audiences, strings.TrimSpace(a))
+		if !serviceInfo.Options.DisableJwtAudCheck {
+			if len(provider.GetAudiences()) != 0 {
+				for _, a := range strings.Split(provider.GetAudiences(), ",") {
+					jp.Audiences = append(jp.Audiences, strings.TrimSpace(a))
+				}
+			} else {
+				// No providers specified by user.
+				// For backwards-compatibility with ESPv1, auto-generate audiences.
+				// See b/147834348 for more information on this default behavior.
+				defaultAudience := fmt.Sprintf("https://%v", serviceInfo.Name)
+				jp.Audiences = append(jp.Audiences, defaultAudience)
 			}
-		} else {
-			// No providers specified by user.
-			// For backwards-compatibility with ESPv1, auto-generate audiences.
-			// See b/147834348 for more information on this default behavior.
-			defaultAudience := fmt.Sprintf("https://%v", serviceInfo.Name)
-			jp.Audiences = append(jp.Audiences, defaultAudience)
 		}
 
 		if serviceInfo.Options.JwtCacheSize > 0 {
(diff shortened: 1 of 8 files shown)


src/go/configgenerator/filterconfig/filter_gen_jwt_authn.go
```diff
@@ -108,16 +108,18 @@ var jaFilterGenFunc = func(serviceInfo *ci.ServiceInfo) (*hcmpb.HttpFilter, []*c
 			PadForwardPayloadHeader: serviceInfo.Options.JwtPadForwardPayloadHeader,
 		}
 
-		if len(provider.GetAudiences()) != 0 {
-			for _, a := range strings.Split(provider.GetAudiences(), ",") {
-				jp.Audiences = append(jp.Audiences, strings.TrimSpace(a))
+		if !serviceInfo.Options.DisableJwtAudCheck {
```
nareddyt, line 111: This logic seems incorrect.

My understanding is that the user wants this behavior:
1) If audience is empty in config and `--disable_jwt_aud_check=true`, then don't check audience for this selector
2) If audience is empty in config and `--disable_jwt_aud_check=false`, then check for `defaultAudience` for this selector
3) If audience is specified in config, always check for specified audience in selector, regardless of `--disable_jwt_aud_check` flag.

This CL breaks 3

nareddyt (changes requested):

src/go/configgenerator/filterconfig/filter_gen_jwt_authn.go
qiwzhang: In my mind, when the flag `--disable_jwt_aud_check` is used,   we just don't check `aud` field in JWT regardless the what is in the service config, period.   This is brutal,  but very easy to understand.   

So I don't agree with your 3).

nareddyt: Why though?

User wants a new flag added because they want to disable JWT aud check for some selectors be adding empty aud ("") to their config. But we currently default empty aud to `defaultAudience`

User didn't ask for a flag to completely disable audience check. User just wants to disable this default behavior when empty aud is in their config.

A better flag to solve this issue is `--disable_default_empty_jwt_aud` or similar

nareddyt: Think about it this way: User has full control over the config.

If they want to completely disable JWT aud checking, they can set `--disable_default_empty_jwt_aud=true` and then modify their API config so every selector has empty ("") audience.

No need for a bigger scoped `--disable_jwt_aud_check` flag

qiwzhang: My concern is:  users will not be able to find `--disable_default_empty_jwt_aud` flag before we learn our such behaviors,they have to do a lot of testing it and debugging to learn the behaviors,  then start to use the flag to only disable it.  

It will be much easier for users to find `--disable_jwt_aud_check` flag, since it is obvious, and brutal.

The key question is: `--disable_jwt_aud_check` flag will not support partial aud checking,  some methods checking "aud" and some other methods not checking.   I think this should be fine, usually they want the same behaviors for all methods.

shuoyang2016: This is a niche feature. We should've told customers that we don't implement non-security related features any more.

Seems either option is OK. Completely disable aud check could be fine given "aud" is optional; so there shouldn't be any security issues? But customers do mention "a solution could be to allow both an empty string or no aud claim at all if this was provided: x-google-audiences: "".

My suggestion would be check back with the customers and implement whichever is simple.

qiwzhang: OK,  let us just wrap it up and finish it since it is already started.  

I will go with @nareddyt approach,  add a new flag with small scope,  to only disable using service_name with empty audiences.  Will update the doc to make it easy to find the flag (which is my main concern of this approach).

qiwzhang: The new flag name is called as `--disable_jwt_service_name_audience`

qiwzhang: Done

qiwzhang:
PTAL

nareddyt (approved):
Thanks for adjusting the flag!

PR #789, opened 2023-03-15
refactor: move business logic to filter generators

nareddyt:
- Remove `PerRouteConfigGenerator` from `MethodInfo`. We should not be mutating the configinfo inputs during filter generation. Instead, each filter generator's `GenPerRouteConfig()` is called on all routes, and the filter decides if the route needs a config generated for it.
- Move the enablement of filters to each filter generator and remove all business logic from `MakeFilterGenerators()`. This allows us to re-use the filter generators internally, or add new filter generators in the future.

Testing done: All unit tests pass with no change to configs.

### DIFF
--- a/src/go/configgenerator/filter_generator.go
+++ b/src/go/configgenerator/filter_generator.go
@@ -38,7 +38,8 @@ type FilterGenerator interface {
 	// have per-route configurations.
 	GenFilterConfig(*ci.ServiceInfo) (*hcmpb.HttpFilter, error)
 
-	// GenPerRouteConfig generates the per-route config for the given HTTP route.
+	// GenPerRouteConfig generates the per-route config for the given HTTP route (HTTP pattern).
+	// The MethodInfo that contains the route is also provided.
 	//
 	// This method is called on all routes. Return (nil, nil) to indicate the
 	// filter does NOT require a per-route config for the given route.
--- a/src/go/configgenerator/filter_generator.go
+++ b/src/go/configgenerator/filter_generator.go
@@ -28,68 +28,50 @@ type FilterGenerator interface {
 	// FilterName returns the name of the filter.
 	FilterName() string
 
-	// GenFilterConfig generates the filter config and a list of methods that need per-route configs.
-	GenFilterConfig(*ci.ServiceInfo) (*hcmpb.HttpFilter, []*ci.MethodInfo, error)
+	// IsEnabled returns true if the filter config should be generated.
+	// If false, none of the generation methods will be called.
+	IsEnabled() bool
+
+	// GenFilterConfig generates the filter config.
+	//
+	// Return (nil, nil) if the filter has no listener-level config, but may
+	// have per-route configurations.
+	GenFilterConfig(*ci.ServiceInfo) (*hcmpb.HttpFilter, error)
 
 	// GenPerRouteConfig generates the per-route config for the given HTTP route.
-	// This method is only called on the routes that `GenFilterConfig` returns.
+	//
+	// This method is called on all routes. Return (nil, nil) to indicate the
+	// filter does NOT require a per-route config for the given route.
 	GenPerRouteConfig(*ci.MethodInfo, *httppattern.Pattern) (*anypb.Any, error)
 }
 
 // MakeFilterGenerators provide of a slice of FilterGenerator in sequence.
 func MakeFilterGenerators(serviceInfo *ci.ServiceInfo) ([]FilterGenerator, error) {
-	var filterGenerators []FilterGenerator
+	return []FilterGenerator{
+		filtergen.NewCORSGenerator(serviceInfo),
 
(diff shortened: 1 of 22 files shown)


src/go/configgenerator/listener_generator.go
```diff
@@ -49,40 +49,33 @@ func MakeListeners(serviceInfo *sc.ServiceInfo) ([]*listenerpb.Listener, error)
...
-		})
-	}
-	return nil
-
-}
-
 // GetFilterConfigAndAddPerRouteConfigGen does two things
 //   - Return all http filter configs in a list
 //   - In place add the perRouteConfigGen function to all methods defined in serviceInfo
 func GetFilterConfigAndAddPerRouteConfigGen(serviceInfo *sc.ServiceInfo, filterGenerators []FilterGenerator) ([]*hcmpb.HttpFilter, error) {
```
TAOXUY, line 55: Rename to `MakeHttpFilterConfigs`?
Remove `- In place add the perRouteConfigGen function to all methods defined in serviceInfo`

TAOXUY (approved):
Nice refactor!

src/go/configgenerator/filter_generator.go
```diff
@@ -28,68 +28,50 @@ type FilterGenerator interface {
...
+	// GenFilterConfig generates the filter config.
+	//
+	// Return (nil, nil) if the filter has no listener-level config, but may
+	// have per-route configurations.
+	GenFilterConfig(*ci.ServiceInfo) (*hcmpb.HttpFilter, error)
 
 	// GenPerRouteConfig generates the per-route config for the given HTTP route.
-	// This method is only called on the routes that `GenFilterConfig` returns.
+	//
+	// This method is called on all routes. Return (nil, nil) to indicate the
```
shuoyang2016, line 43: Is it called on the methodInfo? Is that equal to route?

shuoyang2016:
> Nice refactor!

+1. We are moving towards better modularization and pkg encapsulations.

shuoyang2016 (approved):

src/go/configgenerator/filter_generator.go
```diff
@@ -28,68 +28,50 @@ type FilterGenerator interface {
...
-			CompressorType: filtergen.GzipCompressor,
-		})
-		filterGenerators = append(filterGenerators, &filtergen.CompressorGenerator{
-			CompressorType: filtergen.BrotliCompressor,
-		})
-	}
+		// Health check filter is behind Path Matcher filter, since Service Control
+		// filter needs to get the corresponding rule for health check in order to skip Report
+		filtergen.NewHealthCheckGenerator(serviceInfo),
+		filtergen.NewCompressorGenerator(serviceInfo, filtergen.GzipCompressor),
```
shuoyang2016, line 56: On a second thought, are we able to abstract the filter list as well? Then for the mode don't need these filters, we don't need to bother import the code even.

src/go/configgenerator/listener_generator.go
nareddyt: Done

src/go/configgenerator/filter_generator.go
nareddyt: MethodInfo has multiple routes (HTTP patterns). This function generates config on the HTTP pattern, but the wrapping MethodInfo is also useful to have for generation.

nareddyt: BTW this function does NOT generate route configs for all HTTP patterns under a method. It generates it for a single HTTP pattern at a time. Doing it one-at-a-time makes it easier for caller to fill in route table config.

src/go/configgenerator/filter_generator.go
nareddyt: Not sure I am understanding.

My thinking is that for each additional modes, we just create a new `MakeFilterGenerators` function internally. The function returns only the imported grpc_web + grpc_transcoder and new ext_auth + ext_proc filters generators.

nareddyt:
> > Nice refactor!
> 
> +1. We are moving towards better modularization and pkg encapsulations.

Thanks! In some follow-up CLs, I'll try to modularize the ServiceInfo. My goal is to reduce the amount of inputs needed to generate filter configs. That way, maybe we can create relevant filters (grpc_web, grpc_transcoder, ext_auth, ext_proc) without even having API service config and ServiceInfo, only proto descriptor and Options.

src/go/configgenerator/filter_generator.go
shuoyang2016: `MakeFilterGenerators` for each new mode SG. I thought we are going to use one `MakeFilterGenerators` to include all the filters for every mode, and isEnabled is used to determine the final result. But if we have `MakeFilterGenerators` for different mode, we don't need to bother import the filter gen code we don't need.

nareddyt: Yup exactly :) The `isEnabled()` is actually to keep pre-existing behavior. I don't want `GenFilterConfig()` and `GenPerRouteConfig()` to be called if the filter is not enabled.

PR #790, opened 2023-03-16
debug: change integration test JWT to have a large payload

nareddyt:
For verification that Cloud ESF can correctly handle large (> 4 KB) JWT payloads. All integration tests in `ESPv2-presubmit` pass with the long token.


nareddyt:
PTAL @TAOXUY or @yaoluz

TAOXUY (approved):

PR #792, opened 2023-03-20
Update a few packages in go.mod.

shuoyang2016:


nareddyt (approved):

tests/integration_test/grpc_integration_test/grpc_integration_test.go
```diff
@@ -122,7 +122,7 @@ func TestGRPC(t *testing.T) {
 			desc:           "Http client calling gRPC backend invalid query parameter, causing transcoding to fail by default",
 			clientProtocol: "http",
 			method:         "/v1/shelves/200?key=api_key&foo=bar",
-			wantError:      "503 Service Unavailable", // Transcoder does not send an application/grpc to the backend
+			wantError:      "http response status is not 200 OK: 415 Unsupported Media Type", // Transcoder does not send an application/grpc to the backend
```
nareddyt, line 125: Why the error message change? Seems like it is client-side error instead of server-side error now

shuoyang2016: Not entirely sure. My guess is that it's due to gRPC server change. Why it's 503 before? Does gRPC server return 503 when the media type is not expected? 

Invalid query params should be a client error in the first place.

nareddyt: I think what happens is: ESPv2 does NOT enable [strict request rejection](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/grpc_json_transcoder/v3/transcoder.proto#envoy-v3-api-msg-extensions-filters-http-grpc-json-transcoder-v3-grpcjsontranscoder-requestvalidationoptions) by default in transcoder filter.

So when client sends a bad query param, ESPv2 doesn't run transcoding and forwards raw HTTP/JSON request to gRPC backend, resulting in HTTP 503 on backend.

> Invalid query params should be a client error in the first place.

Yeah, but pre-existing behavior makes it a server problem for us :) We don't want to change that as it can be breaking change for clients

> My guess is that it's due to gRPC server change.

SG if so

shuoyang2016: Yeah, it seems gRPC backend's error code changes. Probably it's already broken for espv2 users who upgrade their gRPC servers. espv2 cannot help anyway.

nareddyt: Yeah good point. SG

PR #793, opened 2023-03-21
Support http backend override in http backend rule. 

shuoyang2016:
This is noop for espv2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,5 @@
+{
+  "dependencies": {
+    "bookstore": "file:tests/endpoints/bookstore"
+  }
+}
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -0,0 +1,5 @@
+{
```
TAOXUY, line 1: Is this file needed?

TAOXUY (commented):
This PR seems a combination of sync for g3 changes like cl/517310974. Could you please put all these changes in the description for context?

nareddyt (approved):

package.json
shuoyang2016: It's auto generated when install dependencies.

shuoyang2016:
> This PR seems a combination of sync for g3 changes like cl/517310974. Could you please put all these changes in the description for context?

Issue #782, opened 2023-01-25
Memory leak in v2.41.0?


davidgwps:
Hey, today we found that when we updated to v2.41.0 of ESPv2 there appears to be a memory leak 

![image](https://user-images.githubusercontent.com/88321759/214626270-98ec4af1-0897-4459-ade9-d1fb90e28976.png)

by contrast v2.38.0 does not have this problem and stays at a low memory usage

![image](https://user-images.githubusercontent.com/88321759/214628638-f572698f-02b3-4ea4-84dc-077b42a23bff.png)

see also this graph prior to the rollout, the big ramp whilst v2.41.0 was in use, and reverting the change restoring the memory back to normal levels

![image](https://user-images.githubusercontent.com/88321759/214630168-57ed8a66-a71f-4621-94ea-b2d67f9b9b90.png)

Our implementation:

```
        - name: esp
          image: gcr.io/endpoints-release/endpoints-runtime:2.41.0
          imagePullPolicy: Always
          args: [
            "--listener_port=8083",
            "--backend=127.0.0.1:8080",
            "--service={{ .Values.cloudEndpoints.hostName }}",
            "--rollout_strategy=managed",
          ]
```

nareddyt:
Thanks for reporting this. I have not been able to reproduce this yet with our test configurations. I wonder if there is a specific feature you are using that triggers this.

If possible, can you email your service configurations to nareddyt@google.com? You can download the file via:

```
gcloud endpoints configs describe "${CONFIG_ID}" \
  --project="${PROJECT}" \
  --service="${SERVICE}" \
  --format=json \
  > service.json
```

nareddyt:
Hi @davidgwps , I spent some time trying to reproduce it but haven't found any issues. This includes running ESPv2 with your configuration and also monitoring our long-running stress tests (https://testgrid.k8s.io/googleoss-esp-v2-periodic). Memory remained fairly constant, even under > 100 QPS load.

My recommendation for you to debug further: temporarily [enable the `--admin_port` flag](https://cloud.google.com/endpoints/docs/openapi/specify-esp-v2-startup-options#debugging). Pass the flag `--admin_port=9000` to view that admin web UI on port `9000`. Once you do this:

- You can monitor the [memory usage](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--memory) endpoint to confirm if the memory leak is actually caused by ESPv2.
[...]

PR #798, opened 2023-03-22
Reduce default jwt_cache_size to 1000

nareddyt:
Based on discussions in https://github.com/GoogleCloudPlatform/esp-v2/issues/782, the original estimate was incorrect. Reduce max memory usage from ~410 MB to 4.35 MB.

Fixes https://github.com/GoogleCloudPlatform/esp-v2/issues/782


docker/generic/start_proxy.py
```diff
@@ -594,7 +594,8 @@ def make_argparser():
         help='''
         Specify JWT cache size, the number of unique JWT tokens in the cache. The cache only stores verified
         good tokens. If 0, JWT cache is disabled. It limits the memory usage. The cache used memory
-        is roughly (token size + 64 bytes) per token. If not specified, the default is 100000.'''
+        is roughly (token size + 64 bytes) per token. If not specified, the default is 1000,
```
TAOXUY, line 597: "4KB per token at most"

TAOXUY (approved):

PR #821, opened 2023-05-08
refactor filtegen: Switch all filter generators to use OP Config constructor

nareddyt:
This PR modifies the pre-existing [filtegen package](https://github.com/GoogleCloudPlatform/esp-v2/tree/master/src/go/configgenerator/filtergen) to remove dependency on `ServiceInfo` and `MethodInfo`. All filter generators are now created directly from OP service config + ESPv2 flags. 

This allows us to support other input formats internally. It also clearly lists the dependencies of each filter generator, allowing us to selectively import only certain filter gens and reduce input processing.

Testing done:

- New unit tests to replace unit tests in `ServiceInfo`
- High-level config tests pass with no diff
- All integration tests pass with no diff


src/go/configgenerator/filtergen/backend_auth.go
```diff
@@ -35,38 +39,84 @@ const (
...
+	UniqueAudiences map[string]bool
+
+	// AudienceBySelector lists the JWT audience for each method.
+	AudienceBySelector map[string]string
+
+	IamURL                  string
+	MetadataURL             string
+	HttpRequestTimeout      time.Duration
+	DependencyErrorBehavior string
+	BackendAuthCredentials  *options.IAMCredentialsOptions
```
shuoyang2016, line 52: Do these have to be public?

src/go/configgenerator/filtergen/grpc_metadata_scrubber.go
```diff
@@ -26,30 +27,28 @@ const (
...
+type GRPCMetadataScrubberGenerator struct{}
 
-// NewGRPCMetadataScrubberGenerator creates the GRPCMetadataScrubberGenerator with cached config.
-func NewGRPCMetadataScrubberGenerator(serviceInfo *ci.ServiceInfo) *GRPCMetadataScrubberGenerator {
-	return &GRPCMetadataScrubberGenerator{
-		skipFilter: !serviceInfo.Options.EnableGrpcForHttp1,
+// NewGRPCMetadataScrubberFilterGensFromOPConfig creates a GRPCMetadataScrubberGenerator from
+// OP service config + descriptor + ESPv2 options. It is a FilterGeneratorOPFactory.
+func NewGRPCMetadataScrubberFilterGensFromOPConfig(serviceConfig *servicepb.Service, opts options.ConfigGeneratorOptions, params FactoryParams) ([]FilterGenerator, error) {
+	if !opts.EnableGrpcForHttp1 {
```
shuoyang2016, line 35: Add logs here when the filter gen is not enabled.

src/go/configgenerator/filtergen/jwt_authn.go
```diff
@@ -39,52 +39,92 @@ const (
...
+	// JWT Authn specific options below.
+
+	JwksCacheDurationInS               int
+	DisableJwksAsyncFetch              bool
+	JwksAsyncFetchFastListener         bool
+	JwksFetchNumRetries                int
+	JwksFetchRetryBackOffBaseInterval  time.Duration
+	JwksFetchRetryBackOffMaxInterval   time.Duration
+	JwtPadForwardPayloadHeader         bool
+	DisableJwtAudienceServiceNameCheck bool
```
shuoyang2016, line 68: Make these private if they don't have to be public.

nareddyt: See comment above

src/go/configgenerator/filtergen/backend_auth.go
nareddyt: I decided to make everything public for now because it will help with copybara.

We will import these into google3 and write new constructors for various other config formats (i.e. TD). They will need to be in a new package in google3, so these should be publicly visible to that new package.

src/go/configgenerator/filtergen/grpc_metadata_scrubber.go
nareddyt: Done

src/go/configgenerator/filtergen/backend_auth.go
shuoyang2016: > I decided to make everything public for now because it will help with copybara.
Could you explain more about this? Isn't copybara just copying text?

>We will import these into google3 and write new constructors for various other config formats (i.e. TD). They will need to be in a new package in google3, so these should be publicly visible to that new package.

If we'll have new constructors, don't we want to hide these params as private to consumers can only use constructors to build objects?

src/go/configgenerator/filtergen/backend_auth.go
```diff
@@ -35,38 +39,85 @@ const (
...
+	}, nil
 }
 
 func (g *BackendAuthGenerator) FilterName() string {
 	return BackendAuthFilterName
 }
 
-func (g *BackendAuthGenerator) IsEnabled() bool {
-	return len(g.audMap) > 0
+func (g *BackendAuthGenerator) MatchAudience(selector string) (string, bool) {
```
shuoyang2016, line 87: make this private too?

src/go/configgenerator/filtergen/backend_auth.go
```diff
@@ -35,38 +39,85 @@ const (
...
+	if audience, ok := g.AudienceBySelector[selector]; ok {
+		return audience, true
+	}
+
+	// Try matching CORS selector.
+	originalSelector, err := CORSSelectorToSelector(selector)
+	if err != nil {
+		// Ignore error, assume no route match.
```
shuoyang2016, line 95: Nit: add a log for the error too

shuoyang2016 (approved):
The rest code LGTM. Thanks

TAOXUY (approved):

src/go/configgenerator/filtergen/backend_auth.go
nareddyt: Changed logic here to separate against "no route match" and error.

src/go/configgenerator/filtergen/backend_auth.go
nareddyt: Yeah this can be private, done.

src/go/configgenerator/filtergen/backend_auth.go
nareddyt: But the constructors will be in a new package, as these new config formats will be in google3, not OSS.

PR #823, opened 2023-05-16
refactor filtergen: final minor changes to use refactor internally

nareddyt:
Changes:

1) Remove `FactoryParams` and instead have each `FilterGenerator` define its own params. SC v2 filter now has its own `ServiceControlOPFactoryParams`. We use currying via lambda to create a `FilterGeneratorOPFactory` that still takes in the extra params (similar to `absl::bind_front`).
2) Split `NewGRPCTranscoderFilterGensFromOPConfig` into another function for ease of import internally. Also add a new `doChecks bool` param to skip sanity checks, as currently we skip these internally.
3) Rename `GetDisabledSelectorsFromOPConfig()` to `GetHTTPBackendSelectorsFromOPConfig()`
4) Split `MakeMethodRequirementsFromOPConfig()` into multiple functions for ease of import internally.
5) Remove `ServiceInfo::getMethod()`. No need to error if method doesn't exist, just skip the selector. Needed internally for API filtering functionality. It is technically a behavior change, but it shouldn't matter in prod because API Compiler already does this validation.
6) Ensure API names follow the same order as the service config.

Testing done:
- No change to generated configs
- PR works internally


src/go/configgenerator/filtergen/grpc_transcoder.go
```diff
@@ -62,8 +62,25 @@ type GRPCTranscoderGenerator struct {
...
+
+	return []FilterGenerator{
+		grpcGen,
+	}, nil
+}
+
+// NewGRPCTranscoderFilterGenFromOPConfig creates a single GRPCTranscoderGenerator.
+//
+// It also has the option to skip sanity checks on if the generator should be created.
+func NewGRPCTranscoderFilterGenFromOPConfig(serviceConfig *confpb.Service, opts options.ConfigGeneratorOptions, doChecks bool) (*GRPCTranscoderGenerator, error) {
```
TAOXUY, line 82: NIT: doSanityChecks

src/go/configinfo/service_info.go
```diff
@@ -952,15 +955,6 @@ func (s *ServiceInfo) processTypes() error {
 	return nil
 }
 
-// Get the MethodInfo by full name. Prefer to use this function when getting methods,
-// as it outputs an actionable error message.
-func (s *ServiceInfo) getMethod(name string) (*MethodInfo, error) {
-	if s.Methods[name] == nil {
```
TAOXUY, line 958: Are we sure we don't want this?

Are there cases: 
- we want to raise the error if it shouldn't happen
- we want to log the api method if it is OK some api methods aren't defined

src/go/configinfo/service_info.go
```diff
@@ -840,9 +843,9 @@ func (s *ServiceInfo) processApiKeyLocations() error {
 			}
 		}
 
-		method, err := s.getMethod(rule.GetSelector())
-		if err != nil {
-			return fmt.Errorf("error processing system parameter rule for operation (%v): %v", rule.Selector, err)
```
TAOXUY, line 845: Why we continue instead of stopping execution here? Would be a behavior change?

src/go/configinfo/service_info.go
nareddyt: > we want to raise the error if it shouldn't happen

It should never happen in ESPv2 because the API compiler already does this validation, as mentioned above.

> we want to log the api method if it is OK some api methods aren't defined

I agree this is a good case. But unfortunately many callers are using `ServiceInfo::Methods` directly, as seen in the screenshot I messaged you.

Since I will remove ServiceInfo soon, let's remove this behavior for now, and we can add logging in the replacement when generating routes.

src/go/configinfo/service_info.go
nareddyt: Yes it is behavior change, I fixed description. It is intended. 

We should only parse operation names from `apis/methods` section of service config. If we see selectors in other parts of the service config that we don't know about, we should skip them and continue generation. The pre-existing logic of stopping all generation is not super useful internally, as we can't easily mutate the service config.

BTW this won't have any behavior change in ESPv2 prod because API compiler already has validation - API compiler will disallow unknown selectors before this code runs. This corner case will only happen if someone manually modifies service config after API compiler, but I'm not aware of any customers that modify service config.

You can also see we don't have any tests that cover this error case in all our internal tests, because it never occurs with the service configs we generate today. https://screenshot.googleplex.com/BYgfuVjoJXipJtw

TAOXUY (approved):

shuoyang2016 (approved):

PR #824, opened 2023-05-20
refactor: create new `filtergentest` and `clustergentest` packages for test helpers

nareddyt:
Currently some test helpers are located in `clustergen_test` and `filtergen_test` packages. These test helpers need to be imported into other packages in google3, but the current package naming makes it impossible to import them.

Move these test helpers into `clustergentest` and `filtergentest` (no `_`) to allow importing them.


shuoyang2016 (commented):
I thought we can import test helpers? https://source.corp.google.com/piper///depot/google3/apiserving/cloudesf/tests/utils/file.go;rcl=476391560;l=12

src/go/configgenerator/clustergen/clustergentest/clustergentest.go
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package clustergen_test
+// Package clustergentest contains test helpers to test cluster generators.
+package clustergentest
```
TAOXUY, line 16: Rename to `utils`?
- it helps differentiate with clustergen_test?
- It is under `clustergen/clustergentest` so it clear that it include the corresponding test helpers

src/go/configgenerator/filtergen/filtergentest/filtergentest.go
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package filtergen_test
+// Package filtergentest contains test helpers to test filter generators.
+package filtergentest
```
TAOXUY, line 16: ditto?

src/go/configgenerator/clustergen/clustergentest/clustergentest.go
nareddyt: Not ideal. In golang packages imports are based on last path segment. So other libraries importing this will see it as `utils.SuccessOPTestCase`, which does not indicate cluster vs filter gen. go/go-utility-packages#good-package-names

BTW this is standard golang convention: https://stackoverflow.com/a/31794241/4402434

TAOXUY (approved):

PR #826, opened 2023-06-16
refactor: remove `ServiceInfo` from HTTP Connection Manager filter gen

nareddyt:
- Move HTTP connection manager into its own filter generator
- Remove the dependence on `ServiceInfo` from HTTP connection manager
- Remove `LocalReplyConfig`. It is not needed in ESPv2, we will override this filter gen in google3.
- Refactor tracing - main change is to remove creation of metadata fetcher from tracing library. Tracing project ID is dependency injected by config manager.
- Refactor tracing - move tracing options into their own struct


src/go/configgenerator/filtergen/http_connection_manager.go
```diff
@@ -0,0 +1,267 @@
...
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+const (
+	// HTTPConnectionManagerFilterName is the Envoy filter name for debug logging.
+	HTTPConnectionManagerFilterName = "envoy.filters.network.http_connection_manager"
+)
+
+type HTTPConnectionManagerGenerator struct {
```
TAOXUY, line 42: Why not put HCM under `configgenerator` folder not `configgenerator/filter` folder? Will we have different HCM HCM configgenrator inside g3? 

HCM containing all http filters is a one-level lower than those in `filter` folder so it is wired to put them together.

nareddyt: We will have a different HCM in google3. Remember you added custom local reply config for our use case? I removed that from ESPv2, but we will override it internally.

I would like to reuse some of the test utilities, like `filtergentest`. Not sure if moving it out of `filtergen` folder is good idea then.

In the future if we have more network filters, I can separate `filtergen` into `httpfiltergen` and `networkfiltergen`

nareddyt:
PTAL

shuoyang2016 (approved):

TAOXUY (approved):

Issue #752, opened 2022-11-15
Use FindDefaultCredentials and a local ADC when the metadata server isn't available


andrewpollock:
I've been removing unnecessary dependencies on exported Service Account keys (see https://github.com/google/osv.dev/pull/822 for context), which works as desired using Cloud Build's default Service Account when running under Cloud Build.

When running locally, things go south at https://github.com/GoogleCloudPlatform/esp-v2/blob/5c4538a5a46f34e30fca009cd190934fd366ceee/src/go/configmanager/config_manager.go#L137 though because of the lack of a metadata server.

Someone who knows more about Cloud Identity matters than me pointed me at https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials, which by the looks of it would fall back to a local ADC in the event of the metadata server not being available.

qiwzhang:
We may not have resource to support such feature requirement.  Supporting ESP in local development is not our priority either.

PR #837, opened 2023-07-18
feat: Add support for application default credentials

rjerrems:
Adds a new flag to allow users to opt for using [application default credentials](https://google.aip.dev/auth/4110) solving this issue https://github.com/GoogleCloudPlatform/esp-v2/issues/752

The flag is disabled by default.

This will allow the osv.dev project to have local development environments without having to to create and distribute security account keys.


src/go/configmanager/config_manager.go
```diff
@@ -132,11 +132,14 @@ func NewConfigManager(mf *metadata.MetadataFetcher, opts options.ConfigGenerator
 	// accessToken is unavailable from imds and --service_account_key must be
 	// set to generate accessToken.
 	// The inverse is not true. We can still use IMDS on GCP when service account key is specified.
-	if mf == nil && opts.ServiceAccountKey == "" {
-		return nil, fmt.Errorf("if flag --non_gcp is specified, flag --service_account_key must be specified")
+	if mf == nil && opts.ServiceAccountKey == "" && opts.EnableApplicationDefaultCredentials == false {
```
nareddyt, line 135: nit: `&& !opts.EnableApplicationDefaultCredentials`

nareddyt (commented):
Overall LGTM, thanks for making this change! Seems like we've created our own fork of ADC back in 2020 that supports a subset of https://google.aip.dev/auth/4110. It's good to standardize now on ADC.

Due to some tech debt, you also need to add the `--enable_application_default_credentials` arg to our python validation logic after these args. This is the flags logic that is directly exposed to the user and plumbed down to this golang code.

https://github.com/GoogleCloudPlatform/esp-v2/blob/759b7712cbf497b43b191d69bd70752fcd935fb3/docker/generic/start_proxy.py#L820-L839

Please consider how this fits in with pre-existing validation like this:

https://github.com/GoogleCloudPlatform/esp-v2/blob/759b7712cbf497b43b191d69bd70752fcd935fb3/docker/generic/start_proxy.py#L1082

nareddyt:
Feel free to ignore 

- ESPv2-anthos-cloud-run-e2e-anthos-cloud-run-http-bookstore — Job failed.
- ESPv2-presubmit-asan — Job failed.

Currently your code has no tests. That's OK, I will add an integration test later to verify your changes.

src/go/configmanager/config_manager.go
rjerrems: Given that the flag is set to False by default, I am not sure that this is correct? Otherwise - if the flag was not supplied and service account was also not supplied, we would not enter this if statement.

i.e opts.ServiceAccountKey = "" and opts.EnableApplicationDefaultCredentials = False

I did this to try and make it as backwards compatible as possible, so you need to explicitly use this flag for the new behaviour.

rjerrems: NVM, I misunderstood :)  - Fixed

docker/generic/start_proxy.py
```diff
@@ -1080,6 +1088,8 @@ def enforce_conflict_args(args):
     if args.non_gcp:
         if args.service_account_key is None:
             return "If --non_gcp is specified, --service_account_key has to be specified, or GOOGLE_APPLICATION_CREDENTIALS has to set in os.environ."
```
TAOXUY, line 1090: Please add unit test start_proxy.py. It should be easy.

rjerrems: Done

TAOXUY (approved):

nareddyt (approved):

PR #843, opened 2023-08-10
[WIP] Make autogenerated health check configurable

dchakarwarti:


src/go/configgenerator/filtergen/util.go
```diff
@@ -38,7 +38,7 @@ var (
 		"grpc.health.v1.Health.Watch": true,
 	}
 
-	corsOperationDelimiter = fmt.Sprintf(".%s_CORS_", util.AutogeneratedOperationPrefix)
+	corsOperationDelimiter = fmt.Sprintf(".%s_CORS_", options.DefaultConfigGeneratorOptions().HealthCheckAutogeneratedOperationPrefix)
```
nareddyt, line 41: Unfortunately you can't use `DefaultConfigGeneratorOptions()` as it wont have the override

Make this a helper function instead of const var. You can use `opts.HealthCheckAutogeneratedOperationPrefix` if you have access to `opts` from this function

https://github.com/GoogleCloudPlatform/esp-v2/blob/fc0eff5e527ecb535a96761e0a1d9af5e23af708/src/go/configinfo/service_info.go#L88

PR #844, opened 2023-08-10
refactor: generate CORS per-vHost config in CORS filter

nareddyt:
Started working on the refactor of `route_generator.go` so we can remove `ServiceInfo`. There are many items to clean up, expect quite a few PRs to finish up refactor work.

This PR:
1) Moves some logic from `route_generator.go` to each `FilterGenerator`, allowing route generation to be generic.
2) Introduces `NoopFilterGenerator` to reduce code duplication of optional methods.

### DIFF
--- a/src/go/configgenerator/filtergen/cors.go
+++ b/src/go/configgenerator/filtergen/cors.go
@@ -15,12 +15,17 @@
 package filtergen
 
 import (
+	"fmt"
+	"strconv"
+	"time"
+
 	"github.com/GoogleCloudPlatform/esp-v2/src/go/options"
-	"github.com/GoogleCloudPlatform/esp-v2/src/go/util/httppattern"
 	corspb "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/cors/v3"
+	matcherpb "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
 	"github.com/golang/glog"
 	servicepb "google.golang.org/genproto/googleapis/api/serviceconfig"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 const (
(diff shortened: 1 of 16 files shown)


src/go/configgenerator/route_generator.go
```diff
@@ -309,22 +318,23 @@ func makeRouteCors(serviceInfo *configinfo.ServiceInfo) (*corspb.CorsPolicy, []*
 	return cors, corsRoutes, nil
 }
 
-// makePerRouteFilterConfig generates the per-route config across all filters for a single method and http rule.
-func makePerRouteFilterConfig(method *configinfo.MethodInfo, httpRule *httppattern.Pattern, filterGenerators []filtergen.FilterGenerator) (map[string]*anypb.Any, error) {
+// makePerResourceFilterConfig generates filter config per-route or per-vHost.
+// Caller is responsible for passing the callback for the corresponding resource.
+func makePerResourceFilterConfig(filterGenerators []filtergen.FilterGenerator, callback perResourceCallback) (map[string]*anypb.Any, error) {
```
TAOXUY, line 323: NIT: makePerResourceFilterConfig and perResourceCallback are confusing.

Can we define makePerRouteFilterConfig/perRouteCallback and makePerHostFilterConfig/perHostCallback separately?

nareddyt: I see. I was trying to reduce code duplication but agreed. Done

src/go/configgenerator/filtergen/cors.go
```diff
@@ -15,42 +15,135 @@
...
+}
+
+func (g *CORSBasicGenerator) GenFilterConfig() (proto.Message, error) {
+	return &corspb.Cors{}, nil
+}
 
-	return []FilterGenerator{
-		&CORSGenerator{},
-	}, nil
+func (g *CORSBasicGenerator) GenPerHostConfig(vHostName string) (proto.Message, error) {
```
TAOXUY, line 105: 2 cents: the readability could be better if put basic and regex in one class, and make `GenPerHostConfig` depends on opts.CorsPreset

TAOXUY (approved):
LGTM overall

src/go/configgenerator/filtergen/cors.go
nareddyt: Done. I'm was trying to fix this problem, but I guess it's tech debt we can leave as-is.

```golang
type CORSGenerator struct {
	Preset string
	// AllowOrigin should only be set if preset=basic
	AllowOrigin string
	// AllowOriginRegex should only be set if preset=cors_with_regex
	AllowOriginRegex string
}
```

PR #845, opened 2023-08-14
refactor: add new `RouteGenerator` for CORS and Health Check

nareddyt:
Similar to `FilterGenerator` and `ClusterGenerator`, add the concept of a `RouteGenerator`.

Start with a few trivial generators and helpers. Much of the code is copied from `service_info.go` and `route_generator.go`.

No production change, these files are tested but not used yet. Copied code will be deleted in follow-up.


TAOXUY:
If the follow-up is for deleting only, maybe add in this PR as well so that reviewers could see which are copied and which are new.

nareddyt:
> If the follow-up is for deleting only, maybe add in this PR as well so that reviewers could see which are copied and which are new.

When I mean follow-up, I mean much later after we switch all of route config to use these generators. I still need to introduce a few generators before this can happen.

It's hard to delete prior code as it will require a lot of refactor to current prod code path. So I don't want to touch it yet. The code was copied exactly, some methods were merged together into one.

I'll add comments to each copied method to explain where it is forked from, hope this helps.

src/go/configgenerator/routegen/routegentest/routegentest.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	t.Run(tc.Desc, func(t *testing.T) {
+		opts := options.DefaultConfigGeneratorOptions()
+		if err := mergo.Merge(&opts, tc.OptsIn, tc.OptsMergeBehavior); err != nil {
+			t.Fatalf("Merge() of test opts into default opts got err: %v", err)
+		}
+
+		gotGenerators, err := factory(tc.ServiceConfigIn, opts)
+		if err != nil {
+			t.Fatalf("NewXYZRouteGensFromOPConfig() got error: %v", err)
```
TAOXUY, line 71: Change XYZ to the respective name?

src/go/configgenerator/routegen/routegentest/routegentest.go
```diff
@@ -0,0 +1,184 @@
...
+			t.Fatalf("Fail to convert generated virtual host config proto to JSON: %v", err)
+		}
+
+		if err := util.JsonEqual(tc.WantHostConfig, gotJson); err != nil {
+			t.Errorf("Fail during route config JSON comparison of generated virtual host \n %v", err)
+		}
+	})
+}
+
+// FactoryErrorOPTestCase is the shared struct to test a RouteGenerator with
```
TAOXUY, line 95: Where is it used?

nareddyt: removed, not used yet

src/go/configgenerator/routegen/routegentest/routegentest.go
nareddyt: Requires using reflection to get the factory function name. We shouldn't introduce reflection unnecessarily.

TAOXUY (approved):

PR #847, opened 2023-08-17
Make health check names configurable

dchakarwarti:
Few points to note:

+ Trying to set `corsOperationDelimiter` from `options.ConfigGeneratorOptions` in src/go/configinfo/service_info.go is breaking the unit tests as serviceInfo is not initialised for those tests. 
Thus, I added `corsOperationDelimiter` as a ConfigGeneratorOptions. This can be overriden as per requirements.

+ I avoided use of function variable in src/go/configgenerator/filtergen/util.go as there is no way to get `options.ConfigGeneratorOptions` within CORSSelectorToSelector() function. Please CMIIW.


src/go/configgenerator/filtergen/util.go
```diff
@@ -48,8 +48,10 @@ func MethodToSelector(api *apipb.Api, method *apipb.Method) string {
 
 // MethodToCORSSelector gets the corresponding autogenerated CORS selector for
 // an OP API and Method.
-func MethodToCORSSelector(api *apipb.Api, method *apipb.Method) string {
-	return api.GetName() + corsOperationDelimiter + method.GetName()
+func MethodToCORSSelector(api *apipb.Api, method *apipb.Method, opts options.ConfigGeneratorOptions) string {
+	// Override existing corsOperationDelimiter
+	corsOperationDelimiter = opts.CorsOperationDelimiter
```
nareddyt, line 53: I like moving `CorsOperationDelimiter` to the options struct. But I don't like continuing to use global variable `corsOperationDelimiter` and overriding it.

It's causing hidden dependencies between functions via global variables, as `MethodToCORSSelector` has side effects. A developer needs to remember to call `CORSSelectorToSelector` due to the global variable change.

Can you find a way to propagate CorsOperationDelimiter into `CORSSelectorToSelector`? You can store as struct variable in filter generators and then pass it down. It makes deps clearer

dchakarwarti: Passing down a structure seems a really good way to avoid dependency between two functions. Thanks for the direction! :)

Seems like we also don't require `CorsOperationDelimiter` as part of options struct anymore. But I am keeping it same as it can help developers override it independent of `HealthCheckAutogeneratedOperationPrefix`.

src/go/configgenerator/filtergen/backend_auth.go
```diff
@@ -53,6 +53,8 @@ type BackendAuthGenerator struct {
 
 	AccessToken *helpers.FilterAccessTokenConfiger
 
+	CorsUtility
```
nareddyt, line 56: nit: no need for its own struct. How about capture variable on line 52 as `CORSOperationLimiter: string`

src/go/configgenerator/filtergen/util.go
```diff
@@ -48,19 +46,19 @@ func MethodToSelector(api *apipb.Api, method *apipb.Method) string {
 
 // MethodToCORSSelector gets the corresponding autogenerated CORS selector for
 // an OP API and Method.
-func MethodToCORSSelector(api *apipb.Api, method *apipb.Method) string {
-	return api.GetName() + corsOperationDelimiter + method.GetName()
+func MethodToCORSSelector(api *apipb.Api, method *apipb.Method, corsUtility CorsUtility, opts options.ConfigGeneratorOptions) string {
```
nareddyt, line 49: opts is no longer needed

nareddyt (commented):
Looking good, thanks!

src/go/configgenerator/filtergen/util.go
```diff
@@ -48,19 +46,19 @@ func MethodToSelector(api *apipb.Api, method *apipb.Method) string {
 
 // MethodToCORSSelector gets the corresponding autogenerated CORS selector for
 // an OP API and Method.
-func MethodToCORSSelector(api *apipb.Api, method *apipb.Method) string {
-	return api.GetName() + corsOperationDelimiter + method.GetName()
+func MethodToCORSSelector(api *apipb.Api, method *apipb.Method, CORSOperationDelimiter string) string {
```
nareddyt, line 49: nit: naming

`corsOperationDelimiter` for local variable use (here)
`CORSOperationDelimiter` for exported struct use

nareddyt (approved):

src/go/configmanager/flags/flags.go
```diff
@@ -30,14 +30,15 @@ var (
...
+	CorsAllowHeaders       = flag.String("cors_allow_headers", defaults.CorsAllowHeaders, "set Access-Control-Allow-Headers to the specified HTTP headers")
+	CorsAllowMethods       = flag.String("cors_allow_methods", defaults.CorsAllowMethods, "set Access-Control-Allow-Methods to the specified HTTP methods")
+	CorsAllowOrigin        = flag.String("cors_allow_origin", defaults.CorsAllowOrigin, "set Access-Control-Allow-Origin to a specific origin")
+	CorsAllowOriginRegex   = flag.String("cors_allow_origin_regex", defaults.CorsAllowOriginRegex, "set Access-Control-Allow-Origin to a regular expression")
+	CorsExposeHeaders      = flag.String("cors_expose_headers", defaults.CorsExposeHeaders, "set Access-Control-Expose-Headers to the specified headers")
+	CorsMaxAge             = flag.Duration("cors_max_age", defaults.CorsMaxAge, "set Access-Control-Max-Age response header for CORS preflight request.")
+	CorsPreset             = flag.String("cors_preset", defaults.CorsPreset, `enable CORS support, must be either "basic" or "cors_with_regex"`)
+	CorsOperationDelimiter = flag.String("cors_operation_delimiter", defaults.CorsOperationDelimiter, "Delimiter for cors operations")
```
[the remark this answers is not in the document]
dchakarwarti: Actually if we don't, it breaks the flags_test: 
`DefaultConfigGeneratorOptions does not match envoyConfigOptionsFromFlags`

PR #851, opened 2023-09-14
Bump version to 2.45.1

yaoluz:


nareddyt (approved):
TBH we've never tried a patch release, AFAIK we only do minor version upgrades. I guess you can try it out, make sure out tooling supports like (like upload to GCR)

PR #852, opened 2023-09-21
Add GFX filter in espv2 config generator

dchakarwarti:
Added **NewGRPCFieldExtractionFilterGensFromOPConfig** to generate GFX filter config as part of config generator. 

* Working on resolving dependencies issues in [src/go/util/marshal.go](https://github.com/GoogleCloudPlatform/esp-v2/pull/852/files#diff-0c0419a43cabd11d8d31795fe86e3e783b3a5e397883231c23a008bc0f0b83e8). This proto shall be used in adding GFX filter using config generator.


src/go/configgenerator/filtergen/grpc_field_extraction.go
```diff
@@ -0,0 +1,171 @@
...
+
+// NewGRPCFieldExtractionFilterGenFromOPConfig creates a single GRPCFielcExtractorGenerator.
+//
+// It also has the option to skip the filter. If enabled, all checks will occur
+// and filter may not be created. Otherwise, filter is always created.
+func NewGRPCFieldExtractionFilterGenFromOPConfig(serviceConfig *confpb.Service, opts options.ConfigGeneratorOptions, maybeSkipFilter bool) (*GRPCFieldExtractionGenerator, error) {
+	var descriptorBin []byte
+	foundDescriptor := false
+
+	for _, sourceFile := range serviceConfig.GetSourceInfo().GetSourceFiles() {
```
nareddyt, line 74: Can you move this into a helper function to share between transcoder and this filter gen?

src/go/configgenerator/filtergen/grpc_field_extraction.go
```diff
@@ -0,0 +1,171 @@
...
+		grpcGen,
+	}, nil
+}
+
+// NewGRPCFieldExtractionFilterGenFromOPConfig creates a single GRPCFielcExtractorGenerator.
+//
+// It also has the option to skip the filter. If enabled, all checks will occur
+// and filter may not be created. Otherwise, filter is always created.
+func NewGRPCFieldExtractionFilterGenFromOPConfig(serviceConfig *confpb.Service, opts options.ConfigGeneratorOptions, maybeSkipFilter bool) (*GRPCFieldExtractionGenerator, error) {
+	var descriptorBin []byte
```
nareddyt, line 71: Similar to transcoder filter, you should check if gRPC support is required first, and skip filter if not. We don't need this filter if user configured OpenAPI backend (supported for ESPv2).

```golang
	isGRPCSupportRequired, err := IsGRPCSupportRequiredForOPConfig(serviceConfig, opts)
	if err != nil {
		return nil, err
	}
	if !isGRPCSupportRequired {
		glog.Infof("gRPC support is NOT required, skip gRPC field extraction filter completely.")
		return nil, nil
	}
```

src/go/configgenerator/filtergen/grpc_field_extraction.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+
+	// Cannot check `descriptorBin == nil` because many tests use empty descriptor.
+	if !foundDescriptor {
+		// b/148605552: Previous versions of the `gcloud_build_image` script did not download the proto descriptor.
+		// We cannot ensure that users have the latest version of the script, so notify them via non-fatal logs.
+		// Log as error instead of warning because error logs will show up even if `--enable_debug` is false.
+		glog.Error("Unable to setup gRPC Field Extraction because no proto descriptor was found in the service config. " +
+			"Please use version 2020-01-29 (or later) of the `gcloud_build_image` script. " +
```
nareddyt, line 93: We can probably break this behavior now, it's been 3 years :)

Let's change it to an error like "Unable to setup gRPC field extraction because no proto descriptor was found in the service config."

src/go/configgenerator/filtergen/grpc_field_extraction.go
```diff
@@ -0,0 +1,171 @@
...
+			"http://google3/third_party/espv2/source/v12/blob/master/docker/serverless/gcloud_build_image/gcloud_build_image")
+		return nil, nil
+	}
+
+	fds := &descpb.FileDescriptorSet{}
+	if err := proto.Unmarshal(descriptorBin, fds); err != nil {
+		glog.Error("failed to unmarshal protodescriptor, error: ", err)
+		return nil, fmt.Errorf("failed to unmarshal proto descriptor, error: %v", err)
+	}
+	descriptorPolicies := ParseMethodPoliciesFromDescriptorSet(fds, opts.MethodAllowlist)
```
nareddyt, line 103: We don't need method allowlist for ESPv2 usage. This was specifically for GDCH. I think you should only use `servicePolicies := ParseMethodPoliciesFromServiceConfig(serviceConfig)`

src/go/configgenerator/filtergen/grpc_field_extraction.go
```diff
@@ -0,0 +1,171 @@
...
+			return nil, fmt.Errorf("selector %q from supplementary OP config not found in proto descriptor set, "+
+				"double check method allowlist, api version, or typos", selector)
+		}
+		descriptorPolicies[selector] = servicePolicy
+	}
+
+	for _, methodPolicy := range descriptorPolicies {
+		for _, requestPolicy := range methodPolicy.GetRequestPolicies() {
+			if requestPolicy.GetIsServiceConsumer() || requestPolicy.GetSelector() != "" {
+				break
```
nareddyt, line 117: This code doesn't do anything.

Also please double check the OSS `policy.proto`. The `is_service_consumer` field is not in 3P yet, it is internal. You cannot use it here.

nareddyt (commented):
Please reconsider if this change should be in OSS repo or our internal repo

TAOXUY:
Can we add this filter gen in g3 since we won't use it in ESPv2 and g3 is definitely a better dev environment?

PR #856, opened 2023-10-03
feat: Add transcoding_match_unregistered_custom_verb to start_proxy.py

same-id:
Add missing flag transcoding_match_unregistered_custom_verb.

### DIFF
--- a/docker/generic/start_proxy.py
+++ b/docker/generic/start_proxy.py
@@ -932,6 +932,19 @@ environment variable or by passing "-k" flag to this script.
         in grpc-json transcoding. This is to support HTML 2.0<https://tools.ietf.org/html/rfc1866#section-8.2.1>.
         Set this flag to true to disable this feature.
         ''')
+
+    parser.add_argument(
+        '--transcoding_match_unregistered_custom_verb', action='store_true',
+        help='''
+        According to the http template[1], the custom verb is `":" LITERAL` at the end of http template.
+
+        For a request with `/foo/bar:baz` and `:baz` is not registered in any url_template, here is the behavior change
+         - By default, `:baz` will not be treated as custom verb, so it will match `/foo/{x=*}`.
+         - If the flag is provided, `:baz` is treated as custom verb, so it will NOT match `/foo/{x=*}` since the template doesn't use any custom verb.
+
+        [1](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L226-L231)
+        ''')
+
     parser.add_argument(
         '--disallow_colon_in_wildcard_path_segment', action='store_true',
         help='''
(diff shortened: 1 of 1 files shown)


docker/generic/start_proxy.py
```diff
@@ -932,6 +932,19 @@ def make_argparser():
         in grpc-json transcoding. This is to support HTML 2.0<https://tools.ietf.org/html/rfc1866#section-8.2.1>.
         Set this flag to true to disable this feature.
         ''')
+
+    parser.add_argument(
+        '--transcoding_match_unregistered_custom_verb', action='store_true',
+        help='''
+        According to the http template[1], the custom verb is **":" LITERAL** at the end of http template.
```
nareddyt, line 939: nit: can we use backticks \`":" LITERAL \` instead of `**` to signify the code blocks? Using `**` is confusing because `**` is valid syntax in the URI template.

I know it was copied over from flags definition, but this is user-facing so let's make it easier to read

docker/generic/start_proxy.py
```diff
@@ -932,6 +932,19 @@ def make_argparser():
...
+
+        For a request with */foo/bar:baz* and *:baz* is not registered in any url_template, here is the behavior change
+         - if the field is not set, *:baz* will not be treated as custom verb, so it will match **/foo/{x=*}**.
+         - if the field is set, *:baz* is treated as custom verb,  so it will NOT match **/foo/{x=*}** since the template doesn't use any custom verb.
```
nareddyt, line 943: ditto on backticks instead of **

docker/generic/start_proxy.py
```diff
@@ -932,6 +932,19 @@ def make_argparser():
...
         Set this flag to true to disable this feature.
         ''')
+
+    parser.add_argument(
+        '--transcoding_match_unregistered_custom_verb', action='store_true',
+        help='''
+        According to the http template[1], the custom verb is **":" LITERAL** at the end of http template.
+
+        For a request with */foo/bar:baz* and *:baz* is not registered in any url_template, here is the behavior change
+         - if the field is not set, *:baz* will not be treated as custom verb, so it will match **/foo/{x=*}**.
```
nareddyt, line 942: Edit bullet points to make behavior clearer

- "By default, ..."
- "If the **flag** is set to true, ..."

docker/generic/start_proxy.py
same-id: Done, this is true for single `*` also, so I changed there too

docker/generic/start_proxy.py
same-id: Done

docker/generic/start_proxy.py
same-id: Done

nareddyt (approved):

nareddyt:
@TAOXUY PTAL and make sure we captured the behavior correctly in the `start_proxy.py` doc

TAOXUY (approved):

same-id:
What's the next steps for this? IIUC the failure in the tests is unrelated

same-id:
Thanks for the swift upstream.
When do you think this will be available as an image I can use from `gcr.io/endpoints-release/endpoints-runtime`?


P.S.
Also currently looking at supporting ext_auth in our fork, which I think is a great feature in espv2

TAOXUY:
We should be able to make a new release in next week. 

As for adding support for ext_authz, currently we don't have resource to add new feature.  If it is a small change, please make a PR and we can review it.

PR #889, opened 2024-04-04
Enable deterministic Go Protobuf marshaling (encoding)

stapelberg:
While developing Go Protobuf, we found this line to be responsible for a test failure when enabling fully lazy extensions (Google-internal bug reference: b/328746084).

Your test seems to compare Protobuf-encoded data. As explained on go/go-proto-stability, you should not rely on the Protobuf wire encoding being stable across different versions of your program.

go/go-proto-stability lists a couple of strategies you could pursue for a more reliable test, but for now, to unblock our optimization, enabling deterministic marshaling side-steps many vectors for change.


dchakarwarti:
cc: @shuoyang2016

PR #892, opened 2024-04-10
Propagate Api Key UID when it is returned by the service controller.

Elliot-xq:
Propagates the API key UID if it is returned in the service controller CheckResponse.

If it is not returned by the CheckResponse, we will still use the api_key as it is.

### DIFF
--- a/src/api_proxy/service_control/request_builder.cc
+++ b/src/api_proxy/service_control/request_builder.cc
@@ -383,7 +383,9 @@ Status set_credential_id(const SupportedLabel& l, const ReportRequestInfo& info,
     ASSERT(!info.api_key.empty(),
            "API Key must be set, otherwise consumer would not be verified.");
     std::string credential_id("apikey:");
-    credential_id += info.api_key;
+    credential_id = credential_id
+           + (info.check_response_info.api_key_uid.empty() ? info.api_key
+                      : info.check_response_info.api_key_uid);
     (*labels)[l.name] = credential_id;
   } else if (!info.auth_issuer.empty()) {
     std::string base64_issuer = Envoy::Base64Url::encode(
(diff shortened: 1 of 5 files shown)


src/api_proxy/service_control/request_builder.cc
```diff
@@ -383,7 +383,9 @@ Status set_credential_id(const SupportedLabel& l, const ReportRequestInfo& info,
     ASSERT(!info.api_key.empty(),
            "API Key must be set, otherwise consumer would not be verified.");
     std::string credential_id("apikey:");
-    credential_id += info.api_key;
+    credential_id = credential_id
+           + (info.check_response_info.api_key_uid.empty() ? info.api_key
+                      : info.check_response_info.api_key_uid);
```
TAOXUY, line 388: NIT: define credential_id as constchar above and use absl::StrCat to concat string.

src/api_proxy/service_control/request_info.h
```diff
@@ -176,6 +176,8 @@ struct CheckResponseInfo {
   // The trust level of the API Key that was checked.
   api_key::ApiKeyState api_key_state;
```
TAOXUY, line 178: NIT: add comment

TAOXUY (approved):

src/api_proxy/service_control/request_builder.cc
Elliot-xq: done.

src/api_proxy/service_control/request_info.h
Elliot-xq: done.

src/api_proxy/service_control/request_builder.cc
TAOXUY: I mean the constexpr char like in l348

Elliot-xq: done.

PR #897, opened 2024-05-02
Add a flag to opt out reporting api key uid and report unknown in network error.

Elliot-xq:
1) adds flag service_control_report_api_key_uid with default false. 
    TODO: enable it before release.

2) reports unknown in check request network failure.


src/api_proxy/service_control/request_builder.cc
```diff
@@ -384,8 +384,16 @@ Status set_credential_id(const SupportedLabel& l, const ReportRequestInfo& info,
       api_key::ApiKeyState::VERIFIED) {
     ASSERT(!info.api_key.empty(),
            "API Key must be set, otherwise consumer would not be verified.");
-    (*labels)[l.name] = absl::StrCat(kApiKeyPrefix, info.check_response_info.api_key_uid.empty() ? info.api_key
+    if (info.enable_api_key_uid_reporting) {
+      if (info.check_response_info.error.is_network_error) {
+        (*labels)[l.name] = absl::StrCat(kApiKeyPrefix, "UNKNOWN");
+      } else {
+        (*labels)[l.name] = absl::StrCat(kApiKeyPrefix, info.check_response_info.api_key_uid.empty() ? info.api_key
```
shuoyang2016, line 391: Do we still report API key if the flag is true?

Elliot-xq: Yes, according to Chemist team, for some old apis, they don't have api_key_uid, thus we still need to report api_key when api_key_uid is empty.

shuoyang2016: Got it. Thanks.

shuoyang2016 (approved):
Nit: add an integration test later https://github.com/GoogleCloudPlatform/esp-v2/blob/master/tests/integration_test/service_control_apikey_location_test/service_control_apikey_location_test.go

src/api_proxy/service_control/request_builder.cc
```diff
@@ -384,8 +384,16 @@ Status set_credential_id(const SupportedLabel& l, const ReportRequestInfo& info,
       api_key::ApiKeyState::VERIFIED) {
     ASSERT(!info.api_key.empty(),
            "API Key must be set, otherwise consumer would not be verified.");
-    (*labels)[l.name] = absl::StrCat(kApiKeyPrefix, info.check_response_info.api_key_uid.empty() ? info.api_key
+    if (info.report_api_key_uid) {
+      if (!info.check_response_info.error.name.empty()) {
+        (*labels)[l.name] = absl::StrCat(kApiKeyPrefix, "UNKNOWN");
+      } else {
+        (*labels)[l.name] = absl::StrCat(kApiKeyPrefix, info.check_response_info.api_key_uid.empty() ? info.api_key
                       : info.check_response_info.api_key_uid);
```
likeulb, line 392: The api_key_uid returned in CheckResponse contains the prefix "apikey:", or you have strip that before saving it under info.check_response_info?

Elliot-xq: I don't think so. let me remove the prefix. Will add integration test later to make sure it works.

Elliot-xq: how about now?

Elliot-xq:
> Nit: add an integration test later https://github.com/GoogleCloudPlatform/esp-v2/blob/master/tests/integration_test/service_control_apikey_location_test/service_control_apikey_location_test.go

sure, will do it in another change.

TAOXUY (approved):

PR #919, opened 2024-07-30
Update golang.org/x/net from v0.9.0 to v0.23.0

numanelahi:


gopkarthik (approved):
If no breaking changes, can you upgrade other golang.org/x repos' too.

PR #934, opened 2024-10-23
Update Envoy SHA

rongqi-g:
Update Envoy SHA based on comment #5 (https://b.corp.google.com/issues/374006086#comment5) in b/374006086.
Reference: https://github.com/envoyproxy/envoy/blob/v1.32.0/bazel/repository_locations.bzl


rongqi-g:
Resolved in https://github.com/GoogleCloudPlatform/esp-v2/pull/936.

PR #1015, opened 2026-05-28
fix out-of-bounds read in extractIPFromForwardedHeader on empty for= token

netliomax25-code:
When the `Forwarded` header arrives as `for=` (or `for=""`), `ip` becomes an empty `absl::string_view` and `ip[0]` / `ip[ip.size() - 1]` in handler_utils.cc:403-408 read out of bounds. Gate the unquote/unbracket steps on `ip.size() >= 2` so the strip only runs when there is something to strip.


netliomax25-code:
any update?

netliomax25-code:
gentle ping

PR #1016, opened 2026-06-01
guard long cast of numeric jwt claim in extractJwtPayload

netliomax25-code:
extractJwtPayload renders a numeric JWT claim via static_cast<long>(value.number_value()), but that cast is undefined when the claim is non-finite or outside long's range. A token whose logged number claim is e.g. 1e19 (> LONG_MAX) trips it; under UBSan: "1e+19 is outside the range of representable values of type 'long'". The number comes straight from the client JWT. Guard the narrowing the way JsonStruct::getInteger already does and render the raw value otherwise.


netliomax25-code:
gentle ping

PR #1017, opened 2026-06-02
fix use-after-free in extractHeader for multi-value headers

netliomax25-code:
extractHeader returns a string_view into the local GetAllOfHeaderAsStringResult. For a header sent with more than one value, getAllOfHeaderAsString joins the values into that result's backing string, so the returned view dangles once extractHeader returns and the result is destroyed. The callers in handler_impl.cc copy attacker-controlled headers like x-android-cert through it, reading freed heap memory. Return std::string so the bytes are copied while the backing string is still alive.


netliomax25-code:
gentle ping

PR #1027, opened 2026-07-13
update go and python versions, update SSL test keys

sunaydagli:


sunaydagli:
SSL keys needed to be updated, as they expired 5 years after their creation on July 12 2021. Added new ones to this PR, tests seen to pass again.

paulhong01 (approved):

Issue #791, opened 2023-03-20
StreamAggregatedResources warning logs on 2.41+


lsiqueira:
Hello,

After upgrading esp to 2.41 the following warning logs started:

```[24][config]StreamAggregatedResources gRPC config stream to @espv2-ads-cluster closed since 317673s ago: 14, upstream connect error or disconnect/reset before headers. reset reason: connection termination```

I'm using it in GKE with http only, with the following args:
```
        args:
            [
              "--listener_port=8081",
              "--backend=my-service:8080",
              "--service=my-service",
              "--rollout_strategy=managed",
              "-z",
              "healthz"
            ]
```

What can I do to fix this?

jdolinar:
Just to mention that we too have been getting this warning and it's across several ESPv2 versions.

nick-romano:
+1

PR #1031, opened 2026-08-03
Disable keepalive for local ADS cluster

### LINKED ISSUE #1030: Local ADS keepalive violates grpc-go policy and closes streams every 130 seconds
## Summary

ESPv2 applies a 30-second HTTP/2 connection keepalive to the local ADS Unix socket cluster. Config Manager creates its grpc-go server without a keepalive enforcement policy. grpc-go therefore uses its default five-minute minimum ping interval and closes the connection after the client exceeds the allowed ping strikes.

Envoy then logs that the StreamAggregatedResources stream closed and reconnects. This appears to explain the behavior reported in #791.

## Source evidence

- [ESPv2 sets a 30-second keepalive interval and a 10-second timeout](https://github.com/GoogleCloudPlatform/esp-v2/blob/v2.55.6/src/go/util/load_assignment.go#L27-L42).
- [The ADS bootstrap applies those HTTP/2 options to the local ADS cluster](https://github.com/GoogleCloudPlatform/esp-v2/blob/v2.55.6/src/go/bootstrap/ads/bootstrap.go#L75-L84).
- [Config Manager creates the gRPC server without grpc.KeepaliveEnforcementPolicy](https://github.com/GoogleCloudPlatform/esp-v2/blob/v2.55.6/src/go/configmanager/main/server.go#L54-L63).
- ESPv2 v2.55.6 uses grpc-go v1.56.3.
- [grpc-go v1.56.3 defaults MinTime to five minutes](https://github.com/grpc/grpc-go/blob/v1.56.3/internal/transport/defaults.go#L32-L40).
- [grpc-go allows two ping strikes and sends GOAWAY with ENHANCE_YOUR_CALM and too_many_pings after the next violation](https://github.com/grpc/grpc-go/blob/v1.56.3/internal/transport/http2_server.go#L863-L914).
- [A server write resets the ping strikes](https://github.com/grpc/grpc-go/blob/v1.56.3/internal/transport/http2_server.go#L988-L1113).

PR #763 added the keepalive settings. v2.41.0 was the first release with that change, which matches the first affected version in #791.

## Runtime evidence

In a 12-hour sample, 3,840 ADS streams that received no server DiscoveryResponse had these lifetimes:

- p25: 127.2 seconds
- median: 129.0 seconds
- p75: 130.9 seconds

Streams that received one midstream DiscoveryResponse lasted about 222 to 227 seconds. Streams that received repeated responses lasted about 286 to 353 seconds.

The response-dependent increase matches grpc-go's reset-on-write behavior. A fixed Envoy timeout would not explain it.

The same warning occurs with v2.55.4 and v2.55.6. Config Manager remains running, and Envoy reconnects after each close. The data plane continues to serve the last accepted configuration.

## Impact

- The ADS connection closes and reconnects about every 130 seconds while idle.
- ESPv2 emits persistent warning logs and creates avoidable connection churn.
- The warning can hide a real ADS failure.
- A control plane update can be delayed during a reconnect.

## Proposed fix

Keep HTTP/2 enabled for the local ADS cluster, but omit connection_keepalive from that cluster.

The shared helper also configures real upstream HTTP/2 backend clusters. The change should preserve connection keepalive for those clusters.

A secondary option is to configure Config Manager with a grpc.KeepaliveEnforcementPolicy whose MinTime is safely below the 30-second client interval.

## Validation

An idle ADS stream should remain open for at least 15 minutes after the change. Backend HTTP/2 clusters should retain their current keepalive settings.

The source and timing behavior strongly support this cause. A packet capture that shows the too_many_pings GOAWAY payload would provide direct wire-level confirmation.

tenextom:
## Summary

Remove connection keepalive from the local ADS Unix socket cluster. Preserve HTTP/2 and connection keepalive for upstream backend clusters.

Config Manager uses the grpc-go default minimum interval of five minutes. Envoy sends keepalive pings every 30 seconds. The server closes the ADS stream after the client exceeds the allowed ping strikes.

This change stops the recurring ADS disconnects without changing backend keepalive behavior.

Fixes #1030.
Addresses #791.

## Tests

- `go test ./src/go/...`
- `go test -race -count=1 ./src/go/bootstrap/ads ./src/go/util`
- `go vet ./src/go/bootstrap/ads ./src/go/util`
- `goimports` on the changed Go files
- `misspell` v0.3.4 on the changed files

The repository's pinned macOS `misspell` binary exited with `SIGSEGV` on arm64. The same v0.3.4 source passed with the current Go toolchain.

### DIFF
--- a/src/go/util/load_assignment_test.go
+++ b/src/go/util/load_assignment_test.go
@@ -1,0 +1,44 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"testing"
+
+	httppb "github.com/envoyproxy/go-control-plane/envoy/extensions/upstreams/http/v3"
+)
+
+func TestCreateUpstreamProtocolOptionsIncludesKeepalive(t *testing.T) {
+	options, ok := CreateUpstreamProtocolOptions()[UpstreamProtocolOptions]
+	if !ok {
+		t.Fatalf("CreateUpstreamProtocolOptions() did not return %q", UpstreamProtocolOptions)
+	}
+
+	httpOptions := &httppb.HttpProtocolOptions{}
+	if err := options.UnmarshalTo(httpOptions); err != nil {
+		t.Fatalf("options.UnmarshalTo(%T) failed: %v", httpOptions, err)
+	}
+
+	keepalive := httpOptions.GetExplicitHttpConfig().GetHttp2ProtocolOptions().GetConnectionKeepalive()
+	if keepalive == nil {
+		t.Fatal("CreateUpstreamProtocolOptions() did not configure connection keepalive")
+	}
+	if got := keepalive.GetInterval().AsDuration(); got != Http2KeepaliveInterval {
+		t.Errorf("keepalive interval = %v, want %v", got, Http2KeepaliveInterval)
+	}
+	if got := keepalive.GetTimeout().AsDuration(); got != Http2KeepaliveTimeout {
+		t.Errorf("keepalive timeout = %v, want %v", got, Http2KeepaliveTimeout)
+	}
+}
--- a/src/go/util/load_assignment_test.go
+++ b/src/go/util/load_assignment_test.go
@@ -42,3 +42,23 @@ func TestCreateUpstreamProtocolOptionsIncludesKeepalive(t *testing.T) {
 		t.Errorf("keepalive timeout = %v, want %v", got, Http2KeepaliveTimeout)
 	}
 }
+
+func TestCreateUpstreamProtocolOptionsWithoutKeepalive(t *testing.T) {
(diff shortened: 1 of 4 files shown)


tenextom:
/cc @paulhong01

Could you please review this PR and add `/ok-to-test` if it looks reasonable? The root-cause details are in #1030.

src/go/util/load_assignment_test.go
paulhong01, line 1: Optional: Consider also adding a companion test for `CreateUpstreamProtocolOptionsWithoutKeepalive` to ensure complete unit test coverage for both functions

paulhong01 (commented):
/ok-to-test

Thanks for the fix! The root cause analysis and approach look great.

src/go/util/load_assignment_test.go
tenextom: Added in 29a7c609. The companion test verifies that HTTP/2 remains configured while connection keepalive is absent. Focused tests, race tests, go vet, and the full Go suite all pass locally.

tenextom:
@paulhong01 I added the optional companion test in 29a7c609; it verifies that HTTP/2 stays enabled while connection keepalive is absent. Fresh Prow is running against current master.

Local verification passes:

- `go test -count=1 ./src/go/...`
- `go test -race -count=1 ./src/go/bootstrap/ads ./src/go/util`
- `go vet ./src/go/bootstrap/ads ./src/go/util`

I also traced the previous failures. API/TSAN were timing or brittle-assertion flakes (the same cases passed in normal presubmit on the same SHA), ASAN was an aborted pod timeout, and the Cloud Function/App Engine/Anthos jobs failed during external setup before ESPv2 ran. The same systemic red pattern appears on merged PRs #1033, #1035, and #1036.

Could you take another look and add `/lgtm` and `/approve` if this is ready? If those known-broken contexts repeat, it may need the same maintainer merge path as the recent PRs.

paulhong01 (approved):

